PR #91, opened 2022-11-06
Fix compilation errors

GavinRay97:


beef9999:
Will check cmake documents later.

beef9999:
The DIRS plural only applies to GTEST_INCLUDE_DIRS. Gmock was added in CMake version 3.23. And GFlags never exists. It's a user defined variable, when using find_library.

PR #93, opened 2022-11-08
multiple changes

beef9999:


GavinRay97:
Coroutines support, I was just going to ask about this -- awesome!!

beef9999:
It’s just a C++20 coroutine wrapper…

The underlay is still stackful coroutine. The default C++ 20 coroutine is stackless coroutine.

GavinRay97:
Right, I just find the `co_await`, etc API's easier to understand than `std::thread`, coming from TypeScript background 👍

beef9999:
You mean standard C++ thread or Photon’s thread?

BTW , if you are familiar with typescript, would you like to write a simple echo server and send to me? So I can test its performance with other libs & languages

GavinRay97:
Both types of thread -- Node.js is single-threaded, if you want to do multi-core programming you have to do it by creating `N` number of processes managed by a daemon 😅 

> BTW , if you are familiar with typescript, would you like to write a simple echo server and send to me? So I can test its performance with other libs & languages

Sure, I can do this easily 👍

PR #102, opened 2022-12-04
Fix: locker can't use CTAD

4kangjc:
locker can't use [CTAD](https://en.cppreference.com/w/cpp/language/class_template_argument_deduction)

### DIFF
--- a/thread/thread.h
+++ b/thread/thread.h
@@ -215,12 +215,10 @@ namespace photon
         int32_t recursive_count = 0;
     };
 
-    template<typename M0>
+    template <typename M>
     class locker
     {
     public:
-        using M1 = typename std::decay<M0>::type;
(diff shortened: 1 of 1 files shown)


4kangjc:
> GCC 4.9 can use `std::decay_t` and `std::remove_pointer_t`

And the project already use `std::enable_if_t`

Coldwings:
Sorry for the delayed reply.

Using `std::decay_t` and `std::remove_pointer_t` looks good to me.
Setting type template arguments in macro is better than current implementation, and may reducing generated template code （considering `locker<T*>` and `locker<T>` acts exactly same).

Coldwings (approved):
LGTM

beef9999 (approved):

PR #103, opened 2022-12-04
Add this_thread namespace

4kangjc:


beef9999:
`thread/std-compact.h` already has an equivalent of std::this_thread::sleep

4kangjc:
OK, my fault

PR #157, opened 2023-07-14
Add a Java benchmark and a Rust benchmark

beef9999:


lihuiba:
Does Java benchmark use the new virtual thread feature?

https://blogs.oracle.com/javamagazine/post/java-loom-virtual-threads-platform-threads

beef9999:
No, it doesn't. The Java netty framework is still using async callback.

PR #190, opened 2023-09-23
ipv6 preview

beef9999:
Still WIP ~

### DIFF
--- a/doc/docs/introduction/how-to-build.md
+++ b/doc/docs/introduction/how-to-build.md
@@ -1,5 +1,5 @@
 ---
-sidebar_position: 2
+sidebar_position: 3
 toc_max_heading_level: 4
 ---
 
--- a/doc/docs/introduction/how-to-integrate.md
+++ b/doc/docs/introduction/how-to-integrate.md
@@ -1,5 +1,5 @@
 ---
-sidebar_position: 3
+sidebar_position: 4
 toc_max_heading_level: 4
 ---
 
--- a/doc/docs/introduction/photon-architecture.md
+++ b/doc/docs/introduction/photon-architecture.md
@@ -1,0 +1,20 @@
+---
+sidebar_position: 2
+toc_max_heading_level: 4
+---
+
+# Photon Architecture
+
+The major goal of Photon is to handle concurrency and I/O (including file I/O and networking I/O).
+
+![architecture](/img/photon.png)
+
+Its components are:
+
+* [Thread](../api/thread.md), [vCPU](../api/vcpu-and-multicore.md), timer, sync primitives, gdb extension, emulation of std::thread, task dispatching ...
+* Multiple IO wrappers: psync, posix_aio, libaio, io_uring
+* Multiple socket implementations: tcp (level-trigger/edge-trigger), unix-domain, zero-copy, libcurl, TLS support, etc.
+* High performance RPC client/server, HTTP client/server.
+* A POSIX-like filesystem abstraction and some implementations: local fs, http fs, fuse fs, e2fs, etc.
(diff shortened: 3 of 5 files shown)


net/utils.cpp
```diff
@@ -93,6 +100,31 @@ int _gethostbyname(const char *name, Delegate<int, IPAddr> append_op) {
...
 
+int _gethostbyname_ipv6(const char* name, Delegate<int, IPAddr> append_op) {
+    int idx = 0;
+    addrinfo* result = nullptr;
+    addrinfo* cur = nullptr;
+    addrinfo hints = {};
+    hints.ai_socktype = SOCK_STREAM;
+    hints.ai_flags = AI_PASSIVE;
+    hints.ai_family = AF_INET6;
+
```
beef9999, line 111: `gethostbyname` is deprecated, and does not support ipv6

net/socket.h
```diff
@@ -21,80 +21,215 @@ limitations under the License.
...
-        }
-        void from(const struct sockaddr_in& addr_in)
-        {
-            from_sockaddr_in(addr_in);
+        explicit EndPoint(const sockaddr_storage* sock_addr) {
+            from_sockaddr_storage(sock_addr);
+        }
+        bool is_ipv4() const {
+            return addr.is_ipv4();
+        };
```
beef9999, line 194: `sockaddr_storage` is defined in `#include <netinet/in.h>`, and is wide enough to hold any type of socket address

net/kernel_socket.cpp
```diff
@@ -981,19 +990,32 @@ class ETKernelSocketServer : public KernelSocketServer, public NotifyContext {
 /* ET Socket - End */
 
 LogBuffer& operator<<(LogBuffer& log, const IPAddr addr) {
-    return log.printf(addr.d, '.', addr.c, '.', addr.b, '.', addr.a);
+    if (addr.is_ipv4())
+        return log.printf(addr.a, '.', addr.b, '.', addr.c, '.', addr.d);
+    else {
+        return log.printf(addr.to_string());
+    }
```
beef9999, line 997: ipv6 addr may has the `::` form. Even we could print every u32 number by order, but using the `ntop` is much simpler.

net/socket.h
```diff
@@ -208,11 +345,31 @@ namespace net
...
+        if (x.is_ipv4()) {
+            hash<uint64_t> hasher;
+            return hasher((x.addr.to_nl() << 16) | x.port);
+        } else {
+            hash<std::string_view> hasher;
+            return hasher(std::string_view((const char*) &x, sizeof(x)));
+        }
     }
 };
+
```
beef9999, line 361: TODO: need a better hash function

net/socket.h
```diff
@@ -208,11 +345,31 @@ namespace net
...
 struct hash<photon::net::EndPoint> {
-    hash<uint64_t> hasher;
     size_t operator()(const photon::net::EndPoint& x) const {
-        return hasher((x.addr.to_nl() << 16) | x.port);
+        if (x.is_ipv4()) {
+            hash<uint64_t> hasher;
+            return hasher((x.addr.to_nl() << 16) | x.port);
+        } else {
+            hash<std::string_view> hasher;
+            return hasher(std::string_view((const char*) &x, sizeof(x)));
```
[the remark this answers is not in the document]
beef9999: > do we need to distinguish between v4 and v6?

Because string_view hash is not balanced enough for integers. For simplicity, we call use string_view hash. 

https://xxhash.com/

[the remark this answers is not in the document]
beef9999: Will change both of them to string_view hash.

PR #192, opened 2023-09-28
ipv6

beef9999:


net/socket.h
```diff
@@ -21,80 +21,215 @@ limitations under the License.
...
 
 #include <photon/common/stream.h>
 #include <photon/common/callback.h>
 #include <photon/common/object.h>
 
+#ifdef __linux__
+#define _in_addr_field s6_addr32
+#else // macOS
+#define _in_addr_field __u6_addr.__u6_addr32
+#endif
```
beef9999, line 35: Cannot use `using` or `typedef`, because of `A.B` in macOS

rpc/rpc.h
```diff
@@ -234,7 +234,8 @@ namespace rpc
     };
```
beef9999, line 235: RPC server can specify IPv6 by using a IPv6 socket, but client can't. So add this flag.

net/socket.h
```diff
@@ -21,81 +21,217 @@ limitations under the License.
...
+            } else {
+                inet_ntop(AF_INET6, &addr, text, INET6_ADDRSTRLEN);
+            }
+            str.assign(text, strlen(text));
+            return str;
+        }
+    public:
+        static IPAddr V6None() {
+            return IPAddr(htonl(0xffffffff), htonl(0xffffffff), htonl(0xffffffff), htonl(0xffffffff));
+        }
```
beef9999, line 142: Even though htonl(0xffffffff) is still 0xffffffff, add this to remind reader that the ip is stored in network byte order.

net/socket.h
```diff
@@ -21,81 +21,217 @@ limitations under the License.
...
+            return true;
+        }
+        // We regard the default IPv4 0.0.0.0 as undefined
+        bool undefined() const {
+            return *this == V4Any();
+        }
+        // Should ONLY be used for IPv4 address
+        uint32_t to_nl() const {
+            return addr._in_addr_field[3];
+        }
```
beef9999, line 105: If your network is still IPv4, use `ipaddr.to_nl()` to convert to uint32_t. So it will be compatible to the previous IP container.

net/socket.h
```diff
@@ -21,81 +21,217 @@ limitations under the License.
...
+            }
+            if (addr._in_addr_field[0] != 0 || addr._in_addr_field[1] != 0) {
+                return false;
+            }
+            return true;
+        }
+        // We regard the default IPv4 0.0.0.0 as undefined
+        bool undefined() const {
+            return *this == V4Any();
+        }
```
beef9999, line 101: A default IPAddr is V4Any 0.0.0.0, which is also `undefined`. (used to be called empty)

beef9999:
@lihuiba Ready to merge

net/socket.h
```diff
@@ -21,81 +21,220 @@ limitations under the License.
...
+            return addr.undefined() && port == 0;
+        }
+        std::string to_string() const {
+            return "[" + addr.to_string() + "]:" + std::to_string(port);
+        }
+    };
+
+    static_assert(sizeof(EndPoint) == 18, "Endpoint size incorrect");
+
+    struct sockaddr_storage {
```
lihuiba, line 186: If it's internal-use only, we can put it in base_socket.h

net/basic_socket.cpp
```diff
@@ -262,11 +262,11 @@ bool ISocketStream::skip_read(size_t count) {
 }
 
 int do_get_name(int fd, Getter getter, EndPoint& addr) {
-    struct sockaddr_in addr_in;
-    socklen_t len = sizeof(addr_in);
-    int ret = getter(fd, (struct sockaddr*) &addr_in, &len);
-    if (ret < 0 || len != sizeof(addr_in)) return -1;
-    addr.from_sockaddr_in(addr_in);
+    sockaddr_storage storage(addr);
```
lihuiba, line 265: Is it possible that ```addr``` is not initialized?

net/http/client.cpp
```diff
@@ -104,11 +106,11 @@ enum RoundtripStatus {
 
 class ClientImpl : public Client {
 public:
-    PooledDialer m_dialer;
+    std::unique_ptr<PooledDialer> m_dialer;
```
lihuiba, line 109: why change it into heap-allocated?

net/basic_socket.cpp
beef9999: Added a new `get_max_socklen`  to `sockaddr_storage`.

net/http/client.cpp
beef9999: Because PooledDialer has its own unique_ptr in default constructor, copy constructor is implicitly deleted.

So no way to assign it in the member initializer list.

net/utils-stdstring.h
```diff
@@ -0,0 +1,48 @@
...
+
+#pragma once
+
+#include <string>
+#include <photon/net/socket.h>
+
+namespace photon {
+namespace net {
+
+std::string to_string(const photon::net::IPAddr& addr) {
```
lihuiba, line 25: inline

net/utils-stdstring.h
```diff
@@ -0,0 +1,48 @@
...
+        ip4.s_addr = addr.to_nl();
+        inet_ntop(AF_INET, &ip4, text, INET_ADDRSTRLEN);
+    } else {
+        inet_ntop(AF_INET6, &addr, text, INET6_ADDRSTRLEN);
+    }
+    str.assign(text, strlen(text));
+    return str;
+}
+
+std::string to_string(const photon::net::EndPoint& ep) {
```
lihuiba, line 39: inline

lihuiba (approved):

PR #196, opened 2023-10-10
Fix libcurl comaptible

Coldwings:
In some version of libcurl, call `curl_multi_socket_action` during callbacks will be checked and prevented by libcurl recursive check, and got a failure that not expected.

Make calling `do_action` asynchronous, make it be called from photon thread stub, will never cause this kind of situation.


lihuiba (approved):

lihuiba:
the patch should also be cherry-picked to the branch that overlaybd depends on.

PR #233, opened 2023-11-05
Fix io_uring cascading engine

beef9999:


photon.cpp
```diff
@@ -38,7 +38,11 @@ static thread_local uint64_t g_event_engine = 0, g_io_engine = 0;
 
 // Try to init master engine with the recommended order
 #if defined(__linux__)
-static const int recommended_order[] = {INIT_EVENT_EPOLL, INIT_EVENT_IOURING, INIT_EVENT_SELECT};
+#if defined(PHOTON_URING)
+static const int recommended_order[] = {INIT_EVENT_IOURING, INIT_EVENT_EPOLL, INIT_EVENT_SELECT};
+#else
+static const int recommended_order[] = {INIT_EVENT_EPOLL, INIT_EVENT_SELECT};
+#endif
 #else   // macOS, FreeBSD ...
```
beef9999, line 46: 上次 EVENT_ENGINE_DEFAULT 的行为改了以后，突然发现跑不了io_uring的CI了

这里把 default 的行为改成跟编译开关挂钩：如果用户开启io_uring编译，表明其有能力适配内核版本，并定位一些关于io_uring的问题。

io/iouring-wrapper.cpp
```diff
@@ -260,6 +264,10 @@ class iouringEngine : public MasterEventEngine, public CascadingEventEngine {
 
         io_uring_prep_poll_remove(sqe, (__u64) &iter->second.io_ctx);
         io_uring_sqe_set_data(sqe, nullptr);
+        int ret = io_uring_submit(m_ring);
+        if (ret < 0) {
+            LOG_ERROR_RETURN(0, -1, "iouring: fail to submit when removing interest, ", ERRNO(-ret));
+        }
         return 0;
     }
```
beef9999, line 272: 之前开发的时候io_uring的cascading一直没跑通，以为是内核问题，没想到是这里漏了submit 🤣

io/iouring-wrapper.cpp
```diff
@@ -244,6 +244,10 @@ class iouringEngine : public MasterEventEngine, public CascadingEventEngine {
             io_uring_prep_poll_multishot(sqe, fd_interest.fd, fd_interest.interest);
         }
         io_uring_sqe_set_data(sqe, &pair.first->second.io_ctx);
+        int ret = io_uring_submit(m_ring);
```
lihuiba, line 247: We'd better to submit the SQEs just before waiting, in order to align with the batching behavior of the master engine. Low-latency mode can be added later, with in-kernel busy-polling.

photon.cpp
lihuiba: We'll provide official pre-built binaries that have iouring engine included, but not as the default one.

lihuiba: The UT of iouring can explicitly initialize iouring engine.

beef9999: @lihuiba  Most of the UT code are using `photon::init`, which by default is `EVENT_ENGINE_DEFAULT`.

How do we specify the io_uring engine?

lihuiba: The cases that test iouring can specify iouring explicitly.

beef9999: What about other cases? They might be affected by the event engine from every aspects.

PR #235, opened 2023-11-06
minor improvements to http

lihuiba:
* make ```void buf_append(char*& ptr, std::string_view sv)``` inline-able
* make some code more readable (hopefully)
* make some branching more efficient with ```? : ``` expression
* avoid ```std::to_string()```
* reimplement ```obsolete_reason()``` with efficient table lookup, instead of the big ```switch```

### DIFF
--- a/common/estring.h
+++ b/common/estring.h
@@ -321,12 +321,8 @@ protected:
     }
 
 public:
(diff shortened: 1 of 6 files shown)


liulanzheng (approved):
LGTM

beef9999 (approved):

PR #242, opened 2023-11-12
Some fixes for networking

beef9999:
1. RPC client now supports both v4 and v6. Don't need to create two clients any more.
2. HTTP client removes v6 temporarily. There are still a lot work to be done for HTTP to truly support IPv6 (url parse, domain resolve, single client that supports both sockets), leave this work to the future...
3. Backup errno when KernelSocketStream destruct
4. Endpoint adds a string constructor
5. CMake can build openssl and curl from source now

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,8 +28,8 @@ option(PHOTON_ENABLE_EXTFS "enable extfs" OFF)
 option(PHOTON_BUILD_DEPENDENCIES "" OFF)
 set(PHOTON_AIO_SOURCE "https://pagure.io/libaio/archive/libaio-0.3.113/libaio-0.3.113.tar.gz" CACHE STRING "")
 set(PHOTON_ZLIB_SOURCE "https://github.com/madler/zlib/releases/download/v1.2.13/zlib-1.2.13.tar.gz" CACHE STRING "")
-set(PHOTON_OPENSSL_SOURCE "" CACHE STRING "")
-set(PHOTON_CURL_SOURCE "" CACHE STRING "")
+set(PHOTON_OPENSSL_SOURCE "https://github.com/openssl/openssl/archive/refs/heads/OpenSSL_1_0_2-stable.tar.gz" CACHE STRING "")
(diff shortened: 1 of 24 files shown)


net/kernel_socket.cpp
```diff
@@ -983,25 +970,6 @@ class ETKernelSocketServer : public KernelSocketServer, public NotifyContext {
 
 /* ET Socket - End */
 
-LogBuffer& operator<<(LogBuffer& log, const IPAddr addr) {
```
beef9999, line 986: move to tail

net/kernel_socket.cpp
```diff
@@ -59,22 +60,6 @@ limitations under the License.
 #define AF_SMC 43
 #endif
 
-LogBuffer& operator<<(LogBuffer& log, const in_addr& iaddr) {
```
beef9999, line 62: move to tail

net/kernel_socket.cpp
```diff
@@ -99,8 +84,10 @@ class KernelSocketStream : public SocketStreamBase {
     }
     ~KernelSocketStream() override {
         if (fd < 0) return;
+        auto errno_bak = errno;
         shutdown(ShutdownHow::ReadWrite);
         close();
+        errno = errno_bak;
```
beef9999, line 90: backup the errno, otherwise it will be overwritten when `KernelSocketStream` destruct in `KernelSocketClient::do_connect`

CMakeLists.txt
```diff
@@ -304,20 +301,26 @@ target_link_libraries(photon_static
         PRIVATE ${static_deps}
 )
 
-# Merge static libs into libphoton.a
-# Do NOT use this target directly.
+# Merge static libs into libphoton.a for manual distribution.
+# Do NOT link to this target directly.
 if (NOT APPLE)
-    set(merge_command ar -crT libphoton.a)
```
beef9999, line 310: 原先的静态库是Thin的，不方便分发

net/kernel_socket.cpp
lihuiba: It's not always needed to save and restore errno in this destructor. We'd better do it in the context that needs it.

lihuiba (approved):

PR #259, opened 2023-11-17
Interface design of simple unified access to json, xml, yaml, ini, etc. 

lihuiba:
with examples and tests to show the usage

### DIFF
--- a/common/simple_dom.h
+++ b/common/simple_dom.h
@@ -81,19 +81,13 @@ class RootNode : public Node {
 protected:
     RootNode() = default;
 
-    // reference counting is only applied on root
+    // reference counting is only applied to root
     // node, which represents the whole document
     std::atomic<uint32_t> _refcnt{0};
 
 public:
-    void add_ref() {
-        _refcnt.fetch_add(1);
-    }
-    void rm_ref() {
-        auto cnt = _refcnt.fetch_sub(1);
-        if (cnt == 1)
-            delete this;
-    }
+    void add_ref() { ++_refcnt; }
+    void del_ref()  { if (--_refcnt == 0) delete this; }
 };
 
 #define IF_RET(e) if (_node) return e; \
@@ -117,7 +111,7 @@ struct NodeWrapper {
         auto rt = root_node();
         auto rrt = rhs.root_node();
         if (rt != rrt) {
-            if (rt) rt->rm_ref();
+            if (rt) rt->del_ref();
             if (rrt) rrt->add_ref();
         }
         _node = rhs._node;
@@ -125,15 +119,15 @@ struct NodeWrapper {
     }
     NodeWrapper& operator = (NodeWrapper&& rhs) {
         if (_node)
-            _node->root()->rm_ref();
+            _node->root()->del_ref();
         _node = rhs._node;
         rhs._node = nullptr;
         return *this;
     }
     ~NodeWrapper() {
-        _node->root()->rm_ref();
+        _node->root()->del_ref();
     }
-    NodeWrapper root() const       { return root_node(); }
+    NodeWrapper root() const       { IF_RET(_node->root()); }
     RootNode* root_node() const    { IF_RET(_node->root()); }
     NodeWrapper parent() const     { IF_RET(_node->_parent); }
     NodeWrapper next() const       { IF_RET(_node->_next); }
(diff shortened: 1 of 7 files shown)


common/stream.h
```diff
@@ -44,6 +45,19 @@ class IStream : public Object
...
+            void operator()(void* ptr) {
+                ::free(ptr);
+            }
+        };
+        std::unique_ptr<void, FreeDeleter> ptr;
+        ssize_t size;   // < 0 if error occured; |size| is always the # of bytes read
+    };
+
+    // read until EOF
+    ReadAll readall(size_t min_buf = 1024, size_t max_buf = 1024 * 1024 * 1024);
```
BigVan, line 59: is the param 'min_buf' necessary?

common/simple_dom.h
```diff
@@ -0,0 +1,274 @@
...
+// 1. simple & convenient usage;
+// 2. unified interface, common needs (may not fit for corner cases);
+// 3. efficient parsing (reading), as well as compiling;
+
+using str = estring_view;
+
+class RootNode;
+struct NodeWrapper;
+
+// the interface for internal implementations
```
liulanzheng, line 49: Can it be concealed for internal use and not disclosed externally?

common/simple_dom.h
```diff
@@ -0,0 +1,274 @@
...
+    bool valid() const { return _node; }
+    operator bool() const { return _node; }
+
+    size_t num_children() const {
+        return _node ? _node->num_children() : 0;
+    }
+    NodeWrapper get(size_t i) const {
+        IF_RET({_node->get(i)});
+    }
+    NodeWrapper get(str key) const {
```
liulanzheng, line 152: What is the result if the get operation fails?

common/simple_dom.h
```diff
@@ -0,0 +1,274 @@
...
+
+inline NodeWrapper Node::wrapper() { return {this}; }
+
+// 1. text is moved to the simple_dom object, which frees it when destruct.
+// 2. the content of text may be modified in-place to un-escape strings.
+// 3. returning a pointer (of Node) is more efficient than an object (of Document),
+//    even if they are equivalent in binary form.
+Node* parse(char* text, size_t size, int flags);
+
+inline Node* parse(IStream::ReadAll&& buf, int flags) {
```
liulanzheng, line 209: Isn't the Node meant for internal use? How come it's being used by users here?

common/simple_dom.h
```diff
@@ -0,0 +1,274 @@
...
+public:
+    void add_ref() { ++_refcnt; }
+    void del_ref()  { if (--_refcnt == 0) delete this; }
+};
+
+#define IF_RET(e) if (_node) return e; \
+                        else return {};
+
+// the interface for users
+struct NodeWrapper {
```
liulanzheng, line 97: For user use, it's better to be named as "Node". The original "Node" can be renamed to NodeBase or something else, and not be exposed externally.

lihuiba: OK

common/simple_dom.h
lihuiba: an NodeWrapper with nullptr pointer to node implementation

common/simple_dom.h
lihuiba: parse(...)->wrap()
```
// 3. returning a pointer (of Node) is more efficient than an object (of Document),
//    even if they are equivalent in binary form.
```

common/stream.h
lihuiba: in case that you want it start from, say, 1MB, to avoid some unnecessary realloc()

common/simple_dom.h
liulanzheng: For users, it's a direct use after parsing, and calling wrap is a burden.

common/simple_dom.h
liulanzheng: This requires a clear explanation and instructions for use.

common/simple_dom.h
lihuiba: Not a big problem

liulanzheng: It a problem. The returned NodeImpl* is not the type used by the user and users must first determine if the returned NodeImpl* is nullptr before calling wrap. it's best not to expose these to users.

common/simple_dom.h
lihuiba: No，because it's also the binary interface.

common/simple_dom.h
lihuiba: OK

Issue #261, opened 2023-11-18
Missed mutex unlock notification in io_uring engine


louiswilliams:
The io_uring engine can miss mutex wakeup notifications and will [sleep here for 10 seconds, stalling all threads](https://github.com/alibaba/PhotonLibOS/blob/684106ed860bc74efd104bd9865f781ff95e12e1/thread/thread.cpp#L1785-L1791). The conditions required to reproduce this are using multiple mutexes and many threads.

I can reliably reproduce by making the following changes:

**Bypass the spinning code in `mutex::lock`**:
```diff
diff --git a/thread/thread.cpp b/thread/thread.cpp
index d2a8542..9a19b97 100644
--- a/thread/thread.cpp
+++ b/thread/thread.cpp
@@ -1547,11 +1547,11 @@ R"(
     }
     int mutex::lock(uint64_t timeout)
     {
-        for (int tries = 0; tries < MaxTries; ++tries) {
-            if (try_lock() == 0)
-                return 0;
-            thread_yield();
-        }
         splock.lock();
         if (try_lock() == 0) {
             splock.unlock();
```

**Compile this program with `-O0` to make it slower:**
```c++
#include <atomic>
#include <chrono>
#include <cstdio>
#include <iostream>
#include <photon/photon.h>
#include <photon/thread/std-compat.h>
#include <unistd.h>
#include <vector>

static constexpr int nMutexes = 2;
static constexpr int nWorkers = 2;
static constexpr int nThreads = 16; // Increase if necessary

photon::mutex mutexes[nMutexes];
std::atomic<int64_t> acquisitionCounters[nMutexes];
std::atomic<int64_t> acquisitionDurations[nMutexes];

void timedLock(photon::mutex &mutex, int index) {
  auto start = std::chrono::steady_clock::now();

  {
    photon_std::lock_guard<photon::mutex> lk(mutex);
    auto end = std::chrono::steady_clock::now();
    acquisitionCounters[index].fetch_add(1);
    auto durationMicros =
        std::chrono::duration_cast<std::chrono::microseconds>(end - start)
            .count();
    acquisitionDurations[index].fetch_add(durationMicros);

    if (durationMicros > 1'000'000) {
      printf("long acquisition. mutex %d, duration: %ldms\n", index,
             durationMicros / 1000);
    }
  }
}

void myThread(int tid) {
  printf("thread %d starting\n", tid);

  while (true) {
    for (int i = 0; i < nMutexes; i++) {
[...]

PR #264, opened 2023-11-19
fix io_uring cancel_wait

beef9999:
Fix https://github.com/alibaba/PhotonLibOS/issues/261

### DIFF
--- a/io/iouring-wrapper.cpp
+++ b/io/iouring-wrapper.cpp
@@ -40,8 +40,7 @@ limitations under the License.
 
 namespace photon {
 
-constexpr static EventsMap<EVUnderlay<POLLIN | POLLRDHUP, POLLOUT, POLLERR>>
-    evmap;
+constexpr static EventsMap<EVUnderlay<POLLIN | POLLRDHUP, POLLOUT, POLLERR>> evmap;
 
 class iouringEngine : public MasterEventEngine, public CascadingEventEngine {
 public:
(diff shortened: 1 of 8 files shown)


io/iouring-wrapper.cpp
```diff
@@ -401,7 +410,8 @@ class iouringEngine : public MasterEventEngine, public CascadingEventEngine {
     struct ioCtx {
         photon::thread* th_id;
         int32_t res;
-        bool is_canceller;
+        bool is_io_canceller;
+        bool is_wait_canceller;
         bool is_event;
     };
```
beef9999, line 416: io_canceller  /  wait_canceller

io/iouring-wrapper.cpp
```diff
@@ -501,21 +511,9 @@ class iouringEngine : public MasterEventEngine, public CascadingEventEngine {
...
-            }
-            eventfd_t val;
-            if (eventfd_read(m_cancel_fd, &val) != 0) {
-                LOG_ERROR("iouring: read eventfd failed, `", ERRNO());
-            }
-        }
+    void run_wait_canceller() {
+        uint32_t poll_mask = evmap.translate_bitwisely(EVENT_READ);
+        async_io(&io_uring_prep_poll_multishot, -1, RingFlagsWaitCanceller, m_wait_canceller_fd, poll_mask);
     }
```
beef9999, line 517: `wait_canceller` will fall into sleep in `async_io`. Because the multishot poll needs to keep the `ioCtx` on stack for the whole time.

io/iouring-wrapper.cpp
```diff
@@ -501,21 +511,9 @@ class iouringEngine : public MasterEventEngine, public CascadingEventEngine {
         }
     }
 
-    void run_cancel_poller() {
-        while (m_cancel_poller_running) {
-            uint32_t poll_mask = evmap.translate_bitwisely(EVENT_READ);
-            int ret = async_io(&io_uring_prep_poll_add, -1, 0, m_cancel_fd, poll_mask);
-            if (ret < 0) {
```
beef9999, line 508: Should not use one-shot poll. There is a risk of missing eventfd notification, right before starting the next round of polling.

louiswilliams:
Thanks for fixing this so quickly! I tested this locally and it eliminates the stalls in my workload.

io/iouring-wrapper.cpp
```diff
@@ -130,16 +131,17 @@ class iouringEngine : public MasterEventEngine, public CascadingEventEngine {
 
         if (m_master) {
             // Setup a cancel poller to watch on master engine
-            m_cancel_fd = eventfd(0, EFD_CLOEXEC);
-            if (m_cancel_fd < 0) {
+            m_wait_canceller_fd = eventfd(0, EFD_CLOEXEC | EFD_NONBLOCK);
```
lihuiba, line 134: m_wait_canceller_fd and m_cascading_event_fd can be unified as m_eventfd?

io/iouring-wrapper.cpp
lihuiba: actually we don't need to use async_io() to register the multishot poll, and we can set the user data as nullptr.

When we get a event with nullptr data, we know it's cancel wait.

lihuiba (approved):

PR #283, opened 2023-12-01
Mute an UBSan check error for SCOPED_LOCK macro

beef9999:
The previous SCOPED_MEMBER_LOCK will visit member address even if the object itself is nullptr.

This behavior will cause an runtime error in the UndefinedBehaviorSanitizer check (https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html), although the compiler didn't regard it as a warning.

Removed SCOPED_MEMBER_LOCK, luckily this is the only usage in the whole project.

### DIFF
--- a/thread/thread.cpp
+++ b/thread/thread.cpp
@@ -72,8 +72,6 @@ inline int posix_memalign(void** memptr, size_t alignment, size_t size) {
    by target vcpu in resume_thread(), when its runq becomes empty;
 */
 
-#define SCOPED_MEMBER_LOCK(x) SCOPED_LOCK(&(x)->lock, ((bool)x) * 2)
-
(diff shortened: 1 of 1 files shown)


thread/thread.cpp
```diff
@@ -72,8 +72,6 @@ inline int posix_memalign(void** memptr, size_t alignment, size_t size) {
    by target vcpu in resume_thread(), when its runq becomes empty;
 */
 
-#define SCOPED_MEMBER_LOCK(x) SCOPED_LOCK(&(x)->lock, ((bool)x) * 2)
```
for-blinded-review, line 75: when x is null, it doesn't actually do the locking.

lihuiba (approved):

Issue #268, opened 2023-11-22
Improve event engine of epoll


lihuiba:
1. Simplify engine logic with multiple epollfd;
2. Using multi-shot mode;
3. Support for edge-trigger mode;

PR #295, opened 2023-12-08
New epoll event engine implemention that may be more effecient

Coldwings:
A new epoll event engine implementation (EPOLL_NG) which use epoll groups to manage read/write events.
This is much simpler and use no on-heap memory allocation, with better performance, as mentioned in #268 

Edge-trigger related feature will be implemented in future.

New engine shows 3%~5% throughput/QPS improvement by `net-perf` test.

### DIFF
--- a/io/epoll-ng.cpp
+++ b/io/epoll-ng.cpp
@@ -122,7 +122,7 @@ public:
         }
     };
 
-    enum class PULLERTYPE {
+    enum class POLLERTYPE : int {
         ENGINE = 0,
         READER = 1,
         WRITER = 2,
@@ -131,16 +131,18 @@ public:
     };
 
     Poller pl[4];
-    Poller& engine = pl[(int)PULLERTYPE::ENGINE];
-    Poller& rpoller = pl[(int)PULLERTYPE::READER];
-    Poller& wpoller = pl[(int)PULLERTYPE::WRITER];
-    Poller& epoller = pl[(int)PULLERTYPE::ERROR];
+
+#define engine (pl[(int)POLLERTYPE::ENGINE])
+#define rpoller (pl[(int)POLLERTYPE::READER])
+#define wpoller (pl[(int)POLLERTYPE::WRITER])
+#define epoller (pl[(int)POLLERTYPE::ERROR])
+#define poller(x) (pl[(int)x])
 
     int evfd = -1;
 
     int init() {
         for (int i = 0; i < 4; i++) {
-            if (pl[i].init() < 0) {
+            if (poller(i).init() < 0) {
                 LOG_ERROR("Failed to create sub poller `, because `", i,
                           ERRNO());
                 goto errout;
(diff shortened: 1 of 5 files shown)


io/epoll-ng.cpp
```diff
@@ -0,0 +1,311 @@
...
+                   epoller.notify_one(datacb, fdcb);
+            fired += turn;
+        } while (turn);
+        if (!fired) {
+            // no events ready
+            eventfd_t value;
+            engine.collect(timeout);
+            engine.notify_all(
+                [&](epoll_data_t data) __INLINE__ {
+                    if (data.fd == evfd) {
```
lihuiba, line 221: better to do it like this:
```cpp
switch(data) {
case 0:
	eventfd_read(evfd, &value);
	break;
case 1:
	rpoller.collect(0);
	break;
case 2:
	wpoller.collect(0);
	break;
case 3:
	epoller.collect(0);
	break;
}
```
so as to improve branch prediction

io/epoll-ng.cpp
```diff
@@ -0,0 +1,311 @@
...
+        template <typename DataCB, typename FDCB>
+        int notify_all(const DataCB& datacb, const FDCB& fdcb) {
+            int fired = 0;
+            while (remains && fdcb()) {
+                fired += notify_one(datacb, fdcb);
+            }
+            return fired;
+        }
+
+        void collect(uint64_t timeout) {
```
lihuiba, line 125: better name it "reap"

io/epoll-ng.cpp
```diff
@@ -0,0 +1,311 @@
...
+        if (ret < 0) {
+            return errno == ETIMEDOUT ? 0 : -1;
+        }
+        auto ptr = data;
+        auto end = data + count;
+        wait_for_events(
+            0, [&](epoll_data_t data) __INLINE__ { *ptr++ = data.ptr; },
+            [&]()
+                __INLINE__ {  // make sure each fd receives all possible events
+                    return (end - ptr) >= 3;
```
lihuiba, line 254: this seems not necessary any more?

io/epoll-ng.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+        int add(int fd, uint32_t events, epoll_data_t data) {
+            return ctl(fd, EPOLL_CTL_ADD, events, data);
+        }
+
+        int rm(int fd, uint32_t events, epoll_data_t data) {
+            return ctl(fd, EPOLL_CTL_DEL, events, data);
+        }
+
+        int wait(epoll_event* evs, size_t len, uint64_t timeout) {
```
lihuiba, line 82: wait() is only invoked by one-liner collect(). consider merging them.

io/epoll-ng.cpp
```diff
@@ -122,44 +97,73 @@ class EventEngineEPollNG : public MasterEventEngine,
...
+        ENGINE = 0,
+        READER = 1,
+        WRITER = 2,
+        ERROR = 3,
+        EVENT = 4,
+    };
+
+    Poller pl[4];
+    Poller& engine = pl[(int)PULLERTYPE::ENGINE];
+    Poller& rpoller = pl[(int)PULLERTYPE::READER];
```
lihuiba, line 135: xpoller can be defined as macros instead. (x = r, w, e)

io/epoll-ng.cpp
```diff
@@ -215,24 +219,18 @@ class EventEngineEPollNG : public MasterEventEngine,
...
-                        return;
+                    switch (data.u64) {
+                        case (uint64_t)PULLERTYPE::READER:
+                        case (uint64_t)PULLERTYPE::WRITER:
+                        case (uint64_t)PULLERTYPE::ERROR:
+                            pl[data.u64].reap(0);
+                            return;
+                        case (uint64_t)PULLERTYPE::EVENT:
+                            eventfd_read(evfd, &value);
+                            return;
```
lihuiba, line 233:
```
default:
    LOG_ERROR(...);

lihuiba (approved):

PR #298, opened 2023-12-14
fs: support xattr for extfs and subfs

yuchen0cc:
Extfs and subfs implement IFileSystemXAttr interface to support xattr.
Add xattr test case for extfs.

### DIFF
--- a/fs/subfs.cpp
+++ b/fs/subfs.cpp
@@ -31,10 +31,11 @@ namespace photon {
 namespace fs
 {
     // Sub tree of a file system
-    class SubFileSystem : public IFileSystem
+    class SubFileSystem : public IFileSystem, public IFileSystemXAttr
     {
     public:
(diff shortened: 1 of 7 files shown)


fs/subfs.cpp
```diff
@@ -228,48 +230,63 @@ namespace fs
             PathCat __(this, path);
             return underlayfs->mknod(path, mode, dev);
         }
-        /*
+
         virtual ssize_t getxattr(const char *path, const char *name, void *value, size_t size)
         {
+            if (!underlay_xattrfs)
+                LOG_ERROR_RETURN(ENOTSUP, -1, "xattr is not supported by underlay fs");
```
Coldwings, line 237: Better to use `ENOSYS` as error number to present "method not implemented" as same as other not implemented situation in Photon.

Coldwings (approved):

PR #302, opened 2023-12-14
[feat] http client with cert

WaberZhuang:
Support HTTP client to use specified certificates. It's necessary when two-way authentication is enabled (the server will verify the client's identity) on the HTTP server.

### DIFF
--- a/net/http/client.cpp
+++ b/net/http/client.cpp
@@ -35,19 +35,24 @@ static constexpr char USERAGENT[] = "EASE/0.21.6";
 class PooledDialer {
 public:
     net::TLSContext* tls_ctx = nullptr;
+    bool tls_ctx_ownership;
     std::unique_ptr<ISocketClient> tcpsock;
     std::unique_ptr<ISocketClient> tlssock;
     std::unique_ptr<Resolver> resolver;
 
     //etsocket seems not support multi thread very well, use tcp_socket now. need to find out why
-    PooledDialer() :
-            tls_ctx(new_tls_context(nullptr, nullptr, nullptr)),
+    PooledDialer(TLSContext *_tls_ctx) :
+            tls_ctx(_tls_ctx ? _tls_ctx : new_tls_context(nullptr, nullptr, nullptr)),
+            tls_ctx_ownership(_tls_ctx == nullptr),
             tcpsock(new_tcp_socket_pool(new_tcp_socket_client(), -1, true)),
             tlssock(new_tcp_socket_pool(new_tls_client(tls_ctx, new_tcp_socket_client(), true), -1, true)),
             resolver(new_default_resolver(kDNSCacheLife)) {
     }
 
-    ~PooledDialer() { delete tls_ctx; }
+    ~PooledDialer() {
+        if (tls_ctx_ownership)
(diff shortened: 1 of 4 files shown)


net/http/client.cpp
```diff
@@ -104,11 +111,19 @@ enum RoundtripStatus {
...
     CommonHeaders<> m_common_headers;
     ICookieJar *m_cookie_jar;
     ClientImpl(ICookieJar *cookie_jar) :
-        m_cookie_jar(cookie_jar) {}
+        m_cookie_jar(cookie_jar),
+        m_dialer(new PooledDialer()) {}
+
+    ClientImpl(ICookieJar *cookie_jar, TLSContext *tls_ctx) :
+        m_cookie_jar(cookie_jar),
+        m_dialer(new PooledDialer(tls_ctx)) {
```
beef9999, line 123: m_dialer(tls_ctx)?

So there will be no need to change m_dialer to a pointer.

WaberZhuang: done

net/http/client.h
```diff
@@ -25,6 +25,7 @@ limitations under the License.
 #include <photon/common/stream.h>
 #include <photon/common/timeout.h>
 #include <photon/net/socket.h>
+#include <photon/net/security-context/tls-stream.h>
```
lihuiba, line 28: no need to include the header file. just make a forward declaration.

net/http/client.cpp
```diff
@@ -47,6 +47,13 @@ class PooledDialer {
             resolver(new_default_resolver(kDNSCacheLife)) {
     }
 
+    PooledDialer(TLSContext *_tls_ctx) :
```
liulanzheng, line 50: No need to add a new constructor; simply add a default parameter to the existing constructor. The constructor is involved in the initialization of multiple members, and it is not advisable to maintain multiple versions.

net/http/client.cpp
```diff
@@ -110,6 +117,11 @@ class ClientImpl : public Client {
     ClientImpl(ICookieJar *cookie_jar) :
         m_cookie_jar(cookie_jar) {}
 
+    ClientImpl(ICookieJar *cookie_jar, TLSContext *tls_ctx) :
```
liulanzheng, line 120: It is also possible to reuse the above constructor.

net/http/client.cpp
```diff
@@ -107,8 +107,10 @@ class ClientImpl : public Client {
     PooledDialer m_dialer;
     CommonHeaders<> m_common_headers;
     ICookieJar *m_cookie_jar;
-    ClientImpl(ICookieJar *cookie_jar) :
-        m_cookie_jar(cookie_jar) {}
+    ClientImpl(ICookieJar *cookie_jar, TLSContext *tls_ctx) :
+        m_cookie_jar(cookie_jar),
+        m_dialer(tls_ctx) {
```
liulanzheng, line 112: It might be necessary to consider ownership of tls_ctx?

net/http/client.h
WaberZhuang: done

liulanzheng (approved):
LGTM

PR #307, opened 2023-12-15
minor fix for std-compat

beef9999:


rpc/serialize.h
```diff
@@ -105,6 +105,7 @@ namespace rpc
         const char* c_str() const { return cbegin(); }
         std::string_view sv() const { return {c_str(), size() - 1}; }
         std::string to_std() { return std::string(sv()); };
+        bool empty() const { return _len == 0; }
```
lihuiba, line 108: better put it in ```array<T>```

lihuiba (approved):

PR #311, opened 2023-12-21
Inlined defer

lihuiba:
make DEFER() always inline


common/utility.h
```diff
@@ -74,25 +74,25 @@ ptr_array_t<T> ptr_array(T* pbegin, size_t n)
     return {pbegin, pbegin + n};
 }
 
+#define __INLINE__ __attribute__((always_inline))
+#define __FORCE_INLINE__ __INLINE__ inline
+
```
beef9999, line 79: Just wondering why force incline doesn't equal to always inline ...

lihuiba: Plain functions need the ```inline``` keyword, otherwise the compiler will generates warnings. Whereas lambdas can not have ```inline```. So we have to have 2 macros. Maybe the wording can be improved.

PR #319, opened 2023-12-26
ObjectCache with intrusive_list specialization

Coldwings:
ObjectCache specialization for intrusive_list value-type.

Do not need to create a intrusive_list(which is only a node pointer) as ObjectCache value in creator. Store the list as a member of ObjectCache item.

### DIFF
--- a/common/expirecontainer.h
+++ b/common/expirecontainer.h
@@ -202,7 +202,7 @@ public:
     using Base = ExpireContainer<T>;
     using Base::Base;
     using typename Base::Item;
-    bool keep_alive(const T& x, bool insert_if_not_exists) {
+    bool keep_alive(const T &x, bool insert_if_not_exists) {
         return Base::keep_alive(Item(x), insert_if_not_exists);
     }
 };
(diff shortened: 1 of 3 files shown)


common/expirecontainer.h
```diff
@@ -372,3 +373,124 @@ class ObjectCache : public ObjectCacheBase {
...
+    using Base = ObjectCacheBase;
+    using KeyedItem = Base::KeyedItem<Base::Item, KeyType>;
+
+    class Item : public KeyedItem {
+    public:
+        ListType _list;
+        using KeyedItem::KeyedItem;
+        virtual Item* construct() const override {
+            auto item = new Item(this->_key);
+            item->_obj = nullptr;
```
lihuiba, line 390: ```item->_obj = item;``` ?

common/expirecontainer.h
```diff
@@ -372,3 +373,124 @@ class ObjectCache : public ObjectCacheBase {
...
+public:
+    ObjectCache(uint64_t expiration) : Base(expiration, expiration / 16) {}
+    ObjectCache(uint64_t expiration, uint64_t timer_cycle)
+        : Base(expiration, timer_cycle) {}
+
+    template <typename Constructor>
+    ItemPtr ref_acquire(const InterfaceKey& key, const Constructor& ctor,
+                        uint64_t failure_cooldown = 0) {
+        auto _ctor = [&](void* item) {
+            ListType list = ctor();
```
lihuiba, line 412: ```((Item*)item)->_list = ctor();``` ?

common/expirecontainer.h
```diff
@@ -372,3 +373,124 @@ class ObjectCache : public ObjectCacheBase {
         return borrow(key, [] { return new ValEntity(); });
     }
 };
+
+template <typename KeyType, typename NodeType>
+class ObjectCache<KeyType, intrusive_list<NodeType>> : public ObjectCacheBase {
```
lihuiba, line 378: inherit ObjectCache<KeyType, ValPtr>, so that Borrow can be reused.

lihuiba (approved):

common/expirecontainer.h
Coldwings: the `construct` member function should keep _obj as nullptr.

In `ref_acquire` implementation, item will create if not exists, and then check if _obj is empty. The real `ctor` will be called only when `_obj` equals to nullptr

common/expirecontainer.h
Coldwings: intrusive_list did not have a move constructor

common/expirecontainer.h
Coldwings: Borrow in list is not a implemention.
Those override operators are different, so as `Item` type. Inheritance may not help.

HanLee13:
LGTM

Coldwings:
Now list and pointer items sharing same ObjectCache code, all differences are controlled by `Item` type.

HanLee13 (approved):

lihuiba (approved):

PR #325, opened 2024-01-10
iouring: remove RLIMIT_MEMLOCK check for new kernel

beef9999:
> io_uring accounts memory it needs under the rlimit memlocked option, which
> can be quite low on some setups (64K). The default is usually enough for
> most use cases, but bigger rings or things like registered buffers deplete
> it quickly. This affects 5.11 and earlier, new kernels are less dependent
> on RLIMIT_MEMLOCK as it is only used for registering buffers.

### DIFF
--- a/io/iouring-wrapper.cpp
+++ b/io/iouring-wrapper.cpp
@@ -76,10 +76,14 @@ public:
     }
 
     int init() {
-        rlimit resource_limit{.rlim_cur = RLIM_INFINITY, .rlim_max = RLIM_INFINITY};
-        if (setrlimit(RLIMIT_MEMLOCK, &resource_limit) != 0) {
-            LOG_ERROR_RETURN(0, -1, "iouring: failed to set resource limit. Use command `ulimit -l unlimited`, or change to root");
+        int compare_result;
+        if (kernel_version_compare("5.11", compare_result) == 0 && compare_result <= 0) {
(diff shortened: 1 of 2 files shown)


io/iouring-wrapper.cpp
```diff
@@ -76,10 +76,14 @@ class iouringEngine : public MasterEventEngine, public CascadingEventEngine, pub
...
 
     int init() {
-        rlimit resource_limit{.rlim_cur = RLIM_INFINITY, .rlim_max = RLIM_INFINITY};
-        if (setrlimit(RLIMIT_MEMLOCK, &resource_limit) != 0) {
-            LOG_ERROR_RETURN(0, -1, "iouring: failed to set resource limit. Use command `ulimit -l unlimited`, or change to root");
+        int compare_result;
+        if (kernel_version_compare("5.11", compare_result) == 0 && compare_result <= 0) {
+            rlimit resource_limit{.rlim_cur = RLIM_INFINITY, .rlim_max = RLIM_INFINITY};
+            if (setrlimit(RLIMIT_MEMLOCK, &resource_limit) != 0)
+                LOG_ERROR_RETURN(0, -1, "iouring: failed to set resource limit. "
```
lihuiba, line 83: is ERROR_RETURN absolutely needed?

beef9999: Because we need to set a value, and we have no idea what's the appropriate number, based on the ring size.

So a strong check makes things easier, and let users know in advance.

lihuiba (approved):

PR #326, opened 2024-01-10
Add extra options in `photon::init`

Coldwings:
This Pull Request proposes enhancements to the photon::init initialization procedure by introducing the ability to set crucial options directly within or prior to its invocation. To improve user convenience and configurability, the following three new options have been added:

1. Libaio Context Size Configuration: Allows users to specify the size of the libaio context.
2. Pooled Stack Allocator Enablement: Provides an option to utilize a pooled stack allocator for improved memory management.
3. Thread Pool Bypass: Introduces a mechanism to bypass the thread pool when necessary.

In line with this enhancement, the testing suite for the pooled-stack-allocator now utilizes the extended photon::init API, taking advantage of these newly introduced options as part of its setup process. This serves as a proof-of-concept for further extension, indicating that additional configuration parameters may be seamlessly integrated following the same approach.

### DIFF
--- a/io/aio-wrapper.cpp
+++ b/io/aio-wrapper.cpp
@@ -38,10 +38,10 @@ limitations under the License.
 
 namespace photon
 {
-    const uint64_t IODEPTH = 2048;
+    constexpr static int IODEPTH_MAX = 2048;
     struct libaio_ctx_t
     {
-        int evfd = -1, running = 0;
+        int evfd = -1, running = 0, iodepth = 32;
         io_context_t aio_ctx = {0};
         thread* polling_thread = nullptr;
         condition_variable cond;
@@ -152,8 +152,8 @@ namespace photon
     static void resume_libaio_requesters()
     {
 retry:
-        struct io_event events[IODEPTH];
-        int n = HAVE_N_TRY(my_io_getevents, (0, IODEPTH, events));
+        struct io_event events[IODEPTH_MAX];
+        int n = HAVE_N_TRY(my_io_getevents, (0, IODEPTH_MAX, events));
         for (int i=0; i<n; ++i)
         {
             auto piocb = (libaiocb*)events[i].obj;
@@ -166,7 +166,7 @@ retry:
                          VALUE(piocb->u.c.buf), VALUE(piocb->u.c.resfd));
             thread_interrupt((thread *)events[i].data, EOK);
         }
-        if (n == IODEPTH)
+        if (n == IODEPTH_MAX)
         {
             thread_yield();
             goto retry;
@@ -355,7 +355,7 @@ retry:
             }
             io_destroy(libaio_ctx->aio_ctx);
             libaio_ctx->aio_ctx = {0};
-            int ret = io_setup(IODEPTH, &libaio_ctx->aio_ctx);
+            int ret = io_setup(libaio_ctx->iodepth, &libaio_ctx->aio_ctx);
             if (ret < 0) {
                 LOG_ERROR("failed to create aio context by io_setup() ", ERRNO(), VALUE(ret));
                 exit(-1);
@@ -366,17 +366,18 @@ retry:
     };
     static thread_local AioResetHandle *reset_handler = nullptr;
 
-    int libaio_wrapper_init()
+    int libaio_wrapper_init(int iodepth)
     {
         if (libaio_ctx)
             return 0;
 
         std::unique_ptr<libaio_ctx_t> ctx(new libaio_ctx_t);
         ctx->evfd = eventfd(0, EFD_NONBLOCK | EFD_CLOEXEC);
+        libaio_ctx->iodepth = iodepth > IODEPTH_MAX ? IODEPTH_MAX : iodepth;
         if (ctx->evfd < 0)
             LOG_ERRNO_RETURN(0, -1, "failed to create eventfd");
 
-        int ret = io_setup(IODEPTH, &ctx->aio_ctx);
+        int ret = io_setup(ctx->iodepth, &ctx->aio_ctx);
         if (ret < 0)
         {
             LOG_ERROR("failed to create aio context by io_setup() ", ERRNO(), VALUE(ret));
--- a/io/aio-wrapper.cpp
+++ b/io/aio-wrapper.cpp
@@ -153,7 +153,7 @@ namespace photon
     {
 retry:
         struct io_event events[IODEPTH_MAX];
-        int n = HAVE_N_TRY(my_io_getevents, (0, IODEPTH_MAX, events));
+        int n = HAVE_N_TRY(my_io_getevents, (0, libaio_ctx->iodepth, events));
         for (int i=0; i<n; ++i)
         {
             auto piocb = (libaiocb*)events[i].obj;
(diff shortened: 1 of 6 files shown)


photon.h
```diff
@@ -48,17 +48,25 @@ const uint64_t INIT_IO_DEFAULT = INIT_IO_LIBCURL;
 
 #undef SHIFT
 
+struct PhotonOptions {
+    int libaio_ctx = 2048;
```
lihuiba, line 52: uint16_t libaio_ctx = 2048;   // perhaps smaller?

photon.h
```diff
@@ -48,17 +48,25 @@ const uint64_t INIT_IO_DEFAULT = INIT_IO_LIBCURL;
...
+
 /**
  * @brief Initialize the main photon thread and ancillary threads by flags.
  *        Ancillary threads will be running in background.
  * @return 0 for success
  */
 int init(uint64_t event_engine = INIT_EVENT_DEFAULT,
-         uint64_t io_engine = INIT_IO_DEFAULT);
+         uint64_t io_engine = INIT_IO_DEFAULT,
+         const PhotonOptions& options = PhotonOptions());
```
lihuiba, line 64: const PhotonOptions& options = {});

io/aio-wrapper.cpp
```diff
@@ -376,7 +376,7 @@ namespace photon
         if (ctx->evfd < 0)
             LOG_ERRNO_RETURN(0, -1, "failed to create eventfd");
 
-        int ret = io_setup(IODEPTH, &ctx->aio_ctx);
+        int ret = io_setup(iodepth < IODEPTH ? iodepth : IODEPTH, &ctx->aio_ctx);
```
lihuiba, line 379: no need for IODEPTH?

photon.cpp
```diff
@@ -45,7 +48,14 @@ static const int recommended_order[] = {INIT_EVENT_EPOLL, INIT_EVENT_IOURING, IN
 static const int recommended_order[] = {INIT_EVENT_KQUEUE, INIT_EVENT_SELECT};
 #endif
 
-int init(uint64_t event_engine, uint64_t io_engine) {
+int init(uint64_t event_engine, uint64_t io_engine, const PhotonOptions& options) {
+    if (options.use_pooled_stack_allocator) {
+        use_pooled_stack_allocator();
+    }
+    if (options.bypass_threadpool) {
+        set_bypass_threadpool(options.bypass_threadpool);
```
lihuiba, line 56: set_bypass_threadpool(true);

photon.h
lihuiba: without ```&``` may be more efficient for small structs

io/aio-wrapper.cpp
Coldwings: IODEPTH is now the maximal value of iodepth, since the stack allocated event array set to fixed size of IODEPTH.

It could rename to MAX_IODEPTH

photon.h
Coldwings: The `PhotonOptions` might be much larger in future, maybe..?

photon.h
Coldwings: Just keep as same as it before.

It could be much smaller as default value, like 256 or event 64.

io/aio-wrapper.cpp
```diff
@@ -388,6 +391,7 @@ namespace photon
         assert(ctx->polling_thread);
         libaio_ctx = ctx.release();
         libaio_ctx->running = 1;
+        libaio_ctx->iodepth = iodepth;
```
lihuiba, line 394: make value range valid here, avoiding repeating in io_setup()

photon.h
```diff
@@ -48,17 +48,25 @@ const uint64_t INIT_IO_DEFAULT = INIT_IO_LIBCURL;
 
 #undef SHIFT
 
+struct PhotonOptions {
+    int libaio_ctx = 128;
```
lihuiba, line 52: libaio_queue_depth

io/aio-wrapper.cpp
```diff
@@ -152,8 +152,8 @@ namespace photon
     static void resume_libaio_requesters()
     {
 retry:
-        struct io_event events[IODEPTH];
-        int n = HAVE_N_TRY(my_io_getevents, (0, IODEPTH, events));
+        struct io_event events[IODEPTH_MAX];
+        int n = HAVE_N_TRY(my_io_getevents, (0, IODEPTH_MAX, events));
```
EricHuangqx, line 156: should use iodepth here instead of IODEPTH_MAX

io/aio-wrapper.cpp
```diff
@@ -166,7 +166,7 @@ namespace photon
                          VALUE(piocb->u.c.buf), VALUE(piocb->u.c.resfd));
             thread_interrupt((thread *)events[i].data, EOK);
         }
-        if (n == IODEPTH)
+        if (n == IODEPTH_MAX)
```
EricHuangqx, line 169: should use iodepth here instead of IODEPTH_MAX

io/aio-wrapper.cpp
```diff
@@ -366,17 +366,18 @@ namespace photon
...
 
-    int libaio_wrapper_init()
+    int libaio_wrapper_init(int iodepth)
     {
         if (libaio_ctx)
             return 0;
 
         std::unique_ptr<libaio_ctx_t> ctx(new libaio_ctx_t);
         ctx->evfd = eventfd(0, EFD_NONBLOCK | EFD_CLOEXEC);
+        libaio_ctx->iodepth = iodepth > IODEPTH_MAX ? IODEPTH_MAX : iodepth;
```
EricHuangqx, line 376: libaio_ctx->iodepth should be >= 1 ？

io/aio-wrapper.cpp
Coldwings: using `iodepth` means here must be a dynamic sized array, might not be a good idea

io/aio-wrapper.cpp
Coldwings: once iodepth <1, initial will be failed.
just let the initial failure tells user that parameter is nat a valid value.

io/aio-wrapper.cpp
Coldwings: Here always collect at most IODEPTH_MAX events
so the maximal n should equal to event array size and the parameter N calling `my_io_getevents`,
It should be still IODEPTH_MAX

lihuiba (approved):

io/aio-wrapper.cpp
EricHuangqx: int io_setup(unsigned nr_events, aio_context_t *ctx_idp);

io/aio-wrapper.cpp
EricHuangqx: if iodepth < IODEPTH_MAX, n always less than IODEPTH_MAX

io/aio-wrapper.cpp
EricHuangqx: just my_io_getevents use iodepth

EricHuangqx (approved):

PR #328, opened 2024-01-13
Treat compiler warning as error

beef9999:


net/http/headers.cpp
```diff
@@ -146,7 +146,7 @@ HeadersBase::KV* HeadersBase::kv_add_sort(KV kv) {
     if ((char*)(begin - 1) <= m_buf + m_buf_size)
         LOG_ERROR_RETURN(ENOBUFS, nullptr, "no buffer");
     auto it = std::lower_bound(begin, kv_end(), kv, HA(this));
-    memmove(begin - 1, begin, sizeof(KV) * (it - begin));
+    // memmove(begin - 1, begin, sizeof(KV) * (it - begin));
     m_kv_size++;
```
beef9999, line 150: The compiler deems that it's unsafe to overwrite std:pair's memory via pointer. Because of vtable ?

How to rewrite this line to remove warning? @lihuiba

thread/thread.cpp
```diff
@@ -928,9 +928,9 @@ R"(
         size_t randomizer = (rand() % 32) * (1024 + 8);
         stack_size = align_up(randomizer + stack_size + sizeof(thread), PAGE_SIZE);
         char* ptr = (char*)photon_thread_alloc(stack_size);
-        auto p = ptr + stack_size - sizeof(thread) - randomizer;
-        (uint64_t&)p &= ~63;
```
lihuiba, line 932: Pointer arithmetics are straight-forward. No need to change them.

beef9999: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]

lihuiba (approved):

lihuiba (approved):

PR #332, opened 2024-01-17
use Timeout 

lihuiba:
thread, io engines, socket

move sat_add(), sat_sub() into common/utility.h, reimplemented with compiler intrinsic, instead of inline assembly.

update the interface of class Timeout slightly


liulanzheng:
These "fix" commits should ideally be squashed into one.

lihuiba:
> These "fix" commits should ideally be squashed into one.

will be squashed merged

EricHuangqx (approved):

PR #336, opened 2024-01-18
Fix HTTPFSv2 using proxy

Coldwings:
In httpfs V2, give it a client with proxy does not makes all file access going with proxy.

Every request will be reset during fetching requests. But`op.set_enable_proxy()` only marks it using or not using proxy.
Do or do not generate request follows proxy standard is only decided by the third argument in request reset  operation (which has default value `false`)

Fix it by setting correct argument of `reset` call.


liulanzheng (approved):
LGTM

lihuiba (approved):

PR #339, opened 2024-01-23
more Timeout usage

lihuiba:


net/basic_socket.h
```diff
@@ -101,36 +105,48 @@ __FORCE_INLINE__ int doio(IOCB iocb, WAIT waitcb) {
...
-    while (v.iovcnt > 0) {
-        ssize_t ret = iocb();
-        if (ret < 0) return ret;
-        if (ret == 0) break;
-        count += ret;
-
-        uint64_t bytes = ret;
-        auto extracted = v.extract_front(bytes);
-        assert(extracted == bytes);
+struct BufAdv {
```
beef9999, line 122: Adv looks like `add v` but the actual meaning we are going to express is `advance` ?

net/basic_socket.h
```diff
@@ -101,36 +105,48 @@ __FORCE_INLINE__ int doio(IOCB iocb, WAIT waitcb) {
...
+    BufAdv(void*& buf, size_t& count) : buf(buf), count(count) { }
+    BufAdv(size_t& count) : buf(_dummy), count(count) { }
+    bool operator()(size_t ret, size_t n) __INLINE__ {
+        assert(0 <= ret && ret <= count);
+        (char*&)buf += ret;
+        count -= ret;
+        return count > 0;
+    }
+};
+
```
beef9999, line 135: If V stands for vector

net/basic_socket.h
lihuiba: Yes, and better naming is welcome.

net/basic_socket.h
lihuiba: Yes, and better naming is welcome.

Coldwings (approved):
LGTM

net/basic_socket.h
beef9999: BufAdvance

lihuiba: how about step?

PR #343, opened 2024-01-30
socket api do not distinguish IPv4 or IPv6

lihuiba:
socket api do not distinguish IPv4 or IPv6
refinement to bind() function series

### DIFF
--- a/net/kernel_socket.cpp
+++ b/net/kernel_socket.cpp
@@ -72,7 +72,7 @@ static int socket(int family, int protocol = 0,
         set_fd_nonblocking(fd);
     if (nodelay) {
         int v = 1;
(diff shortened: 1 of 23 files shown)


net/kernel_socket.cpp
```diff
@@ -195,26 +202,23 @@ class KernelSocketClient : public SocketClientBase {
 
     ISocketStream* connect(const EndPoint& remote, const EndPoint* local) override {
         sockaddr_storage r(remote);
-        if (likely(!local)) {
+        if (likely(!local || local->is_ipv4() != remote.is_ipv4())) {
```
beef9999, line 205: Why we need this check?

lihuiba: In case that you bind to an IPv4 address while connecting to an IPv6 address, or vice versa. This is ridiculous.

Coldwings (approved):
LGTM

PR #351, opened 2024-02-06
FIX httpfs_v2 common header for file able to replace by ioctl

Coldwings:
This pull request addresses an issue with the ioctl set header functionality in `httpfs_v2`. Currently, `httpfs_v2` treats ioctl headers differently than httpfs, by appending records instead of handling them as a key-value map.

The proposed changes aim to align the behavior of httpfs_v2 with httpfs, ensuring that ioctl headers are treated consistently across both implementations. This will improve compatibility and make it easier for developers to work with these file systems.

### DIFF
--- a/fs/httpfs/httpfs_v2.cpp
+++ b/fs/httpfs/httpfs_v2.cpp
@@ -103,7 +103,7 @@ public:
 class HttpFile_v2 : public fs::VirtualReadOnlyFile {
 public:
     std::string m_url;
-    net::http::CommonHeaders<> m_common_header;
+    std::unordered_map<std::string, std::string> m_common_header;
     HttpFs_v2* m_fs;
     struct stat m_stat;
     uint64_t m_stat_gettime = 0;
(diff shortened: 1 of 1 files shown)


liulanzheng (approved):
LGTM

fs/httpfs/httpfs_v2.cpp
```diff
@@ -103,7 +103,7 @@ class HttpFs_v2 : public fs::IFileSystem {
 class HttpFile_v2 : public fs::VirtualReadOnlyFile {
 public:
     std::string m_url;
-    net::http::CommonHeaders<> m_common_header;
+    std::unordered_map<std::string, std::string> m_common_header;
```
lihuiba, line 106: use _string_keyed map instead.

lihuiba:
Why this consistency across both implementations is important?

Coldwings:
> Why this consistency across both implementations is important?

Both `httpfs` and `httpfs_v2` rely solely on the same ioctl command (`HTTP_HEADER`) to set HTTP headers. This shared method emphasizes the need for consistency between the two implementations.

In scenarios where headers are employed to set authorization information, such as in the case of Object Storage Services, the ability to reset headers becomes crucial. Maintaining the same logic for the ioctl `HTTP_HEADER` in both implementations not only streamlines the process but also eliminates any confusion surrounding the usage of HTTP_HEADER. This consistency ultimately benefits users who require flexibility in setting, updating, or resetting their HTTP headers according to their specific needs.

lihuiba (approved):

PR #352, opened 2024-02-06
Fix IPAddress packed layout and iovector flex array

Coldwings:
In issue #350 

Shows new version GCC performs more check for old GNU extensions, and treats union as non-POD member in struct layout.

Here some fixes for them, tested worked in GCC 13.2 and fits the requirements of C++ standard:

1. Use flex array instead of deprecated zero-sized array which causes warning.
2. Prevent inherent to a struct with flex-array.
3. Add attribute for union in `IPAddr`, make `IPAddr` struct able to be packed

### DIFF
--- a/common/iovector.h
+++ b/common/iovector.h
@@ -236,6 +236,10 @@ inline void delete_iovector(iovector* ptr);
 
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Warray-bounds"
+#if __GNUC_MINOR__ >= 13
+#pragma GCC diagnostic ignored "-Wzero-length-bounds"
+#pragma GCC diagnostic ignored "-Wstrict-flex-arrays"
+#endif
 
 class iovector
 {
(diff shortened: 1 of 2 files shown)


common/iovector.h
```diff
@@ -827,7 +828,7 @@ class iovector
     uint16_t capacity;              // total capacity
     uint16_t iov_begin, iov_end;    // [iov_begin, iov_end)
     uint16_t nbases;                // # of allocated pointers
-    struct iovec iovs[0];
```
lihuiba, line 830: [0] is a common practice, and it's more read-able than iovs_ptr(). We can simple ignore the warning with ```#pragma``

Coldwings: It is a common solution in GNU C extension.
C99 standard add new define of flex array, `struct iovec iovs[]` to present a flex array.
though, C++ does not allow inheritance by a class with tailed flex array, that is why here replaced by iovs_ptr()

Seems there is not a better solution than `iovs[0]`

lihuiba (approved):

PR #361, opened 2024-02-18
The instantiation of sockaddr will lose information

beef9999:
禁止sockaddr的实例化

### DIFF
--- a/net/base_socket.h
+++ b/net/base_socket.h
@@ -21,6 +21,7 @@ Internal header provides abstract socket base class
 ***/
 
 #include <netinet/in.h>
+#include <sys/un.h>
 #include <vector>
 
 #include <photon/net/socket.h>
@@ -60,8 +61,8 @@ struct sockaddr_storage {
     explicit sockaddr_storage(const sockaddr_in6& addr) {
         *((sockaddr_in6*) &store) = addr;
     }
-    explicit sockaddr_storage(const sockaddr& addr) {
-        *((sockaddr*) &store) = addr;
+    explicit sockaddr_storage(const sockaddr_un& addr) {
+        *((sockaddr_un*) &store) = addr;
     }
     EndPoint to_endpoint() const {
         EndPoint ep;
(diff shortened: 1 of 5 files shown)


net/socket.h
```diff
@@ -37,7 +37,6 @@ LogBuffer& operator << (LogBuffer& log, const in_addr& iaddr);
 LogBuffer& operator << (LogBuffer& log, const sockaddr_in& addr);
 LogBuffer& operator << (LogBuffer& log, const in6_addr& iaddr);
 LogBuffer& operator << (LogBuffer& log, const sockaddr_in6& addr);
-LogBuffer& operator << (LogBuffer& log, const sockaddr& addr);
```
beef9999, line 40: In most case, `sockaddr` is converted from `sockaddr_storage`, and is always suppose to be used as a pointer.
Print its object will lose information.

net/base_socket.h
```diff
@@ -60,8 +61,8 @@ struct sockaddr_storage {
     explicit sockaddr_storage(const sockaddr_in6& addr) {
         *((sockaddr_in6*) &store) = addr;
     }
-    explicit sockaddr_storage(const sockaddr& addr) {
-        *((sockaddr*) &store) = addr;
+    explicit sockaddr_storage(const sockaddr_un& addr) {
+        *((sockaddr_un*) &store) = addr;
     }
```
beef9999, line 66: `sockaddr_storage` now supports UDS address.

net/base_socket.h
```diff
@@ -60,8 +61,8 @@ struct sockaddr_storage {
     explicit sockaddr_storage(const sockaddr_in6& addr) {
         *((sockaddr_in6*) &store) = addr;
     }
-    explicit sockaddr_storage(const sockaddr& addr) {
-        *((sockaddr*) &store) = addr;
```
beef9999, line 64: `sockaddr` is always supposed to be used as a pointer. Otherwise might lose information.

lihuiba (approved):

PR #363, opened 2024-02-19
fix throttle 0

beef9999:


common/throttle.h
```diff
@@ -30,6 +31,11 @@ class throttle {
 public:
     throttle(uint64_t limit, uint64_t time_window = 1000UL * 1000,
              uint64_t slice = 10) {
+        if (limit == 0) {
```
lihuiba, line 34: treat -1UL as unlimited?

common/throttle.h
```diff
@@ -12,6 +12,7 @@ class throttle {
     uint64_t m_time_window;
     uint64_t m_time_slice;
     uint64_t m_slice_num;
+    bool m_no_limit;
```
Coldwings, line 15: SInce limit is uint64_t unsigned int, using -1UL(UINT64_MAX) to present un-throttled is a designed solution, while `limit = 0` means cut-off control.

PR #364, opened 2024-02-20
ALog Prologue now keeps only last part of source path (filename), using us timestamps

Coldwings:
This pull request introduces a change to the ALog Prologue module. The modification made is to keep only the last part of the source path, specifically the filename, while utilizing microsecond timestamps.

Microsecond timestamps shares photon thread timestamp `photon::now`.

### DIFF
--- a/common/alog.h
+++ b/common/alog.h
@@ -343,7 +343,7 @@ protected:
 
 struct Prologue
 {
-    uint64_t addr_func, addr_file;
+    const char *addr_func, *addr_file;
     int len_func, len_file;
     int line, level;
 };
@@ -448,8 +448,8 @@ struct LogBuilder {
     auto _partial_file =                                                       \
         ConstString::TSpliter<'/', ' ',                                        \
                               decltype(_prologue_file_r)>::Current::reverse(); \
-    const static Prologue prolog{(uint64_t)_prologue_func,                     \
-                                 (uint64_t)_partial_file.chars,                \
+    const static Prologue prolog{_prologue_func,                               \
+                                 _partial_file.chars,                          \
                                  sizeof(__func__) - 1,                         \
                                  (int)_partial_file.len,                       \
                                  __LINE__,                                     \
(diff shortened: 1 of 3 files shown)


thread/thread.cpp
```diff
@@ -1090,6 +1090,12 @@ R"(
         }
         return photon::now;
     }
+    void alog_update_now(uint64_t *sec, uint64_t *usec) {
+        last_tsc = _rdtsc();
+        auto ts = update_now();
+        *sec = ts / 1000000;
+        *usec = ts % 1000000;
+    }
     int timestamp_updater_init() {
```
beef9999, line 1099: Why not move to alog.cpp?

Coldwings: `update_now()` `_rdtsc()` `last_tsc` are defined as inline or static unit in thread.cpp, `alog_update_now()` needs to call those functions and set `last_tsc` `now` so time update can also used by thread scheduler.

common/alog.h
```diff
@@ -525,6 +539,12 @@ inline NamedValue<T> make_named_value(const char (&name)[N], T&& value)
 
 #define VALUE(x) make_named_value(#x, x)
 
+template <size_t N>
+inline LogBuffer& operator<<(LogBuffer& log,
```
lihuiba, line 543: better specialize ```make_named_value()```

thread/thread.cpp
```diff
@@ -1090,6 +1090,12 @@ R"(
         }
         return photon::now;
     }
+    void alog_update_now(uint64_t *sec, uint64_t *usec) {
```
lihuiba, line 1093: better return sec and usec directly as an std::pair, so that they are returned in registers.
```cpp
#include <stdio.h>
#include <inttypes.h>
#include <utility>

std::pair<uint64_t, uint64_t> foo();


int main(int argc, char** argv) {
	auto x = foo();
	return (x.first + x.second) >> 3;
}
```

sec/usec are returned in rax/rdx as a result:
```
(gdb) disassemble main
Dump of assembler code for function main:
   0x0000000000001050 <+0>:	sub    $0x8,%rsp
   0x0000000000001054 <+4>:	callq  0x1030 <_Z3foov@plt>
   0x0000000000001059 <+9>:	add    $0x8,%rsp
   0x000000000000105d <+13>:	add    %rdx,%rax
   0x0000000000001060 <+16>:	shr    $0x3,%rax
   0x0000000000001064 <+20>:	retq
End of assembler dump.
```

lihuiba (approved):

thread/thread.cpp
Coldwings: updated: return `struct timeval` as result

lihuiba (approved):

PR #365, opened 2024-02-20
fix semaphore interrupt

beef9999:


Coldwings:
It's great that the semaphore wait has been made interruptible, it would be more consistent if it allowed for any errno to be used for interruption, just like other synchronization components such as mutex and condition variable.

I think #380 offers a better solution by making all internal wake-ups use a non-common errno (-1) as the interrupt reason, allowing any other interrupt errno to be considered as a user interrupt.

PR #380, opened 2024-02-27
Semaphore now able to be interrupted

Coldwings:
This pull request improves the synchronization components based on `waitq` by adding support for interruption using `thread_interrupt`. The semaphore component, which was previously unable to be interrupted, is now included in this enhancement.

Interruption of a running photon thread can be beneficial in specific situations, such as during the graceful shutdown of a process. This update enables more efficient and controlled termination of threads, resulting in improved system reliability and performance.

Additionally, the internal photon thread interruption errno used by `waitq` has been changed to -1 by default. As a result, users can now utilize all generic error numbers to interrupt synchronization components.

### DIFF
--- a/thread/thread.h
+++ b/thread/thread.h
@@ -398,15 +398,14 @@ namespace photon
     {
     public:
         explicit semaphore(uint64_t count = 0) : m_count(count) { }
-        int wait(uint64_t count, Timeout timeout = {});
-        int wait_uninterruptible(uint64_t count, Timeout timeout = {}) {
+        int wait(uint64_t count, Timeout timeout = {}) {
             int ret = 0;
             do {
-                ret = wait(count, timeout);
-            } while (!timeout.expired() &&
-                     (ret == 0 || (ret < 0 && errno == ESHUTDOWN)));
+                ret = wait_interruptible(count, timeout);
+            } while (ret < 0 && (errno != ESHUTDOWN && errno != ETIMEDOUT));
             return ret;
         }
+        int wait_interruptible(uint64_t count, Timeout timeout = {});
         int signal(uint64_t count)
         {
             if (count == 0) return 0;
--- a/thread/thread.h
+++ b/thread/thread.h
@@ -212,9 +212,9 @@ namespace photon
     protected:
         int wait(Timeout timeout = {});
         int wait_defer(Timeout Timeout, void(*defer)(void*), void* arg);
-        void resume(thread* th, int error_number = ECANCELED);  // `th` must be waiting in this waitq!
-        int resume_all(int error_number = ECANCELED);
-        thread* resume_one(int error_number = ECANCELED);
+        void resume(thread* th, int error_number = -1);  // `th` must be waiting in this waitq!
+        int resume_all(int error_number = -1);
+        thread* resume_one(int error_number = -1);
         waitq() = default;
         waitq(const waitq& rhs) = delete;   // not allowed to copy construct
         waitq(waitq&& rhs) = delete;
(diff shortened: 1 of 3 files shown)


thread/thread.h
```diff
@@ -362,10 +362,30 @@ namespace photon
         {
             return waitq::wait(timeout);
         }
+        template<typename LOCK, typename PRED>
+        int wait_pred(LOCK& lock, PRED&& pred, Timeout timeout = {}) {
```
lihuiba, line 366: simply name it as ```wait```?

thread/thread.h
```diff
@@ -362,10 +362,30 @@ namespace photon
...
+            return do_wait_pred([&]{ return wait(lock, timeout); }, std::forward<PRED>(pred), timeout);
+        }
+        template<typename PRED>
+        int wait_pred_no_lock(PRED&& pred, Timeout timeout = {}) {
```
lihuiba, line 370: simply name it as ```wait_no_lock```?

thread/thread.h
```diff
@@ -362,10 +362,30 @@ namespace photon
...
+            while (!pred()) {
+                ret = do_wait();
+                err = errno;
+            }
+            errno = err;
+            return ret;
+        }
     };
 
     class semaphore : protected waitq
```
lihuiba, line 391: add a ```wait_uninterruptible()```?

thread/thread.h
Coldwings: change to `wait` by using SFINAE

thread/thread.h
Coldwings: change to `wait_no_lock` by using SFINAE

thread/thread.h
Coldwings: done

thread/thread.h
```diff
@@ -212,9 +212,9 @@ namespace photon
     protected:
         int wait(Timeout timeout = {});
         int wait_defer(Timeout Timeout, void(*defer)(void*), void* arg);
-        void resume(thread* th, int error_number = ECANCELED);  // `th` must be waiting in this waitq!
-        int resume_all(int error_number = ECANCELED);
-        thread* resume_one(int error_number = ECANCELED);
+        void resume(thread* th, int error_number = -1);  // `th` must be waiting in this waitq!
+        int resume_all(int error_number = -1);
+        thread* resume_one(int error_number = -1);
         waitq() = default;
```
beef9999, line 218: What is the default behavior of the default value -1. Documents should be adjusted in:

https://github.com/alibaba/PhotonLibOS/blob/main/doc/docs/api/thread.md#thread_interrupt

https://github.com/alibaba/PhotonLibOS/blob/main/doc/i18n/cn/docusaurus-plugin-content-docs/current/api/thread.md#thread_interrupt

lihuiba: If it comes from the inside of the synchronization primitives due to technical reasons, we use -1 to distinguish from user generated ordinary error numbers (>0). And 0 has already been used for another scenario.

thread/thread.h
```diff
@@ -362,17 +362,52 @@ namespace photon
         {
             return waitq::wait(timeout);
         }
+        template <typename LOCK, typename PRED,
+                  typename = typename std::enable_if<std::is_same<
+                      decltype(std::declval<PRED>()()), bool>::value>::type>
```
lihuiba, line 367: as long as PRED is not typed as Timeout

lihuiba: and enable_if_t is shorter

thread/thread.h
Coldwings: > What is the default behavior of the default value -1 ? Documents should be adjusted in:
> 
> https://github.com/alibaba/PhotonLibOS/blob/main/doc/docs/api/thread.md#thread_interrupt
> 
> https://github.com/alibaba/PhotonLibOS/blob/main/doc/i18n/cn/docusaurus-plugin-content-docs/current/api/thread.md#thread_interrupt

Those documents are about `photon::thread_interrupt`. The default errno in `photon::thread_interrupt` is not changed, so it still EINTR, nothing has changed in document.

Here is all about method `waitq::resume`, which is protected method, not a part of API.

thread/thread.h
Coldwings: if do not limit PRED return value type, even `enable_if` part is meaningless.

Once PRED is callable (means PRED()() has a type) is good enough.

thread/thread.h
```diff
@@ -362,17 +362,51 @@ namespace photon
...
+            errno = err;
+            return ret;
+        }
     };
 
     class semaphore : protected waitq
     {
     public:
         explicit semaphore(uint64_t count = 0) : m_count(count) { }
         int wait(uint64_t count, Timeout timeout = {});
```
lihuiba, line 401: Is there any invocations of wait() that should be changed to invocations of wait_uninterruptible()?

Coldwings: Make `wait` uninterruptible, keeps it works like it used to be, and add a new method `wait_interruptible` to present a wait that able to be interrupt might be better.

lihuiba (approved):

PR #389, opened 2024-03-02
Fix endpoint undefined situation

beef9999:
The `undefined` has caused many troubles. Shall we use the rest 16bits padding of Endpoint as a new flag to indicate it's undefined?  @lihuiba @Coldwings 

Extra alignment are no longer needed as well.


Coldwings (approved):
LGTM

Issue #398, opened 2024-03-07
EPOLL: Why not set EPOLLONESHOT in add_interest when ONESHOT is set?


KiventD:
In https://github.com/alibaba/PhotonLibOS/blob/main/io/epoll.cpp#L144, there is comment that "since epoll oneshot shows totally different meanning of ONESHOT in photon all epoll action keeps no oneshot". 
I do not understand what it means. Can anything go wrong if setting `EPOLLONESHOT` to underlying epoll_ctl's flags when `ONESHOT` is set in photon?

lihuiba:
EPOLLONESHOT only disable the event interest of the fd, leaving the registration inside the epoll's internal data structure after an event.

Whereas photon's one-shot is stateless, leaving nothing in the engine after an event. So we delete the registration from epoll  right after an event.

lihuiba:
I made a study to find out that, closing an fd will silently remove possible registration from epoll:
https://stackoverflow.com/questions/43034535/do-i-get-a-notification-from-epoll-when-a-fd-is-closed

The potential leakage of using EPOLLONESHOT in photon is perhaps acceptable.

beef9999:
Solved by https://github.com/alibaba/PhotonLibOS/pull/410

PR #399, opened 2024-03-08
[Discuss] Add register_fd for epoll

KiventD:
This PR mainly modifies the use of epoll, including 4 parts:

1. Always set `EPOLLONESHOT` in `epoll_ctl` when `ONESHOT` is set in photon. As is disscussed in https://github.com/alibaba/PhotonLibOS/issues/398. This change is meant to help address part2 issue.

2. https://github.com/alibaba/PhotonLibOS/pull/388 shows a performance issue when using epoll in a multi-thread program. This PR tries to fix it by allowing user to avoid frequent `EPOLL_CTL_ADD`:
    1. Offer user a new API to register a fd
    2. When `rm_interest` is going to delete the fd from epfd, just do nothing instead if the fd is registered (no more events of fd will be reported by epfd, because `EPOLLONESHOT` is set)
    3. Further `add_interest` will always use `EPOLL_CTL_MOD` rather than `EPOLL_CTL_ADD`
    4. Unregister the fd will `EPOLL_CTL_DEL` it

3. `entry` in `_inflight_events` will not be modified until the whole `add_interest/rm_interest` succeeds, preventing dirty data of `entry` in some failure cases

4. Combine several `rm_interest` in `wait_for_events`, concerned that:
    1. Reduce overhead of multiple calls of `epoll_ctl`
    2. Consider such unexpected case if no combination: fd is registered as part1 says. `EVENT_READ` and `EVENT_WRITE` is reported at the same time by `epoll_wait`. When `rm_interest` `EVENT_READ`, it finds that there is still `EVENT_WRITE` in `entry`'s interests, so `EPOLL_CTL_MOD` is used. After that, when `rm_interest` `EVENT_WRITE`, nothing will be done because no interest is in `entry`. Now it is expected that no events of this fd will be reported, but actually `EVENT_WRITE` is still there.

I don't know if it is necessary to `rm_interest` event individually (i.e. NO combination). If it is, I need to find another way to address issue 4.2

Finally, the new APIs `register_fd` and `unregister_fd` is not that good-designed, as it works only for epoll. Better APIs may be needed. So this PR is not formal and the main goal is looking for more discussion and suggestion.

### DIFF
--- a/io/epoll.cpp
+++ b/io/epoll.cpp
@@ -47,6 +47,8 @@ struct InFlightEvent {
     void* reader_data;
     void* writer_data;
     void* error_data;
+    bool registered = false;
+    bool in_epfd = false;
 };
 
 class EventEngineEPoll : public MasterEventEngine, public CascadingEventEngine, public ResetHandle {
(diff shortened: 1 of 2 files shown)


lihuiba:
Thanks for your PR. It seems to me that ```register_fd()``` is not necessary, as you know in ```add_interests()``` whether an fd has been added in epoll or not, and use proper operation in epoll_ctl().

Correct me if I miss something.

lihuiba:
Besides close(fd), you can also delete the event interest(s) from epoll when you ```wait_for_fd()``` with an empty event set.

beef9999:
Solved by https://github.com/alibaba/PhotonLibOS/pull/410

PR #400, opened 2024-03-10
Add SNI support

benwaffle:
I needed SNI support in the TLS implementation for use in overlaybd, to connect to my container registry. - https://github.com/containerd/overlaybd/issues/317

This change works, you can test it with the example program: `./output/sni`. It should probably be rewritten as a test.

It would be nice to not have to call an extra function, but `tlssock->connect()` takes an endpoint, which only has the IP address and not the hostname. I initially tried to create a method, like `tlssock->set_hostname(...)`, but the class hierarchy made it difficult. I didn't want to move around any classes to not break the API. So I chose to create a separate method, `tls_socket_set_hostname`, which uses `dynamic_cast`.

### DIFF
--- a/net/http/client.cpp
+++ b/net/http/client.cpp
@@ -73,6 +73,7 @@ ISocketStream* PooledDialer::dial(std::string_view host, uint16_t port, bool sec
     if (secure) {
         tlssock->timeout(timeout);
         sock = tlssock->connect(ep);
+        tls_socket_set_hostname(sock, strhost.c_str());
     } else {
         tcpsock->timeout(timeout);
         sock = tcpsock->connect(ep);
--- a/net/security-context/tls-stream.cpp
+++ b/net/security-context/tls-stream.cpp
@@ -411,7 +411,8 @@ ISocketStream* new_tls_stream(TLSContext* ctx, ISocketStream* base,
 
 void tls_socket_set_hostname(ISocketStream *stream, const char *hostname) {
     if (auto s = dynamic_cast<TLSSocketStream*>(stream)) {
-        s->set_hostname(hostname);
+        if (!s->set_hostname(hostname))
+            LOG_ERROR("Failed to set hostname on tls stream: ", VALUE(stream), VALUE(hostname));
     } else if (auto s = dynamic_cast<ForwardSocketStream*>(stream)) {
         auto underlay = static_cast<ISocketStream*>(s->get_underlay_object(0));
         tls_socket_set_hostname(underlay, hostname);
--- a/net/security-context/tls-stream.cpp
+++ b/net/security-context/tls-stream.cpp
@@ -28,6 +28,7 @@ limitations under the License.
 #include <vector>
 
 #include "../base_socket.h"
+#include "photon/common/alog.h"
 
 namespace photon {
 namespace net {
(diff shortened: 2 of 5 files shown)


beef9999:
@benwaffle Noted. Will take a look at it.

BTW, you seemed to have delved into Photon and overlaybd quite a bit. Would you like to share some experience?

net/security-context/tls-stream.h
```diff
@@ -16,6 +16,7 @@ limitations under the License.
 
 #pragma once
 
+#include <photon/net/socket.h>
```
lihuiba, line 19: There is forward declaration for ISocketStream already. Is it still necessary to include this?

beef9999:
@benwaffle Can you authorize push permission for me to your own Photon repo? I want to make some adjustments to your code.

net/http/client.cpp
```diff
@@ -73,6 +73,7 @@ ISocketStream* PooledDialer::dial(std::string_view host, uint16_t port, bool sec
     if (secure) {
         tlssock->timeout(timeout);
         sock = tlssock->connect(ep);
+        tls_socket_set_hostname(sock, strhost.c_str());
```
Coldwings, line 76: Make SNI check optional seems to be better. User should able to set enable or disable SNI check.

Add a method to set by `Operation` or `HTTPClient` should be a good idea.

benwaffle: Is there a reason not to use SNI?

Go's HTTP client, curl, and all modern browsers use SNI by default (which means docker and containerd do as well).

benwaffle:
@beef9999 you should already have write permission to my `sni` branch

net/http/client.cpp
lihuiba: @Coldwings Is there any drawbacks to always use SNI? Compatibility with older versions of OpenSSL?

Coldwings: > @Coldwings Is there any drawbacks to always use SNI? Compatibility with older versions of OpenSSL?

SNI extension and `SSL_set_tlsext_host_name` API is provided in OpenSSL 1.1.x but not exists in OpenSSL 1.0.2 versions. SNI feature is still supported in OpenSSL 1.0.2 but should be set by `SSL_ctrl`.

> Is there a reason not to use SNI all the time?
> 
> Go's HTTP client, curl, and all modern browsers use SNI by default (which means docker and containerd do as well).

Enable SNI by default is good. I just think that as a common library, provide a option to disable SNI might be useful when user needs.

net/security-context/tls-stream.cpp
```diff
@@ -339,6 +340,11 @@ class TLSSocketStream : public ForwardSocketStream {
         SSL_free(ssl);
     }
 
+    bool set_hostname(const char* hostname) {
+        int ret = SSL_set_tlsext_host_name(ssl, hostname);
```
Coldwings, line 344: `SSL_set_tlsext_host_name` is not provided in OpenSSL 1.0.2 series API. The code here still should considering compatibility to OpenSSL 1.0.2.

it is better to use macro check `OPENSSL_VERSION_NUMBER` and use `SSL_ctrl` to set host name when version lower than 1.1.0.

beef9999:
Closed. Continue review in https://github.com/alibaba/PhotonLibOS/pull/401

PR #424, opened 2024-03-23
adaptor for ISocketStream to std::iostream

lihuiba:
This adaptor can be used in scenarios such as redis-cpp:

https://github.com/tdv/redis-cpp


Coldwings (approved):
`ISocketStream` as `iostream`. LGTM.

Since `IStream` not provides partial read/write, only `ISocketStream` can fit iostream interface.

PR #425, opened 2024-03-24
integration for photon and cpp_redis

lihuiba:


Coldwings (approved):
Adaptor from `photon::net::ISocketStream` to `cpp_redis::network::tcp_client_iface` which used by cppredis.
LGTM

PR #427, opened 2024-03-25
a simple promise/future for coordination with background worker thread(s)

lihuiba:
This promise doesn't fully mimic std::promise, so as to allow for a possibility of stack allocation of the the objects.


Coldwings (approved):
A simple tool making async result in promise/future pattern. LGTM.

PR #434, opened 2024-03-31
openssl not use locking after 1.1.0

beef9999:
Going to backport to 0.7

### DIFF
--- a/fs/httpfs/httpfs_v2.cpp
+++ b/fs/httpfs/httpfs_v2.cpp
@@ -281,9 +281,9 @@ IFileSystem* new_httpfs_v2(bool default_https, uint64_t conn_timeout,
                          client, client_ownership);
 }
 
-IFile* new_httpfile_v2(const char* url, HttpFs_v2* httpfs, uint64_t conn_timeout,
+IFile* new_httpfile_v2(const char* url, IFileSystem* httpfs, uint64_t conn_timeout,
                     uint64_t stat_timeout) {
-    return new HttpFile_v2(url, httpfs, conn_timeout, stat_timeout);
+    return new HttpFile_v2(url, (HttpFs_v2*) httpfs, conn_timeout, stat_timeout);
 }
 }  // namespace fs
 }
--- a/net/kernel_socket.cpp
+++ b/net/kernel_socket.cpp
@@ -139,6 +139,7 @@ public:
         return (Object*) (uint64_t) fd;
     }
     int close() final {
+        if (fd < 0) return 0;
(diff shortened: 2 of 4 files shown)


fs/httpfs/httpfs_v2.cpp
```diff
@@ -281,9 +281,9 @@ IFileSystem* new_httpfs_v2(bool default_https, uint64_t conn_timeout,
                          client, client_ownership);
 }
 
-IFile* new_httpfile_v2(const char* url, HttpFs_v2* httpfs, uint64_t conn_timeout,
+IFile* new_httpfile_v2(const char* url, IFileSystem* httpfs, uint64_t conn_timeout,
```
beef9999, line 284: mismatch to header file

net/kernel_socket.cpp
```diff
@@ -139,7 +139,8 @@ class KernelSocketStream : public SocketStreamBase {
         return (Object*) (uint64_t) fd;
     }
     int close() final {
-        get_vcpu()->master_event_engine->wait_for_fd(fd, 0, -1UL);
```
lihuiba, line 142: if (fd < 0) return 0;

net/security-context/tls-stream.cpp
```diff
@@ -76,17 +76,21 @@ class GlobalSSLContext : public Singleton<GlobalSSLContext> {
         OpenSSL_add_all_ciphers();
         OpenSSL_add_all_digests();
         OpenSSL_add_all_algorithms();
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
```
lihuiba, line 79: Is there a function to obtain openssl version dynamically? If so, it's better to use ```if``` instead of ```#if```.

lihuiba:
```c
 int SSL_client_version(const SSL *s);

 const char *SSL_get_version(const SSL *ssl);

 int SSL_is_dtls(const SSL *ssl);

 int SSL_version(const SSL *s);
```
Some of the variants seem to return numerical version?

https://www.openssl.org/docs/man1.1.1/man3/SSL_get_version.html

lihuiba: According to https://github.com/openssl/openssl/blob/OpenSSL_1_1_0-stable/crypto/cversion.c

we can use ```SSLeay()``` for versions before 1.1, and ```OpenSSL_version_num()``` for those after 1.1.

lihuiba: We can probe for the above functions with ```dlopen()``` and ```dlsymbol()```, and decide which one to use.

PR #439, opened 2024-04-02
retval<T> to return a value and an error number at the same time

lihuiba:
so as to avoid some bugs:

* unintentionally change the global errno by some function calls (e.g. printf or LOG_XXXX) in the error handling path,  causing misunderstanding about the error

* return an error but forget to set errno, causing incorrect handling of the error

In addition, ```retval<T>``` makes the error number returned in register, more efficient than the global thread-local ```errno``` variable.

### DIFF
--- a/common/alog.h
+++ b/common/alog.h
@@ -26,6 +26,7 @@ limitations under the License.
 
 #include <photon/common/utility.h>
 #include <photon/common/conststr.h>
+#include <photon/common/retval.h>
 
 class ILogOutput {
 protected:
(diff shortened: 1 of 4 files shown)


common/alog.h
```diff
@@ -508,6 +510,22 @@ struct ERRNO
...
+    return x ? (log << ERRNO((int)x)) : log;
+}
+
+template<typename T> inline
+LogBuffer& operator << (LogBuffer& log, const photon::retval<T>& v) {
+    return v.succeeded() ? (log << v.get()) : (log << v.base());
+}
+
+template<> inline
+LogBuffer& operator << <void> (LogBuffer& log, const photon::retval<void>& v) {
```
Coldwings, line 525: Log on A successful retval<void> breaks assertion in `LogBuffer::operator<<(retval_base)`

common/alog.h
```diff
@@ -508,6 +510,22 @@ struct ERRNO
 
 LogBuffer& operator << (LogBuffer& log, ERRNO e);
 
+inline LogBuffer& operator << (LogBuffer& log, const photon::retval_base& rvb) {
+    auto x = rvb._errno;
+    assert(0<x && x<INT_MAX);
```
Coldwings, line 515: assertion rejects x ==0 situation.

Coldwings (approved):
LGTM

PR #445, opened 2024-04-05
SimpleDOM support for XML (with RapidXml) and JSON (with rapidjson)

lihuiba:


Coldwings (commented):
Put third-party sources like rapidxml and rapidjson code in project is not a good choice.

We will not and should not maintain those codes.

Using cmake `FetchContent` or git submodule will be better.

ecosystem/test/test.cpp
```diff
@@ -0,0 +1,23 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <gtest/gtest.h>
+
+
+int main(int argc, char **argv) {
```
Coldwings, line 20: Seems this empty test caused SIGBUS in ARM.

What is the purpose of this empty test program?

lihuiba: it actually links with test_simple.dom.cpp (and potentially other .cpp files in the future).

lihuiba:
> Put third-party sources like rapidxml and rapidjson code in project is not a good choice.

Actually it's the recommended way of both libraries:

```To use it you just need to copy rapidxml.hpp file to a convenient place (such as your project directory),``` (https://rapidxml.sourceforge.net/manual.html)

```Just copy the include/rapidjson folder to system or project's include path. ``` (https://github.com/Tencent/rapidjson)

ecosystem/test/test.cpp
Coldwings: Got it . Just like `gtest_main`.

Coldwings:
> > Put third-party sources like rapidxml and rapidjson code in project is not a good choice.
> 
> Actually it's the recommended way of both libraries:
> 
> "To use it you just need to copy rapidxml.hpp file to a convenient place (such as your project directory)" (https://rapidxml.sourceforge.net/manual.html)
> 
> "Just copy the include/rapidjson folder to system or project's include path. " (https://github.com/Tencent/rapidjson)
> 
> And I have made some changes to rapidjson to fit for SimpleDOM.

It is able to make use of those header only libs, but still, copy files to project makes future upgrade/fix of third-party deps all by "copy again".

Even there is some changes or adaption, make own fork of `rapidjson` or add a patch file may be much more easier to maintain.

beef9999:
> > > Put third-party sources like rapidxml and rapidjson code in project is not a good choice.
> > 
> > 
> > Actually it's the recommended way of both libraries:
> > "To use it you just need to copy rapidxml.hpp file to a convenient place (such as your project directory)" (https://rapidxml.sourceforge.net/manual.html)
> > "Just copy the include/rapidjson folder to system or project's include path. " (https://github.com/Tencent/rapidjson)
> > And I have made some changes to rapidjson to fit for SimpleDOM.
> 
> It is able to make use of those header only libs, but still, copy files to project makes future upgrade/fix of third-party deps all by "copy again".
> 
> Even there is some changes or adaption, make own fork of `rapidjson` or add a patch file may be much more easier to maintain.

Agreed. CMake add git repo + patch is better, if the patch is not huge.

lihuiba:
@beef9999 @Coldwings 
* The latest release of rapidjson was in 2016, and RapidXml in 2009. They seem to be quite stable and unlikely to get further updates often.
* In case of new updates, copying again will not be big problems.
* Considering the threat of supply-chain attacks like the recent ```xz``` event, we should have to review any incoming updates from the dependent libs. This will be the major effort.

Coldwings (approved):
Now ecosystem third-party headers are download by `FetchContent`.

LGTM

PR #447, opened 2024-04-08
possiable to add a copyable future ?

cageq:
c++11 future is copyable, using a shared memory to hold the state.  

take a look?

### DIFF
--- a/thread/future.h
+++ b/thread/future.h
@@ -16,40 +16,59 @@ limitations under the License.
 
 #pragma once
 #include <utility>
+#include <memory>
 #include <photon/thread/thread.h>
 #include <photon/thread/awaiter.h>
 
 namespace photon {
 
 template<typename T, typename Context = AutoContext>
-class Future {
+class Future   {
     static_assert(sizeof(T) > 0, "in the case that T is void, simply use Awaiter<Context> or semaphore instead!");
 
-    T _value;
-    Awaiter<Context> _awaiter;
-    bool _got = false;
+    struct FutureState{
+        T _value;
+        Awaiter<Context> _awaiter;
+        bool _got = false;
+        ~FutureState() {
+            assert(_got); // supposed to be assigned only once
+        }
+    }; 
+
+    using FutureStatePtr = std::shared_ptr<FutureState>;
+    FutureStatePtr _state;     
 
     template<typename P>
     void set_value(P&& rhs) {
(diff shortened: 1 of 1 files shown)


lihuiba:
> c++11 future is copyable, using a shared memory to hold the state.\

Actually I'm not quite agree with the design of std::future/promise, because:
* It implicitly uses heap-allocated object (with shared_ptr), and I believe the allocation is not alway acceptable in all scenarios. There are cases where stack-allocation is preferred.
* The caller has to obtain first the promise, then the future from it. This process seems unnecessary, as the caller doesn't have any interests on the promise.

So I have made a different design:
* It doesn't use heap-allocation by itself to allow for best performance, and it can be used in conjunction with shared_ptr if necessary.
* It's future-centric, and the promise is subordinate to the future.

lihuiba:
And I understand that the difference in design may have brought troubles to you, as you have probably got used to std::promise/future. I think it's possible that we make an wrapper in std-compat.h ?

cageq:
> > c++11 future is copyable, using a shared memory to hold the state.\
> 
> Actually I'm not quite agree with the design of std::future/promise, because:
> 
> * It implicitly uses heap-allocated object (with shared_ptr), and I believe the allocation is not alway acceptable in all scenarios. There are cases where stack-allocation is preferred.
> * The caller has to obtain first the promise, then the future from it. This process seems unnecessary, as the caller doesn't have any interests on the promise.
> 
> So I have made a different design:
> 
> * It doesn't use heap-allocation by itself to allow for best performance, and it can be used in conjunction with shared_ptr if necessary.
> * It's future-centric, and the promise is subordinate to the future.

Yes, If performance is the first consideration, best to use stack-allocation, then use it with shared_ptr. 

I used it in photon::WorkPool's async_call, to wait the async_call's result  since it ignore the return value(best return future object I think), so I need a shared_ptr wrapper if use stack version . It's ok for me .  

 And wrapper a copyable future into std-compat.h is acceptable. 
  
"The caller has to obtain first the promise"  just for the set_value method of future is private in origin code, I didn't change that. 

Thanks, currently I just made the future can work with my code, wish it can be full implement.

lihuiba:
@cageq take a look at https://github.com/alibaba/PhotonLibOS/pull/448

cageq:
> @cageq take a look at #448

ok， thanks， I'll keep eye on it .

PR #448, opened 2024-04-09
std_compat promise/future

lihuiba:
Wrap photon::Future/Promise as std_compat::future/promise, which is copyable/movable, and similar to std ones.

### DIFF
--- a/common/timeout.h
+++ b/common/timeout.h
@@ -52,11 +52,8 @@ public:
             m_expiration = x;
         return *this;
     }
-    using us = std::chrono::microseconds;
-    template<class Rep, class Period = std::ratio<1>>
-    Timeout(const std::chrono::duration<Rep, Period>& x) :
-        Timeout(std::chrono::duration_cast<us>(x).count()) { }
     auto std_duration() const {
+        using us = std::chrono::microseconds;
         uint64_t max = std::numeric_limits<us::rep>::max();
         return (m_expiration > max) ? us::max() : us(timeout());
     }
--- a/common/timeout.h
+++ b/common/timeout.h
@@ -52,8 +52,11 @@ public:
             m_expiration = x;
         return *this;
     }
+    using us = std::chrono::microseconds;
+    template<class Rep, class Period = std::ratio<1>>
+    Timeout(const std::chrono::duration<Rep, Period>& x) :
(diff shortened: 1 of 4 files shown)


thread/std-compat.h
```diff
@@ -349,6 +350,66 @@ int work_pool_init(int vcpu_num = 1, int event_engine = photon::INIT_EVENT_DEFAU
...
+    }
+
+    template< class Rep, class Period >
+    std::future_status wait_for( const std::chrono::duration<Rep,Period>& duration ) const {
+        return wait(__duration_to_microseconds(duration));
+    }
+
+    template< class Clock, class Duration >
+    std::future_status wait_until( const std::chrono::time_point<Clock,Duration>& timeout_time ) const {
+        return wait_for(timeout_time - std::chrono::system_clock::now());
```
Coldwings, line 375: once timeout_time is before current time point, might should not wait.

Since `timeout_time` takes `Clock` template parameter, it should compare to `Clock::now` as well, instead of `system_clock`.

common/timeout.h
```diff
@@ -52,8 +52,11 @@ class Timeout {
             m_expiration = x;
         return *this;
     }
+    using us = std::chrono::microseconds;
+    template<class Rep, class Period = std::ratio<1>>
+    Timeout(const std::chrono::duration<Rep, Period>& x) :
```
Coldwings, line 57: since duration is a signed value, Timeout to duration define might be not quite well-defined.
Timeout from a negative duration should consider as nearly infinite timeout or as 0 timeout might not definite here.

thread/std-compat.h
```diff
@@ -349,6 +350,67 @@ int work_pool_init(int vcpu_num = 1, int event_engine = photon::INIT_EVENT_DEFAU
...
+    }
+
+    template< class Rep, class Period >
+    std::future_status wait_for( const std::chrono::duration<Rep,Period>& duration ) const {
+        return wait(__duration_to_microseconds(duration));
+    }
+
+    template< class Clock, class Duration >
+    std::future_status wait_until( const std::chrono::time_point<Clock,Duration>& timeout_time ) const {
+        auto dt = timeout_time - std::chrono::system_clock::now();
```
Coldwings, line 375: `auto dt = timeout_time - Clock::now()` would be better.

PR #460, opened 2024-04-18
fix warnings in CI, as well as bugs

lihuiba:


common/lockfree_queue.h
```diff
@@ -472,7 +472,7 @@ class LockfreeSPSCRingQueue : public LockfreeRingQueueBase<T, N> {
         n = std::min(n, tail.load(std::memory_order_acquire) - h);
         if (n == 0) return 0;
         auto first_idx = idx(h);
-        auto last_idx = idx(h + n - 1);
+        auto last_idx = idx(h + n - 1); (void)last_idx;
```
Coldwings, line 475: Looks OK just simply remove this line.

common/string-keyed.h
```diff
@@ -325,7 +325,7 @@ class basic_map_string_kv : public M
     const_iterator end() const
     {
         auto it = base::end();
-        return (B_IT&)it;
+        return __reinterpret_cast<B_IT>(it);
```
Coldwings, line 328: why not simply use
`reinterpret_cast<B_IT&>(t)` instead?

lihuiba: the ```reinterpret_cast``` keyword doesn't support this use

Coldwings (approved):
LGTM

PR #464, opened 2024-04-22
Add mprotect for the top-end of photon stack

Coldwings:
Add a extra page on the top-end of photon stack and set to `mprotect` `PROT_NONE`.

When code access this page, will through a segment fault immediately, so user will notice that there might be stack overflow.

### DIFF
--- a/thread/thread.cpp
+++ b/thread/thread.cpp
@@ -305,6 +305,7 @@ namespace photon
             assert(state == states::DONE);
             // `buf` and `stack_size` will always store on register
             // when calling deallocating.
+            mprotect(buf, PAGE_SIZE, PROT_READ | PROT_WRITE);
             photon_thread_dealloc(buf, stack_size);
         }
     };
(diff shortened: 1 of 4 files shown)


thread/timer.h
```diff
@@ -35,7 +35,7 @@ namespace photon
         // it has a `stack_size`, and the `on_timer` is invoked within the thread's context.
         // The timer object is deleted automatically after it is finished.
         Timer(uint64_t default_timeout, Entry on_timer, bool repeating = true,
-              uint64_t stack_size = 1024 * 64)
+              uint64_t stack_size = 8UL * 1024 * 1024)
```
lihuiba, line 38: DEFAULT_STACK_SIZE

thread/thread.cpp
```diff
@@ -934,13 +935,14 @@ R"(
         if (unlikely(!rq.current))
             LOG_ERROR_RETURN(ENOSYS, nullptr, "Photon not initialized in this vCPU (OS thread)");
         size_t randomizer = (rand() % 32) * (1024 + 8);
-        stack_size = align_up(randomizer + stack_size + sizeof(thread), PAGE_SIZE);
+        stack_size = align_up(randomizer + stack_size + sizeof(thread) + PAGE_SIZE, PAGE_SIZE);
         char* ptr = (char*)photon_thread_alloc(stack_size);
+        mprotect(ptr, PAGE_SIZE, PROT_NONE);
```
lihuiba, line 940: is ```ptr``` guaranteed to be aligned to PAGE_SIZE?

lihuiba (approved):

thread/thread.cpp
Coldwings: > is `ptr` guaranteed to be aligned to PAGE_SIZE?

That is a problem. I will add extra pages allocation and make  protect page aligned.

PR #478, opened 2024-05-10
http client destroy_operation

beef9999:


CMakeLists.txt
```diff
@@ -96,12 +104,6 @@ if (NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE Release)
 endif ()
 
-# If ccache exists, use it to speed up compiling
-find_program(CCACHE_FOUND ccache)
-if(CCACHE_FOUND)
-    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
-endif(CCACHE_FOUND)
-
```
beef9999, line 104: ccache is not safe enough, especially when switching from debug to release.

Besides, CMake has already supported incremental compile.

net/http/client.h
```diff
@@ -92,16 +101,10 @@ class Client : public Object {
               enable_proxy(c->has_proxy()),
               _client(c) {}
         Operation(uint16_t buf_size) : req(_buf, buf_size) {}
+        ~Operation() = default;
+        friend void Client::destroy_operation(Operation* op);
```
beef9999, line 105: Allow client -> destroy_operation(), while forbid calling delete op

lihuiba (approved):

PR #479, opened 2024-05-15
support customized UserAgent in http client

liulanzheng:
merge into release/0.6, as overlaybd is still using release/0.6 and there is a demand from overlaybd users for customizing user-agent.

### DIFF
--- a/net/http/client.cpp
+++ b/net/http/client.cpp
@@ -224,9 +224,10 @@ public:
             LOG_ERROR_RETURN(EINVAL, ROUNDTRIP_FAILED,
                             "Content-Length and Transfer-Encoding conflicted");
         }
-        op->req.headers.insert("User-Agent", USERAGENT);
-        op->req.headers.insert("Connection", "keep-alive");
         op->req.headers.merge(m_common_headers);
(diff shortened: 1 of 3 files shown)


net/http/client.cpp
```diff
@@ -224,9 +224,9 @@ class ClientImpl : public Client {
             LOG_ERROR_RETURN(EINVAL, ROUNDTRIP_FAILED,
                             "Content-Length and Transfer-Encoding conflicted");
         }
-        op->req.headers.insert("User-Agent", USERAGENT);
-        op->req.headers.insert("Connection", "keep-alive");
         op->req.headers.merge(m_common_headers);
+        op->req.headers.insert("User-Agent", m_user_agent.empty() ? USERAGENT : m_user_agent);
```
lihuiba, line 228: m_user_agent.empty() ? string_view(USERAGENT) : string_view(m_user_agent)

lihuiba: otherwise, USERAGENT may be converted (constructed) to std::string, then string_view

lihuiba (approved):

lihuiba (approved):

PR #488, opened 2024-05-22
add option for ecosystem in cmake

liulanzheng:


Coldwings (approved):
LGTM

Patch in ecosystem may modify rapidjson.
These option able to prevent including a modified rapidjson.

PR #494, opened 2024-05-28
new CI workflow for linux.x86

lihuiba:
* compile with multiple gcc versions: 8.5.0, 9.2.1, 10.3.1, 11.2.1, 12.1.1, and test all their outputs
* limit # of concurrent compilation to $(nproc), to prevent OOM
* split `test-thread` into 2 separate test suites, for ease of handling event engine overriding (see below)
* enable tests for executor, memory-stream, etc.
* enable macos arm tests (on github-hosted runners)
* possibility to _override_ event engine in **CI** with environment variable PHOTON_CI_EV_ENGINE, for example:
```
export PHOTON_CI_EV_ENGINE=io_uring
ctest
```
and it outputs:
```
...
INFO |th=0000000000000000|ci-tools.cpp:71|init:argument specified: epoll, io_uring, select (7)
INFO |th=0000000000000000|ci-tools.cpp:72|init:environment specified: 'io_uring'
INFO |th=0000000000000000|ci-tools.cpp:76|init:event engine overridden to: io_uring
INFO |th=0000000000C70D80|iouring-wrapper.cpp:668|new_iouring:Init event engine: iouring [is_master=1]
...
```

image


Coldwings (approved):
LGTM

beef9999:
Don't merge to 0.7, but master instead?

.github/workflows/ci.linux.arm.yml
```diff
@@ -37,12 +37,12 @@ jobs:
           source /opt/rh/gcc-toolset-9/enable
           cmake -B build -D CMAKE_BUILD_TYPE=MinSizeRel -D PHOTON_BUILD_TESTING=ON \
             -D PHOTON_ENABLE_SASL=ON -D PHOTON_ENABLE_FUSE=ON -D PHOTON_ENABLE_EXTFS=ON
-          cmake --build build -j -- VERBOSE=1
+          cmake --build build -j $(nproc) -- VERBOSE=1
```
beef9999, line 40: What's the default value of nproc? Didn't see the definition.

CMake/build-from-src.cmake
```diff
@@ -43,8 +43,8 @@ function(build_from_src [dep])
                 UPDATE_DISCONNECTED ON
                 BUILD_IN_SOURCE ON
                 CONFIGURE_COMMAND ./configure --prefix=${BINARY_DIR}
-                BUILD_COMMAND sh -c "V=1 CFLAGS=\"-fPIC -g -O3 -Wall -Wextra -fno-stack-protector\" $(MAKE) -C src"
-                INSTALL_COMMAND $(MAKE) install
+                BUILD_COMMAND sh -c "V=1 CFLAGS=\"-fPIC -O3 -Wall -Wextra -fno-stack-protector\" $(MAKE) -C src"
+                # INSTALL_COMMAND $(MAKE) install
```
beef9999, line 47: Keep this line?

examples/CMakeLists.txt
```diff
@@ -1,21 +1,21 @@
 add_definitions(-w)
 
 add_executable(simple-example simple/simple.cpp)
-target_link_libraries(simple-example PRIVATE photon_static)
+target_link_libraries(simple-example PRIVATE photon_shared)
```
beef9999, line 5: We'd better not change static libs to shared libs, because we usually need to copy example programs to other machines to run tests. An independent binary is convenient .

.github/workflows/ci.linux.arm.yml
lihuiba: it's the number of processors in current system.

PR #503, opened 2024-06-06
fix dead lock in ExportPerformer

HanLee13:
When using an exported fs，ExportPerformer put called function to a SPSCRingQueue and wait to be waked when finished.   Related code is below:

```
        R perform(IF* _if, Func func, ARGS...args)
        {
            AsyncWaiter<R> w;
            (_if->*func)(args..., w.done(), _timeout);
            return w.wait();
        }
```

In situations where `_if->*func` is unimplemented, the function ` _if->*func` will call `w.done()` directly, which cause a dead lock beacuse it try to lock a locked mutex.

This pr do the following things: 
1. fix the dead lock bug by moving the lock action in ExportPerformer from its construciton function to wait() function.
2. Add an unittest `ExportFS.no_xattr_sync` to verify this dead lock behavior.
3. Implement `fallocate` fuction of ExportAsAsyncFile because this method is widely used in fuse.

Besides, I marked the unittest ExportFS.xattr_sync as DISABLED for now. Because this test failed on MacOS only when compiled as release. Hope someone can fix it.

### DIFF
--- a/fs/async_filesystem.cpp
+++ b/fs/async_filesystem.cpp
@@ -139,7 +139,6 @@ namespace fs
             typename AsyncResult<R>::result_type ret;
             int err = 0;
 
-            AsyncWaiter() { }
             int on_done(AsyncResult<R>* r)
             {
                 std::lock_guard<std::mutex> lock(_mtx);
--- a/fs/async_filesystem.cpp
+++ b/fs/async_filesystem.cpp
@@ -134,13 +134,12 @@ namespace fs
         struct AsyncWaiter
         {
             std::mutex _mtx;
-            std::unique_lock<std::mutex> _lock;
             std::condition_variable _cond;
             bool _got_it = false;
             typename AsyncResult<R>::result_type ret;
             int err = 0;
 
-            AsyncWaiter() : _lock(_mtx) { }
+            AsyncWaiter() { }
             int on_done(AsyncResult<R>* r)
             {
                 std::lock_guard<std::mutex> lock(_mtx);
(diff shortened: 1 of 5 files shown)


fs/test/test_exportfs.cpp
```diff
@@ -104,6 +108,8 @@ int callbackvoid(void*, AsyncResult<void>* ret) {
     }
 
 TEST(ExportFS, basic) {
+    photon_init();
```
lihuiba, line 111: why not doing init in main()?

fs/async_filesystem.cpp
```diff
@@ -134,13 +134,12 @@ namespace fs
...
         {
             std::mutex _mtx;
-            std::unique_lock<std::mutex> _lock;
             std::condition_variable _cond;
             bool _got_it = false;
             typename AsyncResult<R>::result_type ret;
             int err = 0;
 
-            AsyncWaiter() : _lock(_mtx) { }
+            AsyncWaiter() { }
```
lihuiba, line 142: no need for a ctor anymore

lihuiba:
How about PR to 0.7 as well?

fs/test/test_exportfs.cpp
HanLee13: Other tests like ExportFS.no_xattr_sync run in pthread rather than photon thread.

fs/async_filesystem.cpp
HanLee13: done

HanLee13:
> How about PR to 0.7 as well?

will do it after this PR is approved.

lihuiba (approved):

lihuiba (approved):

PR #507, opened 2024-06-08
PHOTON_ENABLE_ECOSYSTEM defaults to off

beef9999:


.github/workflows/ci.linux.x86-64.yml
```diff
@@ -29,7 +30,7 @@ jobs:
                         #  -D PHOTON_GOOGLETEST_SOURCE=""   \
                         #  -D PHOTON_URING_SOURCE=https://github.com/axboe/liburing/archive/refs/tags/liburing-2.3.tar.gz
 
-          cmake --build build -j $(nproc) --clean-first -- VERBOSE=1
+          cmake --build build -j --clean-first -- VERBOSE=1
```
beef9999, line 33: cpu num is already known in CMake

lihuiba (approved):

PR #518, opened 2024-06-25
Fix on stack allocator

Coldwings:
When slog is empty, should alloc memory with size fit in slot.

### DIFF
--- a/thread/stack-allocator.cpp
+++ b/thread/stack-allocator.cpp
@@ -98,7 +98,8 @@ public:
         auto ptr = slots[idx].get();
         if (unlikely(!ptr.first)) {
             // slots[idx] empty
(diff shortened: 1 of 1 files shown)


thread/stack-allocator.cpp
```diff
@@ -96,9 +96,10 @@ class PooledStackAllocator {
             return __alloc(size);
         }
         auto ptr = slots[idx].get();
+        auto aligned_size = MIN_ALLOCATION_SIZE << idx;
```
lihuiba, line 99: why not put it in the scope of the following if

Coldwings: seems good

lihuiba (approved):

PR #519, opened 2024-06-25
Remove stream-messenger/messenger.h from include

beef9999:


thread/std-compat.h
```diff
@@ -293,8 +293,8 @@ class condition_variable : public photon::condition_variable {
             return cv_status::no_timeout;
         // We got a timeout when measured against photon's internal clock,
         // but we need to check against the caller-supplied clock to tell whether we should return a timeout.
-        if (Clock::now() < t)
-            return cv_status::no_timeout;
+        // if (Clock::now() < t)
+        //     return cv_status::no_timeout;
         return cv_status::timeout;
     }
```
beef9999, line 299: photon timeout may return earlier than we expected. Don't need to compare with steady clock.

lihuiba (approved):

PR #521, opened 2024-06-27
improve object cache expire efficiency

beef9999:


common/expirecontainer.cpp
```diff
@@ -38,24 +38,28 @@ auto ExpireContainerBase::find(const Item& key_item) -> iterator {
...
         SCOPED_LOCK(_lock);
         _list.split_by_predicate([&](Item* x) {
             bool ret = x->_timeout.expiration() < photon::now;
             if (ret) _set.erase(x);
             return ret;
         });
-    }).delete_all();
+    }
+    _list.delete_all();
     return 0;
```
beef9999, line 63: I think the so called 'Statements and Declarations in Expressions' brings nothing but vagueness, and it's not c++ standard.

For example

```
struct A {
    void f () {    
        cout << this << endl;
    }
};
int main() {
    A a;
    cout << &a << endl;
    ({a;}).f();
    // Address not same
}
```

common/expirecontainer.cpp
```diff
@@ -38,24 +38,28 @@ auto ExpireContainerBase::find(const Item& key_item) -> iterator {
...
     }
+    for (auto x : tmp)
+        delete x;
 }
 
 uint64_t ExpireContainerBase::expire() {
-    ({
+    if (_lifespan > UINT64_MAX / 2)
+        return 0;
+    {
```
beef9999, line 54: if timeout == -1UL, just pass.

thread/timer.h
```diff
@@ -39,6 +39,7 @@ namespace photon
         {
             _on_timer = on_timer;
             _default_timeout = default_timeout;
+            _reset_timeout = -1UL;
```
beef9999, line 42: https://github.com/alibaba/PhotonLibOS/blob/7c597803832bef0a11b20ba9b8e5f2aa41b539a9/thread/thread.cpp#L1459

common/expirecontainer.cpp
Coldwings: It called "Compound Statement Expressions" or "Statement Expression", and it is a long history feature in almost every C/C++ compiler such as GCC/Clang/MSVC/Intel CC/IBM Compiler...  It is a kind of "De Facto Standard"

Statement expression is widely used in a lot of famous projects like
Linux kernel
https://github.com/torvalds/linux/blob/5bbd9b249880dba032bffa002dd9cd12cd5af09c/drivers/most/core.c#L89

SPDK
https://github.com/spdk/spdk/blob/a44a9620a68f2a8c29db1b7165d67fe7c070458c/lib/env_ocf/ocf_env.h#L65

liburing
https://github.com/axboe/liburing/blob/7b3245583069bd481190c9da18f22e9fc8c3a805/src/lib.h#L32

It is hard to say "brings nothing but vagueness".

Since `({a;})` means a expression, the result have to be a pr-value (temporary object) with type `A`, it will be a new object differ from `a`.

A purely C++ way is using inline lambda evaluation instead of statement expression. The code in example should basically equivalent to `[&]{return a;}().f();`, or with less type deducing, `[&]()->A {return a;}().f();`

common/expirecontainer.h
```diff
@@ -210,6 +210,8 @@ class ExpireList : public ExpireContainer<T> {
     }
 };
 
+////////////////////////////////////////////////////////////////////////////////
```
Coldwings, line 213: what is this for?

thread/timer.h
Coldwings: In timer stub, retry use `_reset_timeout` only when timer function returns a negative result. User should not interrupt timer with EAGAIN in any cases, but only call `Timer::reset(new_timeout)` method.

The `_reset_timeout` only be set when `reset(new_timeout)` is called. Set in constructor is meaningless.

common/expirecontainer.h
```diff
@@ -374,9 +376,9 @@ class __ObjectCache : public ObjectCacheBase {
     using ValEntity = typename Item::ValEntity;
     using Borrow = typename Base::Borrow<__ObjectCache>;
 
-    __ObjectCache(uint64_t expiration) : Base(expiration, expiration / 16) {}
-    __ObjectCache(uint64_t expiration, uint64_t timer_cycle)
-        : Base(expiration, timer_cycle) {}
+    explicit __ObjectCache(uint64_t lifespan) : Base(lifespan, lifespan / 16, true) {}
+    __ObjectCache(uint64_t lifespan, uint64_t timer_cycle, bool expire_if_possible)
```
Coldwings, line 380: It is better to keep API stable.

two-parameter constructor should not be break, and leaves same action as it used to be.

common/expirecontainer.cpp
Coldwings: Statement expression is a important in Photon ALog, and cannot be substitute by lambda.

https://github.com/alibaba/PhotonLibOS/blob/7c597803832bef0a11b20ba9b8e5f2aa41b539a9/common/alog.h#L462

common/expirecontainer.h
beef9999: Split expire container and object cache

Coldwings: Why split? `ObjectCache` is a derived class of `ExpireContainerBase`.

Maybe put all base classes to front and let `ExpireContainer` `ObjectCache` stay together is a better idea.

common/expirecontainer.cpp
```diff
@@ -38,24 +40,31 @@ auto ExpireContainerBase::find(const Item& key_item) -> iterator {
...
 
 uint64_t ExpireContainerBase::expire() {
-    ({
+    if (_lifespan > UINT64_MAX / 2)
+        return 0;
+    intrusive_list<Item> expired;
+    {
         SCOPED_LOCK(_lock);
-        _list.split_by_predicate([&](Item* x) {
+        auto tmp = _list.split_by_predicate([&](Item* x) {
```
Coldwings, line 59: `_list.split_by_predicate(...)` will produce a new `intrusive_list` object, but `intrusive_list` has no move constructor or move assignment. That makes both `tmp` and pr-value of method result point to same list.

That makes pr-value `_list.split_by_predicate(...)` not empty after assignment, and then as a temporary object, goes to destruct, meets the assertion about intrusive_list `__node == nullptr`.

beef9999: Shall we add move semantics to `intrusive_list`, or just clean its pointer?

Coldwings (approved):
LGTM

PR #522, opened 2024-07-09
objcachev2

Coldwings:
A simpler but effective object cache implementation.

`ObjectCacheV2` shows better performance compare to `ObjectCache`, with several
improvements:

1. Less lock in both read and write.
2. Object destruction always goes in background photon thread called reclaimer.
3. Self adjustable timeout for reclaimer. No needs to set cycler timer.
4. New `update` API, able to immediately substitute objects.
5. `ObjectCacheV2` no longer support acquire/release API.

It should work as `ObjectCache` in code do not depends on acquire/release API.

### DIFF
--- a/common/objectcachev2.h
+++ b/common/objectcachev2.h
@@ -110,14 +110,14 @@ protected:
     }
 
     template <typename KeyType>
-    Box* __find_or_create_item(const KeyType& key) {
+    Box* __find_or_create_item(KeyType&& key) {
         Box keyitem(key);
         auto pkey = &keyitem;
         Box* item = nullptr;
         SCOPED_LOCK(maplock);
         auto it = map.find(pkey);
         if (it == map.end()) {
-            item = new Box(key);
+            item = new Box(std::forward<KeyType>(key));
             map.emplace(item);
         } else
             item = *it;
@@ -217,7 +217,7 @@ public:
 
     template <typename KeyType>
     Borrow borrow(KeyType&& key) {
-        return borrow(&key, [&]() { return new V(); });
+        return borrow(std::forward<KeyType>(key), [&]() { return new V(); });
     }
 
     template <typename KeyType, typename Ctor>
--- a/common/objectcachev2.h
+++ b/common/objectcachev2.h
@@ -1,0 +1,256 @@
+#pragma once
+
+#include <photon/common/alog.h>
+#include <photon/common/lockfree_queue.h>
+#include <photon/photon.h>
+#include <photon/thread/list.h>
+#include <photon/thread/thread11.h>
+#include <photon/thread/timer.h>
+#include <photon/thread/workerpool.h>
+
+#include <memory>
+#include <unordered_set>
+
+/**
+
+A simpler but effective object cache implementation.
+
+`ObjectCacheV2` shows better performance compare to `ObjectCache`, with several
+improvements:
+
+1. Less lock in both read and write.
+2. Object destruction always goes in background photon thread called reclaimer.
+3. Self adjustable timeout for reclaimer. No needs to set cycler timer.
+4. New `update` API, able to immediately substitute objects.
+5. `ObjectCacheV2` no longer support acquire/release API.
+
+It should work as `ObjectCache` in code do not depends on acquire/release API.
+
+**/
+
+template <typename K, typename VPtr>
+class ObjectCacheV2 {
+protected:
+    using V = std::remove_pointer_t<VPtr>;
+
+    struct Box : public intrusive_list_node<Box> {
+        const K key;
+        std::shared_ptr<V> ref;
+        photon::spinlock boxlock;
+        photon::spinlock reflock;
+        uint64_t lastcreate = 0;
+        uint64_t timestamp = 0;
+
+        struct Updater {
+            Box* box = nullptr;
+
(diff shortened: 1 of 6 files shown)


common/objectcachev2.h
```diff
@@ -0,0 +1,257 @@
...
+        struct Updater {
+            Box* box = nullptr;
+
+            ~Updater() {}
+
+            // Returned RefPtr is the old one
+            // other reader will get new one after updated
+            std::shared_ptr<V> update(V* val, uint64_t ts = 0,
+                                      std::shared_ptr<V>* newptr = nullptr) {
+                auto r = std::shared_ptr<V>(val);
```
lihuiba, line 52: Can you manage to integrate shared_ptr into ctor, so that there's a chance to use make_shared to improve performance.

common/objectcachev2.h
```diff
@@ -0,0 +1,257 @@
...
+            return a->key == b->key;
+        }
+    };
+
+    photon::thread* reclaimer = nullptr;
+    photon::common::RingChannel<
+        LockfreeMPMCRingQueue<std::shared_ptr<V>*, 4096>>
+        reclaim_queue;
+    photon::spinlock maplock;
+    std::unordered_set<Box*, ItemHash, ItemEqual> map;
```
lihuiba, line 99: is it possible to use ```unordered_set<Box, ItemHash, ItemEqual>```?

common/objectcachev2.h
```diff
@@ -0,0 +1,257 @@
...
+    intrusive_list<Box> cycle_list;
+    uint64_t lifespan;
+    photon::Timer _timer;
+
+    std::shared_ptr<V> __update(Box* item, V* val) {
+        std::shared_ptr<V> ret;
+        reclaim_queue.template send<PhotonPause>(new std::shared_ptr<V>(
```
lihuiba, line 106: split it into two statements.

lihuiba: ```new std::shared_ptr<V>(...)``` looks evil. Try to avoid it.

common/objectcachev2.h
```diff
@@ -0,0 +1,257 @@
...
+                }
+            }
+        });
+        photon::thread_enable_join(reclaimer);
+    }
+
+    ~ObjectCacheV2() {
+        _timer.stop();
+        if (reclaimer) {
+            reclaim_queue.template send<PhotonPause>(nullptr);
```
lihuiba, line 250: is it necessary to use the ```template``` keyword?

rpc/test/test.cpp
```diff
@@ -481,7 +481,8 @@ TEST_F(RpcTest, passive_shutdown) {
 
     // The passive shutdown took 3 seconds, until client closed the connection
     GTEST_ASSERT_GT(duration, 2900);
-    GTEST_ASSERT_LT(duration, 3200);
+    // Since GH macos arm CI is slow, we allow 3.5 secs
+    GTEST_ASSERT_LT(duration, 3500);
```
lihuiba, line 485: I'm moving it to release/0.7

Coldwings: Greate

common/objectcachev2.h
Coldwings: It is necessary.

Calling template member function needs a `template` prefix. The `send<P>` member function has default template argument `ThreadPause`, but here has to call with argument `PhotonPause`, makes here have to be with a `template` prefix.

common/objectcachev2.h
Coldwings: Lockfree queue accepts only POD types, a `shared_ptr<T>` object cannot sen via lockfree queue.

Maybe I can try to specialize a lockfree queue with `shared_ptr<T>` type.

common/objectcachev2.h
Coldwings: Here accept plain `V*` type pointer is for compatibility, the `ctor` in `ObjectCache` returns only simple plain pointer.

It seems good to have a new specialization accept `ctor` returns shared_ptr directly.

common/objectcachev2.h
lihuiba: We have following code in RPC's exmaple:
```cpp
    ExampleServer()
        : skeleton(photon::rpc::new_skeleton()),
          server(photon::net::new_tcp_socket_server()) {
        skeleton->register_service<Testrun, Heartbeat, Echo, ReadBuffer,
                                   WriteBuffer>(this);
    }
```
We call the template member function, ```register_service<...>``` without a ```template``` keyword.

Coldwings: Yeah but those code in RPC is a part of normal class.

Here is in template define.

ignore that `template` prefix will leads to erros

```
error: expected primary-expression before ‘>’ token
```
in GCC and

```
Use 'template' keyword to treat 'send' as a dependent template name (fix available)clang(missing_dependent_template_keyword)
```

in clang

common/objectcachev2.h
Coldwings: It is possible, the code might be a little tricky but tested work.

common/objectcachev2.h
Coldwings: Done

lihuiba (approved):

PR #527, opened 2024-07-18
Make PooledStackAllocator implemention more clear, and fix

Coldwings:
refactoring.

fix allocation larger that max slot size.

### DIFF
--- a/thread/stack-allocator.cpp
+++ b/thread/stack-allocator.cpp
@@ -37,7 +37,11 @@ class PooledStackAllocator {
         __builtin_ffsl(MAX_ALLOCATION_SIZE / MIN_ALLOCATION_SIZE);
 
 public:
-    PooledStackAllocator() {}
+    PooledStackAllocator() {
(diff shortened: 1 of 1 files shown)


thread/stack-allocator.cpp
```diff
@@ -37,7 +37,11 @@ class PooledStackAllocator {
         __builtin_ffsl(MAX_ALLOCATION_SIZE / MIN_ALLOCATION_SIZE);
 
 public:
-    PooledStackAllocator() {}
+    PooledStackAllocator() {
+        for (size_t i = 0; i < N_SLOTS; i++) {
+            slots[i].slotsize = MIN_ALLOCATION_SIZE << i;
```
lihuiba, line 42: This is  constant to a slot, why should we store it in a variable?

Coldwings: `Slot::get` needs allocation with size, and it likes a property of a slot.

lihuiba (approved):

PR #531, opened 2024-08-05
Some modify and fix in ObjectCacheV2

Coldwings:
Some fixes.

### DIFF
--- a/common/objectcachev2.h
+++ b/common/objectcachev2.h
@@ -63,7 +63,7 @@ protected:
             std::swap(r, ref);
             return r;
         }
-        std::shared_ptr<V> clear(uint64_t ts = 0) {
+        std::shared_ptr<V> reset(uint64_t ts = 0) {
             return update({nullptr}, ts);
         }
         std::shared_ptr<V> reader() { return ref; }
@@ -116,7 +116,11 @@ protected:
         return *box;
     }
     uint64_t __expire() {
+#if __cplusplus < 201703L
         std::vector<std::shared_ptr<V>> to_release;
+#else
+        intrusive_list<Box> to_release;
+#endif
         uint64_t now = photon::now;
         uint64_t reclaim_before = photon::sat_sub(now, lifespan);
         {
@@ -132,13 +136,27 @@ protected:
                 if (x->rc == 0) {
                     // make vector holds those shared_ptr
                     // prevent object destroy in critical zone
-                    to_release.push_back(x->clear());
+#if __cplusplus < 201703L
+                    to_release.push_back(x->reset());
                     map.erase(*x);
+#else
+                    // Here is a hacked implementation
+                    auto node = map.extract(*x);
+                    to_release.push_back(&node.value());
+                    /// node_handle should be just key&val pointer
+                    // Force make it empty can prevent destroy object
+                    // when node_handle goes to destroy
+                    memset((void*)&node, 0, sizeof(node));
+#endif
                 }
                 x = lru_list.front();
             }
         }
+#if __cplusplus < 201703L
         to_release.clear();
+#else
+        to_release.delete_all();
+#endif
         return 0;
     }
 
@@ -162,7 +180,7 @@ public:
 
         ~Borrow() {
             if (_recycle) {
-                _box->clear();
+                _box->reset();
             }
             _box->release();
             if (_box->rc == 0) {
@@ -228,7 +246,8 @@ public:
 
     ~ObjectCacheV2() {
         _timer.stop();
-        SCOPED_LOCK(maplock);
+        // Should be no other access during dtor.
+        // modify lru_list do not need a lock.
         lru_list.node = nullptr;
     }
 };
--- a/common/objectcachev2.h
+++ b/common/objectcachev2.h
@@ -91,26 +91,20 @@ protected:
         }
     };
 
-    photon::thread* reclaimer = nullptr;
-    photon::common::RingChannel<LockfreeMPMCRingQueue<std::shared_ptr<V>, 4096>>
-        reclaim_queue;
-    photon::spinlock maplock;
     std::unordered_set<Box, ItemHash, ItemEqual> map;
-    intrusive_list<Box> cycle_list;
+    intrusive_list<Box> lru_list;
     uint64_t lifespan;
     photon::Timer _timer;
-    bool _exit = false;
+    photon::spinlock maplock;
 
     std::shared_ptr<V> __update(Box& item, std::shared_ptr<V> val) {
         auto ret_val = val;
-        auto old_val = item.writer().update(val, photon::now);
-        reclaim_queue.template send<PhotonPause>(old_val);
+        item.writer().update(val, photon::now);
         return ret_val;
     }
 
     std::shared_ptr<V> __update(Box& item, V* val) {
-        std::shared_ptr<V> ptr(val);
-        return __update(item, ptr);
+        return __update(item, std::shared_ptr<V>(val));
     }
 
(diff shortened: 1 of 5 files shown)


common/objectcachev2.h
```diff
@@ -91,26 +91,20 @@ class ObjectCacheV2 {
         }
     };
 
-    photon::thread* reclaimer = nullptr;
-    photon::common::RingChannel<LockfreeMPMCRingQueue<std::shared_ptr<V>, 4096>>
-        reclaim_queue;
-    photon::spinlock maplock;
     std::unordered_set<Box, ItemHash, ItemEqual> map;
```
lihuiba, line 94: use box/Box instead of item/Item

common/objectcachev2.h
```diff
@@ -120,37 +114,32 @@ class ObjectCacheV2 {
...
-        cycle_list.pop(item);
-        cycle_list.push_back(item);
+        lru_list.pop(item);
+        lru_list.push_back(item);
         return *item;
     }
 
     uint64_t __expire() {
-        intrusive_list<Box> delete_list;
         uint64_t now = photon::now;
```
lihuiba, line 130: It's simpler to use now - lifespan

common/objectcachev2.h
```diff
@@ -120,37 +114,32 @@ class ObjectCacheV2 {
...
+        SCOPED_LOCK(maplock);
+        if (lru_list.empty()) return 0;
+        if (photon::sat_add(lru_list.front()->timestamp, lifespan) > now) {
+            return photon::sat_add(lru_list.front()->timestamp, lifespan) -
+                    now;
+        }
+        auto x = lru_list.front();
+        while (x && (photon::sat_add(x->timestamp, lifespan) < now)) {
+            lru_list.pop(x);
+            __update(*x, nullptr);
```
lihuiba, line 140: it's more performant to realize x->reset(), which simply invokes shared_ptr::reset().

common/objectcachev2.h
```diff
@@ -164,20 +153,16 @@ class ObjectCacheV2 {
...
-            : _oc(oc),
-              _item(item),
-              _reader(std::move(reader)),
-              _recycle(false) {}
+        Borrow(ObjectCacheV2* oc, Box* item, const std::shared_ptr<V>& reader)
+            : _oc(oc), _item(item), _reader(reader), _recycle(false) {}
 
         Borrow(Borrow&& rhs) : _reader(nullptr) { *this = std::move(rhs); }
 
         Borrow& operator=(Borrow&& rhs) {
```
lihuiba, line 161: Instead of swap/exchange, it's more semantically right to:
```cpp
assert(_oc == rhs._oc);
std::atomic_exchange(&_reader, std::move(rhs._reader));
_item = rhs._item;
_recycle = rhs._recycle;
rhs._reader.reset();
rhs._item = nullptr;
rhs._recycle = false;
```

Coldwings: according to cppreference https://en.cppreference.com/w/cpp/memory/shared_ptr

>  If multiple threads of execution access the same shared_ptr object without synchronization and any of those accesses uses a non-const member function of shared_ptr then a data race will occur

Since atomic<shared_ptr> is not supported by c++11/14/17, an new implementation using photon::mutex helps keep shared_ptr update in critical zone.

common/objectcachev2.h
Coldwings: What if `lifespan` has been set a extra-big number (like UINT64_MAX, present object in objcache never reclaim

sat_add is necessary.

common/objectcachev2.h
Coldwings: In new implementation still use exchange/swap like stuff instead of reset().

swapped out object still have at lease one reference if result is assigned to another variable, so it could able to control when to perform object destruction.

common/objectcachev2.h
lihuiba: sat_sub(now, lifespan)

common/objectcachev2.h
```diff
@@ -226,33 +220,21 @@ class ObjectCacheV2 {
...
 
     ~ObjectCacheV2() {
         _timer.stop();
-        if (reclaimer) {
-            _exit = true;
-            reclaim_queue.template send<PhotonPause>(nullptr);
-            photon::thread_join((photon::join_handle*)reclaimer);
-            reclaimer = nullptr;
-        }
         SCOPED_LOCK(maplock);
```
lihuiba, line 237: use maplock only to protect a single assignment?

common/objectcachev2.h
```diff
@@ -198,24 +189,27 @@ class ObjectCacheV2 {
...
-                r = item.reader();
+            if (box.createlock.try_lock() == 0) {
+                DEFER(box.createlock.unlock());
+                r = box.reader();
                 if (!r) {
-                    if (photon::sat_add(item.lastcreate, cooldown) <=
+                    if (photon::sat_add(box.lastcreate, cooldown) <=
                         photon::now) {
-                        r = __update(item, ctor());
+                        auto r = std::shared_ptr<V>(ctor());
```
lihuiba, line 202: ctor() may return a shared_ptr?

common/objectcachev2.h
```diff
@@ -36,154 +37,144 @@ class ObjectCacheV2 {
...
-            item = (Box*)&*it;
-        assert(item);
-        item->timestamp = photon::now;
-        cycle_list.pop(item);
-        cycle_list.push_back(item);
-        return *item;
+    Box& __find_or_create_box(KeyType&& key) {
+        Box keybox(key);
+        Box* box = nullptr;
+        {
```
lihuiba, line 110: why this scope?

common/objectcachev2.h
```diff
@@ -36,154 +37,144 @@ class ObjectCacheV2 {
...
+            while (rc.load()) photon::thread_yield();
+        }
+
+        std::shared_ptr<V> update(std::shared_ptr<V> r, uint64_t ts = 0) {
+            SCOPED_LOCK(reflock);
+            lastcreate = ts;
+            std::swap(r, ref);
+            return r;
+        }
+        std::shared_ptr<V> clear(uint64_t ts = 0) {
```
lihuiba, line 70: "reset" seems better

common/objectcachev2.h
```diff
@@ -36,154 +37,138 @@ class ObjectCacheV2 {
...
-                x = cycle_list.front();
+            auto x = lru_list.front();
+            // here requires lock dance for lru_list and kv
+            while (x && x->timestamp < reclaim_before) {
+                lru_list.pop(x);
+                if (x->rc == 0) {
+                    // make vector holds those shared_ptr
+                    // prevent object destroy in critical zone
+                    to_release.push_back(x->clear());
+                    map.erase(*x);
```
lihuiba, line 136: use map.extract().value() obtain the Box element, so as to use the intrusive_list for to_release.

common/objectcachev2.h
Coldwings: code here want to access intrusive_list only for preventing hit assert of list.

Technically in dtor, other threads should not have any other accesses to the whole object cache, the lock here could be ignored.

I will remove the lock acquiring here

common/objectcachev2.h
Coldwings: Seems add macro to check c++ version to show different implementation for those above C++17 is a good idea.

Coldwings: Seem that C++17 node_handle owns the element Box, can not get out the pointer to Box without hacking.

lihuiba (approved):

PR #536, opened 2024-08-12
ObjCache V2 remove reflock

Coldwings:
* using atomic load/store to avoid using lock in box update/read.
* no more extract trick. it seems have problems.

### DIFF
--- a/common/objectcachev2.h
+++ b/common/objectcachev2.h
@@ -37,12 +37,6 @@ protected:
     struct Box : public intrusive_list_node<Box> {
         const K key;
         std::shared_ptr<V> ref;
-        // protect ref with mutex
-        // If multiple threads of execution access the same shared_ptr object
-        // without synchronization and any of those accesses uses a non-const
-        // member function of shared_ptr then a data race will occur
-        // DTOR of V may cause photon thread yield.
-        photon::mutex reflock;
         // prevent create multiple time when borrow
         photon::spinlock createlock;
         // create timestamp, for cool-down of borrow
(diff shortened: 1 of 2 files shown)


common/objectcachev2.h
```diff
@@ -58,15 +52,14 @@ class ObjectCacheV2 {
             : key(std::forward<KeyType>(key)), ref(nullptr) {}
 
         std::shared_ptr<V> update(std::shared_ptr<V> r, uint64_t ts = 0) {
-            SCOPED_LOCK(reflock);
             lastcreate = ts;
-            std::swap(r, ref);
+            r = std::atomic_exchange(&ref, r);
```
lihuiba, line 56: why not simply atomic_store?

Coldwings: atomic_store only stores but cannot load origin shared_ptr value for returning.

common/objectcachev2.h
```diff
@@ -136,27 +125,13 @@ class ObjectCacheV2 {
                 if (x->rc == 0) {
                     // make vector holds those shared_ptr
                     // prevent object destroy in critical zone
-#if __cplusplus < 201703L
                     to_release.push_back(x->reset());
```
Coldwings, line 128: Like here, have to reset & get old value as a atomic instruction, so that it can be collected for destruction.

common/objectcachev2.h
Coldwings: Old value have to be load and return.

See line 128.

Coldwings: The point is old object which ref refereed to should not be destruct during calling update or reset.

the shared_ptr have to be returned, so caller can decide if destruct right now or hold by other struct to destruct later.

lihuiba (approved):

PR #537, opened 2024-08-12
makes mprotect page and thread struct in the stack.

Coldwings:
Makes all photon::thread struct and mprotect page inside of the thread.
So thread stack now do not need to set mprotect repeatedly in pooled stack allocator,
and stack-allocation will be exactly size as it requested, more friendly to most of memory allocators.

### DIFF
--- a/thread/thread.cpp
+++ b/thread/thread.cpp
@@ -135,6 +135,7 @@ namespace photon
 #if defined(__linux__)
         madvise(ptr, stack_size, MADV_NOHUGEPAGE);
 #endif
+        mprotect(ptr, PAGE_SIZE, PROT_NONE);
         return ptr;
     }
 
@@ -306,8 +307,6 @@ namespace photon
             assert(state == states::DONE);
             // `buf` and `stack_size` will always store on register
             // when calling deallocating.
-            char* protect_head = (char*)align_up((uint64_t)buf, PAGE_SIZE);
-            mprotect(protect_head, PAGE_SIZE, PROT_READ | PROT_WRITE);
             photon_thread_dealloc(buf, stack_size);
         }
     };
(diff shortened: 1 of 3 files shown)


thread/thread.cpp
```diff
@@ -937,16 +937,16 @@ R"(
         if (unlikely(!rq.current))
             LOG_ERROR_RETURN(ENOSYS, nullptr, "Photon not initialized in this vCPU (OS thread)");
         size_t randomizer = (rand() % 32) * (1024 + 8);
-        stack_size = align_up(randomizer + stack_size + sizeof(thread), PAGE_SIZE);
-        stack_size += PAGE_SIZE * 2;  // extra 2 pages for alignment and set guard page
+        stack_size = align_up(stack_size, PAGE_SIZE);
+        if (stack_size <
```
lihuiba, line 941: if (stack_size < 16KB)

lihuiba: and check it before align_up()

thread/thread.cpp
```diff
@@ -135,13 +135,15 @@ namespace photon
 #if defined(__linux__)
         madvise(ptr, stack_size, MADV_NOHUGEPAGE);
 #endif
+        mprotect(ptr, PAGE_SIZE, PROT_NONE);
```
lihuiba, line 138: should we move the default allocator into stack-allocator.cpp ?

thread/thread.cpp
Coldwings: done

thread/thread.cpp
Coldwings: done

thread/thread.cpp
```diff
@@ -86,7 +86,7 @@ inline int posix_memalign(void** memptr, size_t alignment, size_t size) {
                            #name": "
 #endif
 
-static constexpr size_t PAGE_SIZE = 1 << 12;
+static const size_t PAGE_SIZE = getpagesize();
```
Coldwings, line 89: Turns out our ARM64 testing runner built on a kernel with 64KB page size, that makes all tests and codes that assumed pagesize is 4KB failed.

lihuiba (approved):

lihuiba (approved):

PR #541, opened 2024-08-15
update workpool-perf example

beef9999:
Test results:
```
Fire 64000 async works and solved by 4 vCPU, QPS is 17548, and average task deliver latency is 16346179 ns
Fire 64000 sync works and solved by 4 vCPU, QPS is 18427, and average task deliver latency is 1723612 ns
Fire 64000 sync works in std context and solved by 4 vCPU, QPS is 153561, and average task deliver latency is 752 us
```

Looks like we should always use WorkPool's sync call API in Photon Context, for better task deliver latency.


lihuiba:
Why is the ```average task deliver latency``` so huge?

thread/test/perf_workpool.cpp
```diff
@@ -41,20 +40,36 @@ void* task_async(void*) {
...
+            sum_time.fetch_add(std::chrono::duration_cast<std::chrono::nanoseconds>(end - start).count(),
+                               std::memory_order_relaxed);
+
+        });
+    }
+}
+
+static inline size_t get_qps(std::chrono::time_point<std::chrono::steady_clock> start,
+                             std::chrono::time_point<std::chrono::steady_clock> end) {
+    return FLAGS_fires * 1000 * 1000 / std::chrono::duration_cast<std::chrono::microseconds>(end - start).count();
```
lihuiba, line 72: why ```* 1000 * 1000```?

thread/test/perf_workpool.cpp
```diff
@@ -66,26 +81,37 @@ int main(int argc, char** arg) {
...
 
+    // 2. Same as 1, but use sync API
     sum_time = 0;
-    jhs.clear();
-
-    start = photon::now;
-    photon::threads_create_join(FLAGS_fires, task_sync, nullptr);
-    end = photon::now;
+    start = std::chrono::steady_clock::now();
+    task_sync();
```
lihuiba, line 98: it seems that you are using the same workpool as No. 1

lihuiba:
> because of the lack of thread pool

For both 1 and 2, you are using the same workpool that doesn't include a thread pool.

thread/test/perf_workpool.cpp
```diff
@@ -66,26 +81,37 @@ int main(int argc, char** arg) {
...
+
+    // 3. non-thread mode WorkPool. Tasks are from std context
+    delete pool;
+    pool = new photon::WorkPool(FLAGS_vcpu_num, photon::INIT_EVENT_DEFAULT, photon::INIT_IO_NONE, -1);
+    sum_time = 0;
+    start = std::chrono::steady_clock::now();
+    std::thread th(task_sync_in_std_context);
+    th.join();
+    end = std::chrono::steady_clock::now();
+    LOG_INFO("Fire ` sync works in std context and solved by ` vCPU, QPS is `, and average task deliver latency is ` us",
```
lihuiba, line 113: sum_time is in ns, not us

thread/test/perf_workpool.cpp
lihuiba, line 72: QPS seems to be limited by the length of queue. This should be avoided in benchmarks.

lihuiba (approved):

PR #545, opened 2024-08-19
update perf_workpool.cpp

beef9999:


thread/test/perf_workpool.cpp
```diff
@@ -30,9 +30,22 @@ DEFINE_uint64(fires, 80000, "How many tasks to fire");
...
-void task_async() {
+__attribute__ ((noinline))
+static void do_fill(char* buf, size_t size) {
+    memset(buf, 0, size);
+}
+// A function that costs approximately 1 us
+static void func_1us() {
+    for (int i = 0; i < 5; ++i) {
+        char buf[32 * 1024];
+        do_fill(buf, sizeof(buf));
```
lihuiba, line 41: It's better to write a loop and repeatedly query time.

thread/test/perf_workpool.cpp
```diff
@@ -30,9 +30,22 @@ DEFINE_uint64(fires, 80000, "How many tasks to fire");
...
+    }
+}
+
+static void task_async() {
     photon::semaphore sem(0);
     for (uint64_t i = 0; i < FLAGS_fires; ++i) {
+        func_1us();
```
lihuiba, line 48: But what's the purpose to permanently add this wait?

beef9999: To simulate real work loads, and make the test results more reasonable.
Otherwise the performance of case 1 is too low.

thread/test/perf_workpool.cpp
beef9999: ok

lihuiba (approved):

PR #546, opened 2024-08-20
unordered_inline_set

lihuiba:
A hash table with open addressing, i.e. collision is resolved by probing, or searching through alternative locations in the array. 

This design avoids allocation and deallocation during insertion and deletion, except rehashing.

it also features the ability to extract (move out) a stored element itself, without involving a "handle".

This is only an initial version that need further test and polish.

### DIFF
--- a/common/unordered_inline_set.h
+++ b/common/unordered_inline_set.h
@@ -1,0 +1,741 @@
+#pragma once
+#include <assert.h>
+#include <inttypes.h>
+#include <unordered_set>
+#include <memory>
+#include <type_traits>
+#include <limits>
+#include <cmath>
+#include <string.h>
+
+// an array with possible missing values
+template<typename T>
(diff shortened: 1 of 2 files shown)


common/unordered_inline_set.h
```diff
@@ -0,0 +1,741 @@
...
+    }
+
+    void swap( unordered_inline_set& other ) noexcept;
+
+    template< class K >
+    value_type&& extract( K&& x ) {
+        return extract(find(std::forward<K>(x)));
+    }
+    value_type&& extract( iterator position ) {
+        assert(position._a == &_data);
```
Coldwings, line 624: `position._a` `position._i` are protected field of `iterator` type
and `get_bit` is protected

Coldwings (approved):
LGTM

PR #552, opened 2024-09-03
fix RPC stop by timeout during calling

Coldwings:
PR #164 add retry by thread yield in mutex,breaks assumption in OutOfOrder engine, that in single vCPU, when a photon thread entered critical zone, other threads already slept or still not add into tag map.

Also, since RPC call now with Timeout to limit waiting time, it should set to OutOfOrder waiting procedure so RPC call can make sure break-out when timed out.

This PR fixed those problems, and add a test for such case.

### DIFF
--- a/common/stream.h
+++ b/common/stream.h
@@ -47,8 +47,13 @@ public:
     }
 
     // get/set default timeout, in us, (default +∞)
-    virtual uint64_t timeout() const = 0;
-    virtual void timeout(uint64_t tm) = 0;
+    virtual uint64_t timeout() const {
+        errno = ENOSYS;
+        return -1;
+    }
+    virtual void timeout(uint64_t tm) {
+        errno = ENOSYS;
+    }
 
     struct ReadAll {
         struct FreeDeleter {
--- a/common/stream.h
+++ b/common/stream.h
@@ -15,10 +15,12 @@ limitations under the License.
 */
 
 #pragma once
+#include <errno.h>
+#include <photon/common/object.h>
 #include <stdlib.h>
 #include <sys/types.h>
+
 #include <memory>
-#include <photon/common/object.h>
 
 struct iovec;
 
--- a/common/stream.h
+++ b/common/stream.h
@@ -46,6 +46,10 @@ public:
         return writev(iov, iovcnt);
     }
 
+    // get/set default timeout, in us, (default +∞)
+    virtual uint64_t timeout() const = 0;
+    virtual void timeout(uint64_t tm) = 0;
+
     struct ReadAll {
         struct FreeDeleter {
             void operator()(void* ptr) {
--- a/rpc/out-of-order-execution.h
+++ b/rpc/out-of-order-execution.h
@@ -27,7 +27,7 @@ collection of result. The first 2 parts are realized via callbacks.
 #pragma once
 #include <photon/common/callback.h>
 #include <photon/common/timeout.h>
-#include <atomic>
+#include <photon/thread/thread.h>
 
 namespace photon{
 
@@ -42,6 +42,12 @@ namespace rpc {
 
     void delete_ooo_execution_engine(OutOfOrder_Execution_Engine* engine);
 
+    enum class OooPhase : int {
+        BEFORE_ISSUE = 0,
+        ISSUED = 1,
+        WAITING = 2,
+        COLLECTED = 3
+    };
     struct OutOfOrderContext
     {
         OutOfOrder_Execution_Engine* engine;
@@ -49,7 +55,7 @@ namespace rpc {
         // an unique tag of the opeartion, which can be filled
         // by user (together with `flag_tag_valid` = true),
         // by the `engine`, or by `do_completion`.
-        uint64_t tag;
+        uint64_t tag = 0;
 
         // The `CallbackType` have an prototype of
         // either `int (*)(void*, OutOfOrderContext*)`,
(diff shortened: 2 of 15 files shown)


rpc/rpc.h
```diff
@@ -243,10 +247,23 @@ namespace rpc
...
+    2. When socket connecting, it will fail by be timed out after `connect_timeout` microsecs.
+    3. When socket needs to read/write (like sending request and waiting for any response), but got
+       nothing in `transfer_timeout` microsecs, it will be timed out. It's just a timeout for a single
+       read/write operation, not a RPC timeout, which is controlled by the caller when invoking `Stub::call`.
+    4. `Stub::call` measures the time from invoking `call` before sending request to received
+       response head. Receiving response body is not considered.
+    **/
+    extern "C" StubPool* new_stub_pool(uint64_t expiration,
+                                       uint64_t connect_timeout,
+                                       uint64_t transfer_timeout);
```
beef9999, line 262: rpc_timeout?

lihuiba: it's actually transfer timeout

rpc/rpc.cpp
```diff
@@ -421,12 +419,12 @@ namespace rpc {
 
     class StubPoolImpl : public StubPool {
     public:
-        explicit StubPoolImpl(uint64_t expiration, uint64_t connect_timeout, uint64_t rpc_timeout) {
+        explicit StubPoolImpl(uint64_t expiration, int64_t connect_timeout, uint64_t transfer_timeout) {
```
lihuiba, line 422: uint64_t

rpc/rpc.h
```diff
@@ -59,10 +59,13 @@ namespace rpc
         /**
          * @param req Request of Message
          * @param resp Response of Message
+         * @param timeout timeout in milliseconds, -1UL for no timeout
```
beef9999, line 62: micro?

rpc/rpc.h
Coldwings: > transfer_timeout -> rpc_timeout?

When socket needs to read/write (like sending request and waiting for any response), but got nothing in `transfer_timeout` microsecs, it will be timed out. It's just a timeout for a single read/write operation, not a RPC timeout, which is controlled by the caller when invoking `Stub::call`.

That means that if on same stream, a heartbeat RPC keep beeping, the stream keeps reading and writing without trigger this timeout. Other concurrent RPC call on this stream will not report as timedout by `transfer_timeout`. So I think `rpc_timeout` is not proper name for it.

rpc/rpc.h
Coldwings: Changed to use `Timeout`

rpc/rpc.cpp
Coldwings: fixed

rpc/test/test.cpp
```diff
@@ -485,6 +485,123 @@ TEST_F(RpcTest, passive_shutdown) {
...
+        photon::thread_usleep(5UL*1000*1000);
+        return 0;
+    }
+
+    int serve(photon::net::ISocketStream* stream) {
+        return m_skeleton->serve(stream);
+    }
+    int run() {
+        if (m_socket->bind_v4localhost() != 0)
+        // if (m_socket->bind(9527, net::IPAddr::V6Any()) != 0)
```
beef9999, line 534: remove unused code?

rpc/test/test.cpp
```diff
@@ -485,6 +485,123 @@ TEST_F(RpcTest, passive_shutdown) {
...
+    // photon::net::EndPoint ep(net::IPAddr::V4Loopback(), 9527);
+    auto& ep = rpc_server.m_endpoint;
+    auto pool = photon::rpc::new_stub_pool(-1, -1, 1'000'000);
+    DEFER(delete pool);
+    auto th1 = photon::thread_enable_join(photon::thread_create11([&]{
+        // Should always succeed in 3 seconds
+        auto stub = pool->get_stub(ep, false);
+        if (!stub) abort();
+        DEFER(pool->put_stub(ep, false));
+        Timeout timeout(7'000'000);
```
beef9999, line 574: 3 seconds or 7 seconds?

rpc/test/test.cpp
```diff
@@ -485,6 +485,123 @@ TEST_F(RpcTest, passive_shutdown) {
...
+        // Should get connection refused after 2 seconds. Because socket closed listen fd at 1 second.
+        auto stub = pool->get_stub(ep, false);
+        if (!stub) {
+            abort();
+        }
+        DEFER(pool->put_stub(ep, false));
+        RpcServerTimeout::OperationT::Request req;
+        RpcServerTimeout::OperationT::Response resp;
+        auto before = photon::now;
+        auto ret = stub->call<RpcServerTimeout::OperationT>(req, resp, 1'000'000);
```
beef9999, line 596: If transfer timeout is 1s and rpc_call's timeout is 0.5s, you will find the end-to-end RPC cost time is still around 1s, which is very strange.

Coldwings:
A new OutOfOrderEngine implementation, do not depended on 'non-switch' assumption between issue and wait.
Able to share between multiple vCPUs.

rpc/test/test.cpp
Coldwings: > 4. `Stub::call` measures the time from invoking `call` before sending request to received
> response head. Receiving response body is not considered.

once got response head for the request, the new implementation will keep receiving body till success or another transfer timeout.

Since RPC framework using IStream as data stream, there is no API to set timeout during RPC receiving. Currently socket timeout can only set by ISockStream.

A better idea is add timeout for IStream, or make it possible set timeout for a single read/write operation for IStream. That will leads basic interface api change (IStream is the basic interface for IFile, ISocketStream, etc.), will need more discussion.

rpc/test/test.cpp
Coldwings: Comment should change.

rpc/test/test.cpp
Coldwings: Good idea

All unused code in this cpp has been removed.

Coldwings:
We have discussed and decide to add timeout methods for the `IStream` interface. 

After that, RPC could control all part timeout including socket sending/receiving, `RPC::call` becoming able to set roundtrip timeout.

But, that leads to `IFile` `http::Message` `http::Body` also have a new method may need to be implemented.

common/stream.h
```diff
@@ -46,6 +46,10 @@ class IStream : public Object
         return writev(iov, iovcnt);
     }
 
+    // get/set default timeout, in us, (default +∞)
+    virtual uint64_t timeout() const = 0;
```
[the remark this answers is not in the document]
Coldwings: done

rpc/out-of-order-execution.h
```diff
@@ -25,8 +25,9 @@ collection of result. The first 2 parts are realized via callbacks.
 */
 
 #pragma once
-#include <cinttypes>
 #include <photon/common/callback.h>
+#include <photon/common/timeout.h>
+#include <atomic>
```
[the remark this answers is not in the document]
Coldwings: done

rpc/test/test.cpp
Coldwings: Since new timeout methods are add to IStream, now transfer-timeout become no meaning.
RPC will controls timeout for a whole roundtrip.

lihuiba (approved):

lihuiba:
is there any performance impact?

Coldwings:
> is there any performance impact?

Not found obvious performance impact yet.
With long-keep tcp, some RPC application even shows some improvement.

PR #553, opened 2024-09-09
FIX: wait_for_fd in epoll reap events when timeout is 0

Coldwings:
socket pool use `wait_for_fd` to check if released fd still have bytes left to read, but set timeout to 0.

since thread_usleep will always timedout when sleep time left 0, set 10us timeout, so that thread can actually put into sleep and wait for notify.

### DIFF
--- a/io/epoll.cpp
+++ b/io/epoll.cpp
@@ -322,7 +322,7 @@ ok:     entry.interests |= eint;
             }
             return ret;
         }
-        ret = thread_usleep(timeout.timeout() ? timeout : Timeout(10));
+        ret = thread_usleep(timeout);
         ERRNO err;
         if (ret == -1 && err.no == EOK) {
             return 0;  // Event arrived
--- a/io/epoll.cpp
+++ b/io/epoll.cpp
@@ -301,7 +301,9 @@ ok:     entry.interests |= eint;
             return rm_interest({fd, EVENT_RWE| ONE_SHOT, 0}); // remove fd from epoll
         int ret = add_interest({fd, interest | ONE_SHOT, CURRENT});
         if (ret < 0) LOG_ERROR_RETURN(0, -1, "failed to add event interest");
-        ret = thread_usleep(timeout);
+        // if timeout is just simple 0, wait for a tiny little moment
+        // so that events can be collect.
+        ret = thread_usleep(timeout.timeout() ? timeout : Timeout(10));
         ERRNO err;
         if (ret == -1 && err.no == EOK) {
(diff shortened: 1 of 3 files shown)


HanLee13:
kqueue needs exactly the same fix, which will also fix macos CI problems in client_function_test unittest.

HanLee13 (commented):
kqueue needs exactly the same fix, which will also fix macos CI problems in client_function_test unittest.

HanLee13 (approved):

beef9999:
merge to 0.8?

HanLee13:
> merge to 0.8?

Better merging from 0.6 ?

io/epoll.cpp
```diff
@@ -301,7 +301,9 @@ ok:     entry.interests |= eint;
             return rm_interest({fd, EVENT_RWE| ONE_SHOT, 0}); // remove fd from epoll
         int ret = add_interest({fd, interest | ONE_SHOT, CURRENT});
         if (ret < 0) LOG_ERROR_RETURN(0, -1, "failed to add event interest");
-        ret = thread_usleep(timeout);
+        // if timeout is just simple 0, wait for a tiny little moment
+        // so that events can be collect.
+        ret = thread_usleep(timeout.timeout() ? timeout : Timeout(10));
```
lihuiba, line 306: In case of timeout == 0, we'd better reap the events directly without blocking, so as to conform to the semantics.

io/kqueue.cpp
```diff
@@ -99,7 +99,7 @@ class KQueue : public MasterEventEngine, public CascadingEventEngine, public Res
             return 0;
         short ev = (interests == EVENT_READ) ? EVFILT_READ : EVFILT_WRITE;
         enqueue(fd, ev, EV_ADD | EV_ONESHOT, 0, CURRENT);
-        int ret = thread_usleep(timeout);
+        int ret = thread_usleep(timeout.timeout() ? timeout : Timeout(10));
```
lihuiba, line 102: In case of timeout == 0, we'd better reap the events directly without blocking, so as to conform to the semantics.

io/epoll.cpp
Coldwings: done

io/kqueue.cpp
Coldwings: done

Coldwings:
Once this PR checked and approved, I will request other PR for releases from 0.6 to 0.8

lihuiba (approved):

PR #554, opened 2024-09-10
Fix sockpool wait_for_fd for old releases

Coldwings:
Same as #553 but for dev/0.6

### DIFF
--- a/io/epoll.cpp
+++ b/io/epoll.cpp
@@ -285,7 +285,7 @@ public:
                 },
                 [&]() __INLINE__ { return true; });
             if (ret < 0) {
-                rm_interest({fd, interests, 0});
+                rm_interest(event);
                 errno = ETIMEDOUT;
             }
         }
@@ -294,7 +294,7 @@ public:
         if (ret == -1 && err.no == EOK) {
             return 0;  // Event arrived
         }
-        rm_interest({fd, interests, 0}); // no ONE_SHOT, to reconfig epoll
+        rm_interest(event); // no ONE_SHOT, to reconfig epoll
         errno = (ret == 0) ? ETIMEDOUT :    // Timeout
                              err.no; // Interrupted by other thread
         return -1;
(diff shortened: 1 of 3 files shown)


io/epoll.cpp
```diff
@@ -266,22 +266,38 @@ class EventEngineEPoll : public MasterEventEngine, public CascadingEventEngine {
...
-        } else if (ret == 0) {
-            rm_interest(event);  // Timeout
-            errno = ETIMEDOUT;
-            return -1;
-        } else {
-            rm_interest(event);  // Interrupted by other thread
-            errno = err.no;
-            return -1;
         }
+        rm_interest({fd, interests, 0}); // no ONE_SHOT, to reconfig epoll
```
lihuiba, line 297: it's might be safer to ```rm_interest(event)``` as before.

io/kqueue.cpp
```diff
@@ -164,11 +187,12 @@ class KQueue : public MasterEventEngine, public CascadingEventEngine {
...
 
     ssize_t wait_for_events(void** data,
-            size_t count, uint64_t timeout = -1) override {
-        int ret = get_vcpu()->master_event_engine->wait_for_fd_readable(_kq, timeout);
+            size_t count, uint64_t timeout) override {
+        int ret = ::photon::wait_for_fd_readable(_kq, timeout);
         if (ret < 0) return errno == ETIMEDOUT ? 0 : -1;
         if (count > LEN(_events))
             count = LEN(_events);
+        static const struct timespec _tm = {0, 0};
```
lihuiba, line 195: Why static here, while not in enqueue()?

io/epoll.cpp
Coldwings: done

io/kqueue.cpp
Coldwings: Use on-stack temporary timespec is fine. Changed in latest commit.

Coldwings:
Discussed that old version may not need such fix. This fix will be dropped.

PR #559, opened 2024-09-13
thread_yield() possible to return error_number obtained from thread_i…

lihuiba:
making thread_yield() possible to return error_number obtained from thread_interrupt()


lihuiba:
not to patch release/0.6, unless absolutely necessary.

PR #560, opened 2024-09-14
thread_yield() possible to return error_number obtained from thread_i…

lihuiba:
making thread_yield() possible to return error_number obtained from thread_interrupt()

### DIFF
--- a/thread/thread.cpp
+++ b/thread/thread.cpp
@@ -1162,41 +1162,44 @@ R"(
         return (states) th->state;
     }
 
-    void thread_yield()
+    int thread_yield()
     {
-        assert(!AtomicRunQ().single());
-        auto sw = AtomicRunQ().goto_next();
+        RunQ rq;
         if_update_now();
+        rq.current->error_number = 0;
+        auto sw = AtomicRunQ(rq).goto_next();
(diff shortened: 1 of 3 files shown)


thread/thread.cpp
```diff
@@ -1319,9 +1325,14 @@ R"(
     {
         if (unlikely(!th))
             LOG_ERROR_RETURN(EINVAL, , "invalid parameter");
-        if (unlikely(th->state != states::SLEEPING)) return;
+        if (unlikely(th->state != states::SLEEPING)) {
+        out: // may have thread_yield()-ed
+            if (th->state != states::READY || th->error_number == 0)
```
Coldwings, line 1330: Should it be
`th->state == states::READY && th->error_number == 0`?
Means thread is in READY queue and error number not set

thread/thread.cpp
```diff
@@ -1319,9 +1325,14 @@ R"(
...
+                th->error_number = error_number;
```
Coldwings, line 1331: multi vcpu condition may breaks interrupt error_number.

due to visibility, and non-atomic check and set for `th->state` and `th->error_number`

Coldwings (approved):
LGTM

PR #561, opened 2024-09-16
add an emulation layer for state threads, passing its UT

lihuiba:
The official site of ST (no maintenance anymore):
https://state-threads.sourceforge.net/docs/

Maintained fork of ST:
https://github.com/ossrs/state-threads


Coldwings (approved):
LGTM

PR #568, opened 2024-09-25
Fix getpagesize() when first used

beef9999:


thread/stack-allocator.cpp
```diff
@@ -28,7 +28,10 @@ limitations under the License.
 
 namespace photon {
 
-const static size_t PAGE_SIZE = getpagesize();
+inline uint32_t get_pagesize() {
```
Coldwings, line 31: How about put this inline function `get_pagesize()` into thread.h file,
so that same code could not appear twice in both `stack-allocator.cpp` and `thread.cpp`?

thread/thread.cpp
```diff
@@ -86,11 +86,15 @@ inline int posix_memalign(void** memptr, size_t alignment, size_t size) {
                            #name": "
 #endif
 
-static const size_t PAGE_SIZE = getpagesize();
-
 namespace photon
 {
     inline uint64_t min(uint64_t a, uint64_t b) { return (a<b) ? a : b; }
+
+    inline uint32_t get_pagesize() {
```
Coldwings, line 93: How about put this inline function `get_pagesize()` into thread.h file,
so that same code could not appear twice in both `stack-allocator.cpp` and `thread.cpp`?

thread/stack-allocator.cpp
beef9999: Talked to @lihuiba , he suggested if page_size being used somewhere else again, we encapsulate it.

lihuiba: If it is used in a third place in the future, let's put it in a common header file, but not in thread.h, because it's conceptually unrelated to thread.

thread/stack-allocator.cpp
```diff
@@ -28,7 +28,10 @@ limitations under the License.
...
+    static uint32_t page_size = getpagesize();
+    return page_size;
```
lihuiba, line 33: It's better not to use auto init, as it incurs an implicit mutex, which is useless and inflates the code size.

Simply init it manually as:
```cpp
static uint32_t page_size = 0;
if (unlikely(page_size == 0)) 
    page_size = getpagesize();
return page_size;
```

lihuiba: Or rather, we can define an global variable that gets initialized in vcpu_init().

PR #571, opened 2024-09-25
case-insensitive map_string_key

lihuiba:


common/string-keyed.h
```diff
@@ -156,6 +157,40 @@ template<class T,
...
 
+class Hasher_CaseInsensitive {
+    const size_t BUF_CAP = 64;
+    size_t partial_hash(std::string_view sv, size_t i, size_t n) const {
+        char buf[BUF_CAP];
+        sv = sv.substr(i, n);
+        assert(sv.size() <= BUF_CAP);
+        for (size_t j = 0; j < sv.size(); ++j)
+            buf[j] = (char)tolower(sv[j]);
+        return std::hash<std::string_view>()({buf, n});
```
Coldwings, line 168: Might be better if using template parameter to set hasher (and even combine hasher) for parts instead of `std::hash<std::string_view>`.

hasher for `unordered_map_string_key` can be set by template parameter, so `unordered_map_string_key_case_insensitive` would be better able to set hasher.

Coldwings (approved):
LGTM

PR #573, opened 2024-09-25
Fix photon_std::thread::id default constructor

beef9999:
The standard lib provides such syntax:
```std::thread::id tid;```

For photon_std, it is `photon:thread*`, and probably not init to 0, so caused a CI failure in RocksDB.

This PR makes it follow the standard.

### DIFF
--- a/thread/std-compat.h
+++ b/thread/std-compat.h
@@ -55,8 +55,6 @@ inline uint64_t __duration_to_microseconds(const ::std::chrono::duration<Rep, Pe
 
 class thread {
 public:
-    using id = photon::thread*;
-
     thread() = default;
 
     ~thread() {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: thread/std-compat.h [hunk 1]
```diff
@@ -93,6 +91,17 @@ class thread {
         return m_th != nullptr;
     }
 
+    class id {
```
lihuiba, line 94: ```using id = std::tuple<photon::thread*>;```  may work?

beef9999: Does tuple support auto hash?

And I think a class is more clear.

lihuiba: tuple has not hash, but it's clear by itself, as long as the reader know what is a tuple.

lihuiba (approved):

PR #580, opened 2024-09-30
Fix test-throttle CI

beef9999:


common/test/test_throttle.cpp
```diff
@@ -429,8 +433,8 @@ TEST_P(ThrottlePriorityTest, run) {
 #endif
 
 int main(int argc, char** argv) {
-    if (!photon::is_using_default_engine()) return 0;
```
lihuiba, line 432: does throttle need to be tested against all event engine?

beef9999: yes

beef9999:
epoll-ng succeeded after retry. Ready to merge.

lihuiba (approved):

PR #582, opened 2024-10-01
retval refinement

lihuiba:


common/retval.h
```diff
@@ -35,14 +37,17 @@ T failure_value() { return 0; }
...
+    retval(const retval&) = default;
+    retval(int _errno, T val) : retval_base{_errno}, _val(val) { }
+    retval(int _errno) : retval_base{_errno}, _val(failure_value<T>()) {
         assert(failed());
     }
     retval(const retval_base& rvb) : retval_base(rvb) {
         assert(failed());
         _val = failure_value<T>();
     }
+    template<typename P, typename Q = typename std::enable_if<!std::is_integral<P>::value>::type>
```
lihuiba, line 49: prevent direct construction from T, if T is an integral type, so as to avoid conflicts with error number of int.

thread/test/st_utest.cpp
```diff
@@ -26,7 +26,9 @@ GTEST_API_ int main(int argc, char **argv) {
 #endif
 
     // Initialize state-threads, create idle coroutine.
-    assert(st_init() == 0);
```
lihuiba, line 29: assert() can be removed if NDEBUG is defined, and st_init() is missed.

Coldwings (approved):
LGTM

Coldwings (approved):
LGTM

PR #595, opened 2024-10-10
Fix tes-thread CI OOM killed

beef9999:


thread/test/test.cpp
```diff
@@ -685,7 +685,7 @@ TEST(Semaphore, heavy) {
     semaphore sem(0);
     const int thread_num = 100000;
     for (int i=1; i<=thread_num;i++) {
-        thread_create11(16 * PAGE_SIZE, semaphore_heavy, sem, i);
+        thread_create11(4 * PAGE_SIZE, semaphore_heavy, sem, i);
```
beef9999, line 688: Self-hosted ARM CI machine doesn't have plenty of memory. Test might be OOM killed.

thread/test/test.cpp
```diff
@@ -1764,11 +1764,12 @@ TEST(future, test2) {
 int main(int argc, char** arg)
 {
     if (!photon::is_using_default_engine()) return 0;
+    srand(time(nullptr));
     ::testing::InitGoogleTest(&argc, arg);
     gflags::ParseCommandLineFlags(&argc, &arg, true);
     default_audit_logger.log_output = log_output_stdout;
     photon::vcpu_init();
-    set_log_output_level(ALOG_DEBUG);
+    set_log_output_level(ALOG_INFO);
```
beef9999, line 1772: Less log, less memory. And running faster is helpful to find bugs

lihuiba (approved):

PR #601, opened 2024-10-16
string_socket_stream

lihuiba:


common/memory-stream/memory-stream.cpp
```diff
@@ -198,6 +198,81 @@ class FaultStream : public IStream
...
+        _out.append(p, p + n);
+        return n;
+    }
+    virtual ssize_t send(const struct iovec *iov, int iovcnt, int flags = 0) override {
+        while(!iov->iov_base || !iov->iov_len)
+            if (iovcnt) { ++iov, --iovcnt; }
+            else return -1;
+        return send(iov->iov_base, iov->iov_len);
+    }
+    virtual ssize_t sendfile(int in_fd, off_t offset, size_t count) override { return 0; }
```
Coldwings, line 231: return -1 with errno = ENOSYS?

common/memory-stream/memory-stream.cpp
```diff
@@ -198,6 +198,81 @@ class FaultStream : public IStream
...
+    virtual ssize_t write(const void *buf, size_t count) override {
+        auto p = (const char*)buf;
+        _out.append(p, p + count);
+        return count;
+    }
+    virtual ssize_t writev(const struct iovec *iov, int iovcnt) override {
+        ssize_t s = 0;
+        for (int i = 0; i < iovcnt; ++i) {
+            ssize_t ret = write(iov[i].iov_base, iov[i].iov_len);
+            if (ret < 0) return ret;
```
Coldwings, line 257: should return `s` if `s > 0`?

common/memory-stream/memory-stream.cpp
```diff
@@ -198,6 +198,81 @@ class FaultStream : public IStream
...
+        return n;
+    }
+    virtual ssize_t send(const struct iovec *iov, int iovcnt, int flags = 0) override {
+        while(!iov->iov_base || !iov->iov_len)
+            if (iovcnt) { ++iov, --iovcnt; }
+            else return -1;
+        return send(iov->iov_base, iov->iov_len);
+    }
+    virtual ssize_t sendfile(int in_fd, off_t offset, size_t count) override { return 0; }
+    virtual int close() override { return 0; }
```
Coldwings, line 232: close (and shutdown) should actually prevent more recv/send

Coldwings (approved):

common/memory-stream/memory-stream.cpp
lihuiba: in the case of an error, just return the error

PR #618, opened 2024-11-06
making thread_local fiber_local

lihuiba:


glibc-2.40.patch
```diff
@@ -0,0 +1,11 @@
+--- a/nptl/allocatestack.c
++++ b/nptl/allocatestack.c
+@@ -210,7 +210,7 @@ advise_stack_range (void *mem, size_t size, uintptr_t pd, size_t guardsize)
+    new stack or reusing a cached stack of sufficient size.
+    ATTR must be non-NULL and point to a valid pthread_attr.
+    PDP must be non-NULL.  */
+-static int
++int
```
lihuiba, line 8: making the internal function ```allocate_stack()``` in lib pthread public and accessible from user programs. (std::thread depends on pthread too)

thread/thread.cpp
```diff
@@ -534,14 +540,7 @@ namespace photon
             vcpu = current->get_vcpu();
             (plock = &vcpu->runq_lock) -> foreground_lock();
         }
-        mutable bool update_current = false;
-        void set_current(thread* th) const {
```
lihuiba, line 538: There's no longer the need to set/switch the thread-local variable CURRENT as part of context switching, given that it has become a fiber-local varible.

thread/thread.cpp
```diff
@@ -648,42 +645,38 @@ namespace photon
...
     }
 
-    static void _photon_thread_die(thread* th) asm("_photon_thread_die");
-
 #if defined(__x86_64__)
 #if !defined(_WIN64)
     asm(
 DEF_ASM_FUNC(_photon_switch_context) // (void** rdi_to, void** rsi_from)
 R"(
+        rdfsbase %rax
```
lihuiba, line 653: Save and restore fs register as part of context switching. It points to thread control block (TCB) and thread-local storage (TLS)

thread/thread.cpp
```diff
@@ -896,14 +889,17 @@ R"(
         _photon_switch_context_defer_die(
             arg, func, sw.to->stack.pointer_ref());
     }
-    static __attribute__((used, noreturn))
-    void _photon_thread_die(thread* th) {
+    static __attribute__((noreturn))
+    void _photon_thread_stub() {
+        register thread* th asm("rbp");
```
lihuiba, line 894: On entry rbp points to struct thread.

thread/thread.cpp
```diff
@@ -922,7 +918,24 @@ R"(
...
             stack_size = least_stack_size;
         }
-        char* ptr = (char*)photon_thread_alloc(stack_size);
+        // char* ptr = (char*)photon_thread_alloc(stack_size);
+        struct pthread* pd;
+        char* ptr;  // ptr to stack
+        size_t pstacksize;
+        pthread_attr_t attr;
+        pthread_attr_init(&attr);
+        allocate_stack((struct pthread_attr*)&attr, &pd, (void**)&ptr, &pstacksize);
```
lihuiba, line 927: We must make use of ```allocate_thread()``` in lib pthread to allocate the thread, TCB (that compatible to threading facilities) and TLS for all modules loaded at startup time.

thread/thread.cpp
```diff
@@ -922,7 +918,24 @@ R"(
...
+
+// this should be moved into allocate_stack(),
+// as we don't have the defination of *pd
+#if TLS_TCB_AT_TP
+        /* Reference to the TCB itself.  */
+        pd->header.self = pd;
+
```
lihuiba, line 934: These lines are copied from ```pthread_create()```, and should be moved to ```allocate_thread()```.

glibc-2.40.patch
```diff
@@ -0,0 +1,53 @@
...
+            }
+
++  /* Initialize the TCB.  All initializations with zero should be
++     performed in 'get_cached_stack'.  This way we avoid doing this if
++     the stack freshly allocated with 'mmap'.  */
++
++#if TLS_TCB_AT_TP
++    /* Reference to the TCB itself.  */
++    pd->header.self = pd;
++
```
lihuiba, line 23: These lines are moved from pthread_create(), because fiber_create() doesn't have the definition of *pd, and can not do it outside.

thread/thread.cpp
```diff
@@ -648,42 +645,38 @@ namespace photon
...
 DEF_ASM_FUNC(_photon_switch_context_defer_die) // (void* rdi_arg, void (*rsi_defer)(void*), void** rdx_to_th)
 R"(
         mov     (%rdx), %rsp
+        pop     %rax
         pop     %rbp
+        wrfsbase %rax
         jmp     *%rsi
 )"
-
-DEF_ASM_FUNC(_photon_thread_stub)
```
lihuiba, line 678: _photon_thread_stub() becomes more complex and would better be implemented in C++.

PR #625, opened 2024-11-22
make cross-vcpu interrupt latency lower

lihuiba:
make cross-vcpu interrupt latency lower, by more frequent checking of standby queue.


Coldwings (approved):
LGTM

PR #633, opened 2024-12-02
Able to use RDMACore RSocket in photon

Coldwings:
A simple ISocketStream/ISocketServer/ISocketClient implementation as a adaptor for `rsocket` in `rdma-core`, able to use `rsocket` as other socket streams in photon.


net/rsocket/rsocket.h
```diff
@@ -0,0 +1,6 @@
+#pragma once
+
+#include <photon/net/socket.h>
+
```
lihuiba, line 4: namespace photon { 
namespace net {

net/rsocket/rsocket.cpp
```diff
@@ -0,0 +1,530 @@
...
+        errno = ENOSYS;
+        return -1;
+    }
+    int getpeername(char* path, size_t count) override {
+        errno = ENOSYS;
+        return -1;
+    }
+};
+
+extern "C" photon::net::ISocketClient* new_rsocket_client() {
```
lihuiba, line 516: no need for photon::net::

net/rsocket/rsocket.h
```diff
@@ -0,0 +1,6 @@
...
+extern "C" photon::net::ISocketClient* new_rsocket_client();
+extern "C" photon::net::ISocketServer* new_rsocket_server();
```
lihuiba, line 6: no need for photon::net::

lihuiba (approved):

PR #667, opened 2024-12-17
semaphore plus

lihuiba:
(1) add an option for in-order resume of threads (by default not); 
(2) avoid thundering herd of threads whose needs are not met;

### DIFF
--- a/thread/thread.cpp
+++ b/thread/thread.cpp
@@ -1749,44 +1749,52 @@ insert_list:
     {
         if (count == 0) return 0;
         splock.lock();
-        CURRENT->semaphore_count = count;
-        int ret = 0;
+        DEFER(splock.unlock());
+        auto& counter = CURRENT->semaphore_count;
+        counter = count;
+        DEFER(counter = 0);
         while (!try_subtract(count)) {
-            ret = waitq::wait_defer(timeout, spinlock_unlock, &splock);
-            ERRNO err;
-            splock.lock();
-            if (ret < 0) {
-                CURRENT->semaphore_count = 0;
-                // when timeout, we need to try to resume next thread(s) in q
-                if (err.no == ETIMEDOUT) try_resume();
-                splock.unlock();
-                errno = err.no;
+            int ret = waitq::wait_defer(timeout, spinlock_unlock, &splock);
+            splock.lock();  // assuming errno NOT changed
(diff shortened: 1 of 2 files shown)


thread/thread.h
```diff
@@ -432,7 +432,8 @@ namespace photon
     class semaphore : protected waitq
     {
     public:
-        explicit semaphore(uint64_t count = 0) : m_count(count) { }
+        explicit semaphore(uint64_t count = 0, bool in_order_resume = false)
```
beef9999, line 435: Remove explicit?

lihuiba: why? if removed, an int might be unexpectedly converted to semaphore.

thread/thread.cpp
```diff
@@ -1795,6 +1790,22 @@ R"(
                 return true;
         }
     }
+    int semaphore::signal(uint64_t count) {
+        if (count == 0) return 0;
+        SCOPED_LOCK(splock);
+        auto cnt = m_count.fetch_add(count) + count;
+        if (!q.th) return 0;
+        if (cnt >= q.th->semaphore_count) {
+            do_resume: resume_one();
```
beef9999, line 1799: `goto do_resume;` here, and move `do_resume: resume_one();` to the end of the function?

lihuiba: The following loop depends on the condition that ```q.th``` has been checked and processed. The loop starts from the 2nd element, and uses ```q.th``` as the end.

thread/thread.cpp
```diff
@@ -1749,30 +1749,27 @@ R"(
...
             ret = waitq::wait_defer(timeout, spinlock_unlock, &splock);
-            ERRNO err;
-            splock.lock();
-            if (ret < 0) {
+            splock.lock();  // assuming errno NOT changed
+            if (unlikely(ret < 0)) {    // got interrupted
                 CURRENT->semaphore_count = 0;
                 // when timeout, we need to try to resume next thread(s) in q
+                ERRNO err;
                 if (err.no == ETIMEDOUT) try_resume();
```
Coldwings, line 1762: Once a wait timeout expired, it will always try resume no matter the `in_order_resume` is.

Is that works as expected?

Coldwings (approved):
LGTM

PR #670, opened 2024-12-17
fast case conversion and comparison

lihuiba:
comparison is 4+ times faster on MBP M1:
```
[ RUN      ] tolowerupper.perf_strncasecmp
[       OK ] tolowerupper.perf_strncasecmp (3270 ms)
[ RUN      ] tolowerupper.perf_photon__stricmp
[       OK ] tolowerupper.perf_photon__stricmp (777 ms)
```


Coldwings (approved):
LGTM

PR #683, opened 2024-12-23
iouring sq poll and io poll

lihuiba:


io/iouring-wrapper.cpp
```diff
@@ -198,6 +241,10 @@ class iouringEngine : public MasterEventEngine, public CascadingEventEngine, pub
             io_uring_sqe_set_data(sqe, &timer_ctx);
         }
 
+        int ret = io_uring_submit(m_ring);
+        if (ret < 0)
+            LOG_ERROR_RETURN(-ret, -1, "iouring: fail to submit when adding interest, ", ERRNO(-ret));
+
```
beef9999, line 247: We used to only submit IO in the reap process, in order to reduce the number of syscalls.

Will it cause performance decline?

photon.h
```diff
@@ -40,17 +43,21 @@ const uint64_t INIT_IO_EXPORTFS = SHIFT(10);
...
+                                    INIT_EVENT_SIGNAL;
+const uint64_t INIT_IO_DEFAULT    = INIT_IO_LIBCURL;
 #endif
 
 #undef SHIFT
 
 struct PhotonOptions {
     int libaio_queue_depth = 32;
+    uint32_t iouring_sq_thread_cpu;
+    uint32_t iouring_sq_thread_idle = 1000;     // by default polls for 1s
```
beef9999, line 60: Rename to `iouring_sq_thread_idle_ms` ?

io/iouring-wrapper.cpp
lihuiba: We already have it in async_io() and rm_interest(). It might be missing here?

io/iouring-wrapper.cpp
```diff
@@ -664,18 +698,17 @@ int iouring_unregister_files(int fd) {
...
-    LOG_INFO("Init event engine: iouring ", VALUE(is_master));
-    return NewObj<iouringEngine>(is_master) -> init();
+void* new_iouring_event_engine(iouring_args args) {
+    LOG_INFO("Init event engine: iouring ",
+        make_named_value("is_master",     args.is_master),
+        make_named_value("setup_sqpoll",  args.setup_sqpoll),
+        make_named_value("setup_sq_aff",  args.setup_sq_aff),
+        make_named_value("sq_thread_cpu", args.sq_thread_cpu));
+    auto uring = NewObj<iouringEngine>() -> init(args);
+    if (args.is_master) return uring;
```
beef9999, line 708: uring's type is MasterEventEngine? Don't need to do conversion?

beef9999 (approved):

io/iouring-wrapper.cpp
lihuiba: no need for the 1st parent class

PR #690, opened 2025-01-01
support for crc64-ecma (reflected), compatible with OSS

lihuiba:
Support for crc64-ecma (reflected), compatible with OSS.
The implementations are from Mark Adler (sw) and ARM official (hw for arm).

There is also a refactory of crc32c, and UT of crc.
The UT of crc64ecma has been verified with ossutil.


common/checksum/crc32c.h
```diff
@@ -15,28 +15,32 @@ limitations under the License.
...
+}
 
-inline uint32_t crc32c(const std::string &text) {
-  return crc32c_extend(text.data(), text.size(), 0);
+inline uint32_t crc32c(const void *data, size_t nbytes) {
+  return crc32c_extend((uint8_t*)data, nbytes, 0);
 }
 
-inline uint32_t crc32c_extend(const std::string &text, uint32_t crc) {
+inline uint32_t crc32c_extend(std::string_view text, uint32_t crc = 0) {
```
Coldwings, line 39: `crc32c_extend` now able to call with single string view with default `crc=0`, do things that original `crc32c` do.

why not make crc32c_extend with no default crc parameter and keeps origin crc32c with string_view?

beef9999:
What's the crc64's performance like, compared to crc32, and when shall we use it?

common/checksum/crc32c.cpp
```diff
@@ -15,21 +15,37 @@
...
+#if defined(__x86_64__)
   __builtin_cpu_init();
-  if (__builtin_cpu_supports("sse4.2")) {
-    crc32c_func = crc32c_hw;
-  } else {
-    crc32c_func = crc32c_sw;
-  }
-#elif (defined(__aarch64__) && defined(__ARM_FEATURE_CRC32))
-  crc32c_func = crc32c_hw;
+  bool hw = __builtin_cpu_supports("sse4.2");
```
beef9999, line 31: If the cpu already supported sse4, do we still need to add `-msse4.2` in the gcc command line?

lihuiba: Yes, otherwise gcc doesn't support the crc32c instruction.

common/checksum/crc32c.h
lihuiba: add back the missing function

Coldwings (approved):
LGTM

PR #698, opened 2025-01-16
fix potential race condition in DefaultResolver when discard_cache()

lihuiba:
delete_all() and recycle() is not atomic; and there may be read-write race as well.

### DIFF
--- a/net/utils.cpp
+++ b/net/utils.cpp
@@ -260,17 +260,12 @@ protected:
         IPAddr addr;
         IPAddrNode(IPAddr addr) : addr(addr) {}
     };
-    using IPAddrList = intrusive_list<IPAddrNode>;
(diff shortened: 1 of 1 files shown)


net/utils.cpp
```diff
@@ -304,6 +300,7 @@ class DefaultResolver : public Resolver {
 
     void discard_cache(const char *host, IPAddr ip) override {
         auto ipaddr = dnscache_.borrow(host);
+        scoped_rwlock _(*ipaddr, WLOCK);
         if (ip.undefined() || ipaddr->empty()) {
             ipaddr->delete_all();
```
Coldwings, line 305: recycle will destruct ipaddrlist so do not need to delete items right here, and also do not need a write lock in such condition.

PR #700, opened 2025-01-16
reimplement crc64ecma_sw to reduce redundancy

lihuiba:
Reimplement crc64ecma_sw to reduce redundant code, by unifying crc32 and crc64.
And make the table generated dynamically at runtime, so as to reduce redundant static data that is rarely used today, given that crc hw is common.

```
 common/checksum/{crc32c.cpp => crc.cpp} | 459 +++++++++-------------------------------------------------------------------------
 common/checksum/crc64ecma.cpp           | 221 ----------------------------------------
```


common/checksum/crc.cpp
```diff
@@ -767,407 +762,54 @@ uint64_t crc64ecma_hw(const uint8_t *buf, size_t len, uint64_t crc) {
     return crc64ecma_hw_portable(buf, len, crc);
 }
 
+template<typename T, T POLY>
```
Coldwings, line 765: The table is able to create in compile-time, by using template.

why create it in runtime?

lihuiba: that is rarely used today, given that crc hw is common.

lihuiba: it's a waste in both compile time and binary size to generate it in compile-time.

Coldwings (approved):

PR #702, opened 2025-01-23
HTTP Body read should able to accept long iovecs

Coldwings:
HTTP Body read stream used deal with io-vector with `IOVector`, which is `IOVectorEntity<32,4>`, has only 28 slots for iovec array.

Using `SmartCloneIOV` with `iovector_view` can accept iovec array with any length.


liulanzheng (approved):
LGTM

lihuiba (approved):

PR #703, opened 2025-01-31
crc64_with_avx512

lihuiba:
|     | 128KB Data (Cached) | 1GB Data (Non-Cached)|
| -------- | ------- | ------- |
| AVX512-cpp  | 48.47 GB/s    | 13.88 GB/s|
| SSE-cpp         | 25.86 GB/s    |12.23 GB/s|
| SSE-asm                | 25.81 GB/s    |11.99 GB/s|

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,7 +76,7 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_CXX_FLAGS_DEBUG "-O0 -g")
 set(CMAKE_CXX_FLAGS_RELEASE "-O2 -DNDEBUG")
 set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O2 -DNDEBUG -g")
-set(CMAKE_CXX_FLAGS_MINSIZEREL "-O2")    # Only for CI test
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -86,7 +86,7 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
 endif()
 
 if (${ARCH} STREQUAL x86_64)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -msse4.2 -mpclmul")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -msse4.2 -mpclmul -march=native")
```
beef9999, line 89: Will this change brings portability issues? native arch will enable all CPU features that the local host have.

Maybe we can add a new build option, for example `PORTABLE=false`

lihuiba: changed to specific flags

lihuiba: changed to optional specific flags

lihuiba: removed the flags

Coldwings (approved):

PR #716, opened 2025-02-17
minor update for http url and headers

liulanzheng:
- Change HeaderAssistant to dictionary order sorting, as some scenarios require traversing headers in dictionary order, such as OSS signature v4.
- Add a parameter to url_escape, as sometimes / does not need to be escaped.

### DIFF
--- a/net/http/url.cpp
+++ b/net/http/url.cpp
@@ -103,13 +103,15 @@ static bool isunreserved(char c) {
     return false;
 }
 
-std::string url_escape(std::string_view url) {
+std::string url_escape(std::string_view url, bool escape_slash) {
     static const char hex[] = "0123456789ABCDEF";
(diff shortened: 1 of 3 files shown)


net/http/url.cpp
```diff
@@ -103,14 +103,16 @@ static bool isunreserved(char c) {
     return false;
 }
 
-std::string url_escape(std::string_view url) {
+std::string url_escape(std::string_view url, bool escape_slash) {
     static const char hex[] = "0123456789ABCDEF";
-    std::string ret;
+    estring ret;
```
lihuiba, line 108: 这里类型跟返回类型保持一致，可以启用RVO优化

liulanzheng: 笔误😂

PR #729, opened 2025-03-11
Fix curl timeout return code handling

coding-nobody:
If curl time out, msg will return CURLMSG_DONE with error code in data result.
CURLMSG_DONE only stands for message transfer is done.
https://curl.se/libcurl/c/curl_multi_info_read.html

image

### DIFF
--- a/net/curl.cpp
+++ b/net/curl.cpp
@@ -48,6 +48,11 @@ struct CurlCallCtx {
     bool canceled = false;
 };
 
+struct CurlRequestPrivateData {
+    photon::semaphore sem{0};
+    CURLcode ret_code = CURLcode::CURLE_OK;
+};
+
 __thread CurlThCtx cctx;
 static CURLcode global_initialized;
 static int do_action(curl_socket_t fd, int event) {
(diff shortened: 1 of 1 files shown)


beef9999:
@coding-nobody How to reproduce this bug? I want to add a test case using libcurl client and Photon's HTTP server.

coding-nobody:
In our scenario, we are using photon::net::curl_perform to download data from object storage.
We have noticed when our object storage hit bandwidth limit or object storage not response data in time, curl_perform will download data with insufficient buffer length sometimes.
We add some log here to print ret_code in photon::net::do_action and we get 28(CURLE_OPERATION_TIMEDOUT).
So we can confirm that 28 error code is not returning to curl_perform and not bring error code back to our caller function.

image

We set object storage bandwidth to a very limited value to reproduce this error.

lihuiba:
Thanks for pointing it out the issue. We'll deal with it ASAP.

BTW, which branch are you using? We'll make sure that branch gets fixed.

lihuiba:
BTW again, you may want to check out and try Photon's implementation of HTTP client/server. It's more efficient.

lihuiba (approved):

coding-nobody:
> Thanks for pointing it out the issue. We'll deal with it ASAP.
> 
> BTW, which branch are you using? We'll make sure that branch gets fixed.

We are using a forked branch based on tag v0.7.10. I will cherry-pick your fix to our branch. Thanks for fixing this problem!

coding-nobody:
> BTW again, you may want to check out and try Photon's new implementation of HTTP client/server. It's more efficient.

I will check it later, thanks 😄

lihuiba:
@coding-nobody Thanks for your fix. I appreciate it very much!

(I had though it was an issue, before I realized it was actually a PR)

PR #731, opened 2025-03-12
[feat.] implement extfs::get_underlay_object()

BigVan:
Implement the method 'ExtFileSystem::get_underlay_object()' to return its underlay file

### DIFF
--- a/fs/extfs/extfs.cpp
+++ b/fs/extfs/extfs.cpp
@@ -1288,7 +1288,7 @@ class ExtFileSystem : public photon::fs::IFileSystem, public photon::fs::IFileSy
 public:
     ext2_filsys fs;
     io_manager extfs_io_manager;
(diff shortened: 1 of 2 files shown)


fs/extfs/extfs.cpp
```diff
@@ -1491,9 +1491,15 @@ class ExtFileSystem : public photon::fs::IFileSystem, public photon::fs::IFileSy
         return buffer_file ? buffer_file->fdatasync() : 0;
     }
 
+    Object* get_underlay_object(int i = 0) override {
+        return base_file;
```
lihuiba, line 1495: does buffer_file have get_underlay_object()?

BigVan: not yet

lihuiba (approved):

PR #745, opened 2025-03-26
Allow init Workpool from non-photon threads(std::threads); Increase ring size of Workpool

beef9999:
1. Similar to https://github.com/alibaba/PhotonLibOS/pull/735.  Allow init Workpool from non-photon threads
2. Increase ring size of Workpool. Avoid slow task run out of space of ring buffer.

### DIFF
--- a/thread/workerpool.cpp
+++ b/thread/workerpool.cpp
@@ -31,7 +31,7 @@ namespace photon {
 
 class WorkPool::impl {
 public:
-    static constexpr uint32_t RING_SIZE = 256;
+    static constexpr uint32_t RING_SIZE = 65536;
     static constexpr uint64_t QUEUE_YIELD_COUNT = 256;
     static constexpr uint64_t QUEUE_YIELD_US = 1024;
 
@@ -40,7 +40,6 @@ public:
     std::vector<photon::vcpu_base *> vcpus;
     std::atomic<uint64_t> vcpu_index{0};
     photon::semaphore queue_sem;
-    photon::semaphore ready_vcpu;
     photon::condition_variable exit_cv;
     photon::common::RingChannel<
         LockfreeMPMCRingQueue<Delegate<void>, RING_SIZE>>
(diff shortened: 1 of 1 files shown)


thread/workerpool.cpp
```diff
@@ -40,21 +41,24 @@ class WorkPool::impl {
...
+        : queue_sem(0), mode(mode) {
         for (size_t i = 0; i < vcpu_num; ++i) {
             owned_std_threads.emplace_back(
                 &WorkPool::impl::worker_thread_routine, this, ev_engine,
                 io_engine);
         }
-        ready_vcpu.wait(vcpu_num);
+        {
+            std::unique_lock<std::mutex> lock(vcpu_ready_mu);
+            vcpu_ready_cv.wait(lock, [&] { return vcpu_ready_num >= vcpu_num; });
```
lihuiba, line 60: ```vcpu_ready_num``` can be an atomic<size_t> variable in function-local scope, and get passed to new std::threads via its pointer. In fact, ```vcpu_ready_mu``` and even ```vcpu_ready_cv`` can be local too. Possibly make them a struct.

thread/workerpool.cpp
```diff
@@ -40,21 +41,24 @@ class WorkPool::impl {
     std::vector<photon::vcpu_base *> vcpus;
     std::atomic<uint64_t> vcpu_index{0};
     photon::semaphore queue_sem;
-    photon::semaphore ready_vcpu;
     photon::condition_variable exit_cv;
-    photon::common::RingChannel<
```
lihuiba, line 45: don't change code due to coding style only.

thread/workerpool.cpp
beef9999: How about encapsulate a new semaphore inside the WorkPool class?

```c++
class SemaphoreWrapper {
public:
    Semaphore(int count = 0) : count_(count) {}

    void notify() {
        std::unique_lock<std::mutex> lock(mutex_);
        ++count_;
        cv_.notify_one();
    }

    void wait(int n = 1) {
        std::unique_lock<std::mutex> lock(mutex_);
        cv_.wait(lock, [this, n] { return count_ >= n; });
        count_ -= n;
    }

private:
    std::mutex mutex_;
    std::condition_variable cv_;
    int count_;
};
```

lihuiba: Good idea! And ```atomic<size_t>``` is also better in this case.

lihuiba (approved):

PR #749, opened 2025-04-01
fix: -Wvla-cxx-extension errors

tzcnt:
These lines both cause a warning on the following line, which aborts the build due to -Werror:
`error: variable length arrays in C++ are a Clang extension [-Werror,-Wvla-cxx-extension]`

This happens even though I am building with Clang (19.1.6).

Making them `const` resolves the error.


beef9999:
Thanks for contribution. What is your OS type and version by the way?

lihuiba (approved):

PR #750, opened 2025-04-03
RPC: Able to change socket type; Improve doc;  Example support echo perf test

beef9999:
1. Able to change RPC socket type
2. Improve doc of rpc.h
3. RPC example now support echo_perf option, able to run performance test loop.
4. Fix example bind to localhost bug, cannot be tested on different hosts.

### DIFF
--- a/examples/perf/net-perf.cpp
+++ b/examples/perf/net-perf.cpp
@@ -215,7 +215,7 @@ static int echo_server() {
     photon::thread_enable_join(stop_th);
 
     server->set_handler(handler);
-    server->bind_v4localhost(FLAGS_port);
+    server->bind_v4any(FLAGS_port);
     server->listen();
     server->start_loop(true);
 
--- a/examples/rpc/client_main.cpp
+++ b/examples/rpc/client_main.cpp
@@ -14,38 +14,59 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+#include <chrono>
 #include <gflags/gflags.h>
 #include <photon/common/alog-stdstring.h>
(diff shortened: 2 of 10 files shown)


examples/perf/net-perf.cpp
```diff
@@ -215,7 +215,7 @@ static int echo_server() {
     photon::thread_enable_join(stop_th);
 
     server->set_handler(handler);
-    server->bind_v4localhost(FLAGS_port);
+    server->bind_v4any(FLAGS_port);
```
beef9999, line 218: v4 localhost listens 127.0.0.1. So cannot be tested on different hosts.

examples/rpc/client_main.cpp
```diff
@@ -14,31 +14,53 @@ See the License for the specific language governing permissions and
 limitations under the License.
```
beef9999, line 14: 像net-perf程序一样，引入持续运行的echo perf评测RPC性能

rpc/rpc.h
```diff
@@ -244,95 +264,25 @@ namespace rpc
...
     extern "C" StubPool* new_uds_stub_pool(const char* path,
                                            uint64_t expiration,
                                            uint64_t connect_timeout);
+
     extern "C" Skeleton* new_skeleton(uint32_t pool_size = 128);
 
     __attribute__((deprecated))
     inline Skeleton* new_skeleton(bool /*concurrent*/, uint32_t pool_size) {
         return new_skeleton(pool_size);
     }
```
beef9999, line 286: 在example里面已有示例，故删除

lihuiba (approved):

PR #760, opened 2025-04-15
[feat.] Add cache module

BigVan:
Moving the cache module from containerd/overlaybd to Photon

### DIFF
--- a/cache/ocf_cache/include/ocf
+++ b/cache/ocf_cache/include/ocf
@@ -1,0 +1,1 @@
+../../../build/_deps/ocf_lib-src/inc/
--- a/cache/ocf_cache/photon_bindings/volume.h
+++ b/cache/ocf_cache/photon_bindings/volume.h
@@ -1,0 +1,36 @@
+/*
+Copyright 2022 The Photon Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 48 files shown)


cache/ocf_cache/include/ocf
```diff
@@ -0,0 +1 @@
+../../../build/_deps/ocf_lib-src/inc/
```
beef9999, line 1: Link to src dir instead of build dir?

BigVan: We use 'fetchContent' instead of pulling submodule, so 'ocf_lib-src' is the subdir of 'build/_deps'....

lihuiba: how about using ```-I ../../../build/_deps/ocf_lib-src/inc/``` instead of a symbol link?

cache/ocf_cache/photon_bindings/volume.h
```diff
@@ -0,0 +1,36 @@
...
+
+#pragma once
+
+#include <photon/fs/filesystem.h>
+
+extern "C" {
+#include <ocf/ocf.h>
+}
+
+#define EASE_OCF_VOLUME_TYPE 1
```
lihuiba, line 25: how about removing prefixes like "EASE_" or "ease_"

EricHuangqx (approved):

beef9999 (approved):

lihuiba (approved):

PR #767, opened 2025-04-17
Set TCP_NODELAY by default

Coldwings:
TCPSocket before release/0.8 always set TCP_NODELAY by default, there fore http client/server and tcp client/server are always goes with TCP_NODELAY.

Somehow, last year, a socket refactoring changed the behavior. That leads to long 40ms latency kept happen in TCP related io.

Fix it by changing default initialize parameters.

Thanks @f0cii helps found and fix this problem.

Co-authored-by: f0cii <2201441955@qq.com>


lihuiba:
@f0cii Thanks for pointing it out!

PR #812, opened 2025-06-07
FIX autopr when commit message contains quota ' character

Coldwings:


lihuiba (approved):

Coldwings:
seems macos CI for release/0.7 already broken since homebrew gtest upgraded.

PR #835, opened 2025-06-20
fuse-adaptor: Implement the sync io mode 

haolianglh:
This new mode using non block read, block read and epoll to fetch fuse request from fuse device


CMakeLists.txt
```diff
@@ -225,7 +225,11 @@ else ()
     endif ()
 endif ()
 if (PHOTON_ENABLE_FUSE STREQUAL "ON" OR PHOTON_ENABLE_FUSE STREQUAL "2" OR PHOTON_ENABLE_FUSE STREQUAL "3")
-    list(APPEND PHOTON_SRC fs/fuse_adaptor.cpp)
+    list(APPEND PHOTON_SRC
+         fs/fuse_adaptor/fuse_adaptor.cpp
```
lihuiba: how about ```fs/fuse_adaptor/*.cpp``` ?

fs/fuse_adaptor/session_loop_impl.cpp
lihuiba, line 1: simply name it session_loop.cpp

fs/fuse_adaptor/session_loop_impl.cpp
```diff
@@ -0,0 +1,425 @@
...
+          se_(se),
+          num_worker_(0) {
+          max_workers_ = 32;
+          numavail_ = 0;
+          waitting_ = false;
+
+        for (int i = 0; i < max_workers_; ++i) {
+            WorkerArgs<void *, int> *wrkargs = new WorkerArgs<void *, int>();
+            wrkargs->args = std::make_tuple(reinterpret_cast<void *>(this), i);
+            auto th = photon::thread_create(
```
lihuiba: use ```thread_create11()``` to avoid WorkerArgs and static function

fs/fuse_adaptor/session_loop_impl.cpp
```diff
@@ -0,0 +1,425 @@
...
+
+    explicit SyncSessionLoop(struct fuse_session *se)
+        : error_(0),
+          se_(se),
+          num_worker_(0) {
+          max_workers_ = 32;
+          numavail_ = 0;
+          waitting_ = false;
+
+        for (int i = 0; i < max_workers_; ++i) {
```
lihuiba: Constructor should be simple. Put these work in init() or run().

lihuiba: Constructor should be simple to avoid failures. Put these work in init() or run().

fs/fuse_adaptor/session_loop_impl.cpp
```diff
@@ -0,0 +1,425 @@
...
+            auto th = workers_.back();
+            photon::thread_join((photon::join_handle *)th);
+            workers_.pop_back();
+            num_worker_--;
+        }
+        assert(num_worker_ = 0); 
+    }
+
+    ~SyncSessionLoop() {
+        if (se_) fuse_session_exit(se_);
```
lihuiba: need to close the file descriptors?

fs/fuse_adaptor/session_loop_impl.cpp
```diff
@@ -0,0 +1,425 @@
...
+
+        for (int i = 0; i < (int)(workers_.size()); ++i) {
+            auto th = workers_.back();
```
lihuiba: Duplicated code in run(). Extract common function.

fs/fuse_adaptor/session_loop_impl.cpp
```diff
@@ -0,0 +1,425 @@
...
+        return fuse_session_custom_io(se, &custom_io, fuse_session_fd(se));
+#endif
+    }
+
+private:
+    int error_;
+    struct fuse_session *se_;
+    int blk_fd_;
+    int nonblk_fd_;
+    int num_worker_;
```
lihuiba: num_worker_ is not necessary as it is always equal to workers_.size()

fs/fuse_adaptor/session_loop_impl.cpp
```diff
@@ -0,0 +1,425 @@
...
+        loop->idlers_.erase(idx);
+        while(!fuse_session_exited(se)) {
+            if (loop->numavail_ == loop->num_worker_ -1) {
+                *iofd = loop->blk_fd_;
+                int res = fuse_session_receive_buf(se, &fbuf);
+                if (res <= 0) {
+                    break;
+                }
+
+                if (loop->idlers_.empty()) {
```
lihuiba: use idlers_ when it is empty()?

fs/fuse_adaptor/session_loop_impl.cpp
```diff
@@ -0,0 +1,425 @@
...
+                            if (errno == EALREADY) {
+                                photon::thread_usleep(-1);                        
+                            }
+                        }
+                    }
+                    --(loop->numavail_);
+                    loop->idlers_.erase(idx);
+                } else {
+                    if (!loop->idlers_.empty()) {
+                        photon::thread_interrupt(loop->workers_[loop->idlers_.begin()->first]);
```
lihuiba: It seems you don't use its ```second```? Use unordered_set instead.

fs/fuse_adaptor/session_loop_impl.cpp
```diff
@@ -0,0 +1,425 @@
...
+                        loop->waitting_ = true;
+                        auto ret = photon::wait_for_fd_readable(loop->nonblk_fd_);
+                        loop->waitting_ = false;
+                        if (ret < 0) {
+                            if (errno == EALREADY) {
+                                photon::thread_usleep(-1);                        
+                            }
+                        }
+                    }
+                    --(loop->numavail_);
```
lihuiba: ```loop->numavail_--``` is more concise.

fs/fuse_adaptor/session_loop_impl.cpp
```diff
@@ -0,0 +1,425 @@
...
+
+        LOG_INFO("masterfd:`", masterfd, " block fd `", blk_fd_, "  nonblock fd: `", nonblk_fd_);
+        return 0;
+    }
+
+    explicit SyncSessionLoop(struct fuse_session *se)
+        : error_(0),
+          se_(se),
+          num_worker_(0) {
+          max_workers_ = 32;
```
lihuiba: assign the constants directly at declaration.

fs/fuse_adaptor/session_loop_impl.cpp
```diff
@@ -0,0 +1,425 @@
...
+        struct fuse_session *se = loop->se_;
+        struct fuse_buf fbuf = {
+            .mem = NULL,
+        };
+
+        loop->sem_.wait(1);
+        --loop->numavail_;
+        loop->idlers_.erase(idx);
+        while(!fuse_session_exited(se)) {
+            if (loop->numavail_ == loop->num_worker_ -1) {
```
lihuiba: This seems to be the only usage of numavail_? If you want to make sure there is one and only one thread running this case, simply use a boolean value to indicate whether there is one running this case.

lihuiba: This seems to be the only usage of numavail_? If you want to make sure there is one and only one thread running this case, simply use a boolean value to indicate whether there is one running this case.

haolianglh: Yes, the only purpose of numavail_ is to determine whether there is currently a single worker active. Logically this is a boolean, but it requires some condition to determine its true or false value, specifically the count of idle workers. In reality, it's checking the size of idlers_. For sake of code simplification, it may be better to directly use idlers_.size() for this judgment.

fs/fuse_adaptor/session_loop_impl.cpp
haolianglh: ok

fs/fuse_adaptor/session_loop_impl.cpp
haolianglh: OK, i will remove it

fs/fuse_adaptor/session_loop_impl.cpp
haolianglh: This is a typo. I will correct it.

fs/fuse_adaptor/session_loop_impl.cpp
haolianglh: numavail_ will be removed

fs/fuse_adaptor/session_loop_impl.cpp
haolianglh: ok

fs/fuse_adaptor/session_loop_impl.cpp
haolianglh: ok

fs/fuse_adaptor/session_loop_impl.cpp
haolianglh: ok

fs/fuse_adaptor/session_loop_impl.cpp
haolianglh: ok

CMakeLists.txt
haolianglh: ok

fs/fuse_adaptor/session_loop_impl.cpp
haolianglh: done

PR #848, opened 2025-07-02
io_uring cascading engine supports for I/O

lihuiba:
It is observed that, when using io_uring to interact with libfuse, it's more performant to submit I/O requests to ```fuse_dev_fd``` without blocking on io_uring. So we need to use cascading engine to issue I/O and block on eventfd and epoll master engine.

### DIFF
--- a/io/iouring-wrapper.cpp
+++ b/io/iouring-wrapper.cpp
@@ -241,6 +241,8 @@ public:
             io_uring_sqe_set_data(sqe, &timer_ctx);
         }
 
+        if (try_submit() < 0) return -1;
+
         SCOPED_PAUSE_WORK_STEALING;
         photon::thread_sleep(-1);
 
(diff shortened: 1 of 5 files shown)


io/iouring-wrapper.cpp
```diff
@@ -406,31 +369,66 @@ class iouringEngine : public MasterEventEngine, public CascadingEventEngine, pub
...
+
+            } else /* if (ctx->is_event) */ {
+
+                if (num >= count) break;
+                eventCtx* event_ctx = container_of(ctx, eventCtx, io_ctx);
+                fdInterest fd_interest{event_ctx->event.fd, (uint32_t)evmap.translate_bitwisely(event_ctx->event.interests)};
+                if (ctx->res == -ECANCELED) {
+                    m_event_contexts.erase(fd_interest);
+                } else if (event_ctx->one_shot) {
+                    data[num++] = event_ctx->event.data;
```
haolianglh, line 408: So, does this require the caller to ensure that the memory pointed to by ’data‘ is valid when ’ctx->is_event‘ is true?

lihuiba: yes

Coldwings (approved):
LGTM

PR #884, opened 2025-07-21
workpool minor refinement

lihuiba:


thread/workerpool.cpp
```diff
@@ -39,15 +39,12 @@ class WorkPool::impl {
     std::vector<std::thread> owned_std_threads;
```
Coldwings, line 39: `worker_mtx` still depending on photon environment since it is a `photon::mutex`

Seems using `photon::spin_lock` instead should works fine and no more have to be in photon environment.

PR #898, opened 2025-07-23
reimplement AsyncLogOutput to avoid malloc/free

lihuiba:


common/lockfree_queue.h
```diff
@@ -440,35 +433,54 @@ class LockfreeSPSCRingQueue : public LockfreeRingQueueBase<T, N> {
...
+    size_t produce_push_batch(size_t n, Producer&& produce) {
         auto t = tail.load(std::memory_order_relaxed);
         n = std::min(
             n, Base::capacity - (t - head.load(std::memory_order_acquire)));
         if (n == 0) return 0;
         auto first_idx = idx(t);
         auto part_length = Base::capacity - first_idx;
         if (likely(part_length >= n)) {
-            memcpy(&slots[first_idx], x, sizeof(T) * n);
+            produce(&slots[first_idx], n, nullptr, 0);
```
Coldwings, line 453: Use `struct iov*, int` as parameter for `producer` `consumer` might be better than two pointers and segment size.
Both `LockfreeBatchMPMCQueue` and `LockfreeSPSCQueue` suits for such kind of batch API.

common/lockfree_queue.h
```diff
@@ -440,35 +433,54 @@ class LockfreeSPSCRingQueue : public LockfreeRingQueueBase<T, N> {
...
+            if (n2) { memcpy(p2, x, n2 * sizeof(T)); }
+        });
+    }
+
+    // Producer must be callable as produce(T* offset1, size_t n, T* offset2, size_t n),
+    // and produce (write) data directly to the provided buffer.
+    template<typename Producer>
+    size_t produce_push_batch(size_t n, Producer&& produce) {
         auto t = tail.load(std::memory_order_relaxed);
         n = std::min(
```
Coldwings, line 447: push_batch used able to push partial batch.

seems this is not fit the async logs

PR #904, opened 2025-07-28
Fix cross-compiling by using CMAKE_SYSTEM_PROCESSOR and respecting -march|-mcpu

chewi:
There is no need to check the architecture twice. Also check for arm64 (MacOS) wherever we check for aarch64 (Linux).

Packagers may specify a target system up front, which should be respected. The CRC extensions are not strictly necessary as there is a software fallback.

Otherwise, use `-mcpu=native` if it works (it won't when cross-compiling) or `-mcpu=generic+crc`.


Coldwings:
Thanks for the contribute. We do should considering when `-mcpu=native` not working. Seems this PR is a much better solution.

We are looking forward to merge it, still, need you to signing the CLA.

beef9999 (approved):

beef9999 (approved):
LGTM

Coldwings (approved):

PR #907, opened 2025-07-30
net-perf server listens to multiple ports; Refine semaphore doc

beef9999:


examples/perf/net-perf.cpp
```diff
@@ -40,11 +40,12 @@ DEFINE_uint64(vcpu_num, 1, "vCPU number for both client and server");
 DEFINE_string(socket_type, "tcp", "Support tcp/rsocket/io_uring");
```
beef9999, line 41: Changes:
1. global variables use atomic
2. server listens to individual ports for multiple vCPUs

lihuiba (approved):

PR #910, opened 2025-08-01
[fix #909] fix compile error

### LINKED ISSUE #909: Compile Error: error 'constexpr const bool __builtin_expect' conflits with a preview declaration
compiler: gcc-12
std: c++20

```
error: declaration of 'constexprconst bool _builtin_expect' conflicts with built-in declaration 'long int _builtin_expect(long int, long int)
315 | constexpr bool likely(bool expr) { return builtin_expect(expr, true); }
```

This conflict maybe from dpdk.

ivanallen:
fix #909 

'likely' and 'unlikey' is also defined in dpdk or other project. Declaration conflict occur when photonlibs and dpdk are compiled together.

### DIFF
--- a/common/utility.h
+++ b/common/utility.h
@@ -312,8 +312,13 @@ public:
 	}
 */
 
+#ifndef likely
 constexpr bool likely(bool expr) { return __builtin_expect(expr, true); }
+#endif /* likely */
+
+#ifndef unlikely
 constexpr bool unlikely(bool expr) { return __builtin_expect(expr, false); }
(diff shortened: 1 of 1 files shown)


zhanghan-mq (approved):

lihuiba:
Do you want to PR it to 0.8? (and the system will auto PR it to 0.9 and main)

ivanallen:
using this pr: https://github.com/alibaba/PhotonLibOS/pull/912

PR #927, opened 2025-08-06
Log rotation

yyaoling:
modify the log rotation method when program start

### DIFF
--- a/common/alog.cpp
+++ b/common/alog.cpp
@@ -433,7 +433,7 @@ public:
 };
 
 ILogOutput* new_log_output_file(const char* fn, uint64_t rotate_limit,
-                                int max_log_files, uint64_t throttle) {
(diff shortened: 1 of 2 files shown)


common/alog.cpp
```diff
@@ -433,21 +433,34 @@ class AsyncLogOutput final : public ILogOutput {
...
     return ret;
 }
 
-ILogOutput* new_log_output_file(int fd, uint64_t throttle) {
+ILogOutput* new_log_output_file(int fd, uint64_t throttle, bool rotate_on_start) {
     auto ret = new LogOutputFile();
     ret->log_output_file_setting(fd);
     ret->set_throttle(throttle);
+
+    // when init the new log output file, rotate the log files that last program created
```
lihuiba: without a file name, it's not possible to perform rotation

lihuiba (approved):

beef9999 (approved):

Issue #919, opened 2025-08-04
The flame graph is displayed abnormally


ktxdgtx:
We introduced Photon and intended to capture flame graphs using perf to analyze some performance issues. However, the flame graph stacks captured look quite strange. There are many "unknown" stacks, and even when the stacks are visible, the call relationships seem odd. 

unknown stack:

<img width="2388" height="1134" alt="Image" src="https://github.com/user-attachments/assets/77da20fa-342b-43f4-901e-7d37c7112da1" />

stack looks so odd

<img width="2888" height="720" alt="Image" src="https://github.com/user-attachments/assets/06b2e4f2-a805-4d55-8b29-eecb1e779634" />

ktxdgtx:
> As to the high proportion of [unknown], can you first try `perf top -p <pid>` with your program, and make sure you have build photon with debug-info(cmake)

Thanks, we compile photon with the compilation parameter `-g -ggdb3 -fno-omit-frame-pointer`. 

Using `perf top`, I can see a more normal performance dashboard. However, is there a way to generate a relatively normal flame graph? This is because the more commonly method we use to analyze performance issues is `perf record + FlameGraph`.

dusteye:
> The problem is caused by Thread11 using bottom of stack storing function and parameters, make it looks not like a formally thread stack.
> 
> It might have some solution (like add some separate spaces between paras and stack frame). We will try to fix that.

Does ‘Thread11’ here refer to the thread created via thread_create11 ?

lihuiba:
> Does ‘Thread11’ here refer to the thread created via thread_create11 ?
Yes, but I don't think that's the reason.

lihuiba:
Boost.Context also has a ```0x0000000000000000 in ?? ()``` frame in gdb, which I believe is the ```[unkonwn]``` in perf.

```
Breakpoint 1, 0x000000000040bcd0 in jump_fcontext ()
(gdb) bt
#0  0x000000000040bcd0 in jump_fcontext ()
#1  0x000000000040318f in void boost::context::detail::fiber_entry<boost::context::detail::fiber_record<boost::context::fiber, boost::context::basic_fixedsize_stack<boost::context::stack_traits>, test_move()::{lambda(boost::context::fiber&&)#1}> >(boost::context::detail::transfer_t) ()
#2  0x000000000040bcbf in make_fcontext ()
#3  0x0000000000000000 in ?? ()
(gdb)
```

What is the flame graph like?

[...]

PR #937, opened 2025-08-11
Prettify thread stack bottom fp for perf unwind with fp

Coldwings:
As mentioned in #919 , the flame-graph generated by `perf` with call-graph fp (default call-graph capture mode) and `-fno-omit-frame-pointer` may capture weird stack bottom which contains thread pointer as frame pointer.

Turns out that `_photon_thread_stub` using register `rbp` as thread pointer parameter, may break the assumption of frame pointer on stack.
Fix it by make both stack `rbp` and `rbp` register to zero, now stack bottom can be clear and simple when compiling with `-fno-omit-frame-pointer`

### DIFF
--- a/thread/thread.cpp
+++ b/thread/thread.cpp
@@ -754,10 +754,16 @@ R"(
 DEF_ASM_FUNC(_photon_thread_stub)
 R"(
         mov     %rbp, %rbx
-        mov     $0, %rbp
-        call    _asan_start
+        xor     %rbp, %rbp
+)"
+#ifdef __SANITIZE_ADDRESS__
+R"(
+        call _asan_start
+)"
+#endif
+R"(
         mov     0x40(%rbx), %rdi
-        movq    $0, 0x40(%rbx)
+        movq    %rbp, 0x40(%rbx)
         call    *0x48(%rbx)
         mov     %rax, 0x48(%rbx)
         mov     %rbx, %rdi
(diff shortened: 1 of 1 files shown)


thread/thread.cpp
```diff
@@ -753,12 +753,14 @@ R"(
 
 DEF_ASM_FUNC(_photon_thread_stub)
 R"(
+        mov     %rbp, %rbx
+        mov     $0, %rbp
```
lihuiba: xor %rbp, %rbp

thread/thread.cpp
```diff
@@ -753,12 +753,14 @@ R"(
...
         call    _asan_start
-        mov     0x40(%rbp), %rdi
-        movq    $0, 0x40(%rbp)
-        call    *0x48(%rbp)
-        mov     %rax, 0x48(%rbp)
-        mov     %rbp, %rdi
+        mov     0x40(%rbx), %rdi
+        movq    $0, 0x40(%rbx)
```
lihuiba: mov %rbp, 0x40(%rbx)

thread/thread.cpp
```diff
@@ -753,12 +753,14 @@ R"(
...
+        call    *0x48(%rbx)
+        mov     %rax, 0x48(%rbx)
+        mov     %rbx, %rdi
         call    _photon_thread_die
```
lihuiba, line 770: update ARM version of _photon_thread_stub() accordingly  as well

PR #969, opened 2025-08-29
make workpool's dtor consume less cpu if there's photon env

lihuiba:


Coldwings (approved):

thread/workerpool.cpp
```diff
@@ -53,10 +53,14 @@ class WorkPool::impl {
...
-    ~impl() { // avoid depending on photon to make it destructible wihout photon
+    ~impl() {
         for (auto num = vcpus.size(); num; --num) enqueue({});
         for (auto &worker : owned_std_threads) worker.join();
-        while (vcpus.size()) std::this_thread::yield();
+        if (likely(CURRENT)) {
+            while (vcpus.size()) thread_yield();
+        } else {
+            while (vcpus.size()) std::this_thread::yield();
+        }
```
beef9999, line 64: 或许我们可以用一个mutex保护vcpus，因为扩容vcpus可以看做一个低频操作，不太要求性能。然后在析构的时候可以用std::condition_variable。非Photon环境最好也能降低点cpu负载

-----

Perhaps we can use a mutex to protect the vcpus, because scaling vcpus can be regarded as a low-frequency operation and does not require much performance. Then, we can use std's condition variable during destruction. It's better to reduce the load a bit in non-photon environments as well.

lihuiba: ```vcpus``` is already protected with a spinlock. CPU usage is not important here. The aim is to make it run well with or without photon environment, and we have to treat the 2 situation separately.

PR #1010, opened 2025-10-02
fix iovector not checking size before extracting continuous buffer

NewbieOrange:
The comments for `extract_front_continuous` and `extract_back_continuous` indicate they should return nullptr if data are not enough. However this is not checked in code and result in crashes when deserializing invalid or malicious packets.

### DIFF
--- a/common/iovector.h
+++ b/common/iovector.h
@@ -532,7 +532,9 @@ public:
         }
         
         auto buf = do_malloc(bytes);
-        extract_front(bytes, buf);
+        if (buf) {
+            extract_front(bytes, buf);
+        }
         return buf;
     }
 
@@ -624,7 +626,9 @@ public:
         }
 
         auto buf = do_malloc(bytes);
-        extract_back(bytes, buf);
+        if (buf) {
+            extract_back(bytes, buf);
+        }
         return buf;
     }
 


common/iovector.h
```diff
@@ -522,6 +522,10 @@ class iovector
     void* extract_front_continuous(size_t bytes)
     {
         auto va = view();
+        if (va.sum() < bytes) {
```
NewbieOrange: This could potentially be checked after line 533, saving the call on the fast path. However it looks a bit weird and I'm not sure if it would make a difference in performance at all.

common/iovector.h
```diff
@@ -527,12 +527,13 @@ class iovector
...
+            return nullptr;
+        }
+        
         auto buf = do_malloc(bytes);
-        auto ret = extract_front(bytes, buf);
-        return ret == bytes ?
-            buf :
-            nullptr;
+        extract_front(bytes, buf);
```
NewbieOrange: extract_front should always have ret == bytes, should we assert it?

Coldwings: It seems better to check if allocation has failure (which means `buf` is a nullptr) than a assertion.

NewbieOrange: @Coldwings edited

PR #1012, opened 2025-10-09
fix deserialize returning invalid message on failure

NewbieOrange:
After #1010, `extract_front_continuous` now failed with invalid cases and returns `nullptr`, but `DeserializerIOV::deserialize` is missing the error handling path, and invoke rpc service with invalid message (`x._ptr == nullptr`)

Fixed by checking `failed` flag on return. `nullptr` is handled by rpc layer and will disconnect on invalid message instead of processing.

### DIFF
--- a/rpc/serialize.h
+++ b/rpc/serialize.h
@@ -436,6 +436,8 @@ namespace rpc
 
         void process_field(buffer& x)
         {
+            if (x.size() == 0)
+                return;
             x._ptr = _iov->extract_front_continuous(x.size());
             if (!x._ptr)
                 failed = true;


rpc/serialize.h
```diff
@@ -436,8 +436,6 @@ namespace rpc
 
         void process_field(buffer& x)
         {
-            if (x.size()==0)
```
NewbieOrange, line 439: btw, why does it return early if x.size() == 0?

does that mean a empty string buffer could have pointer as nullptr? it could be undefined behavior if user tried to memcpy or construct a std::string from a nullptr.

Coldwings: Notice that all Non-POD fields of message should be `photon::rpc::buffer`, string-like field should use `photon::rpc::string`.

When processing an empty string buffer, it will deserialized as a `photon::rpc::string` with buffer size 0.
the `photon::rpc::string` builds string via creating string_view, and a string_view constructor with character pointer and length accepting `nullptr` with 0 length is not a UB, or, at least UBSAN will not complain about it.

the photon::rpc::string constructing std::string is also constructed via a string_view, which is acceptable when ptr is NULL.

> https://en.cppreference.com/w/cpp/string/basic_string_view/basic_string_view.html
> 3) Constructs a view of the first count characters of the character array starting with the element pointed by s. s can contain null characters. The behavior is undefined if [s, s + count) is not a valid range (even though the constructor may not access any of the elements of this range). After construction, [data()](https://en.cppreference.com/w/cpp/string/basic_string_view/data.html) is equal to s, and [size()](https://en.cppreference.com/w/cpp/string/basic_string_view/size.html) is equal to count.

Without this early return, `photon::rpc::string` may not able to present a empty string, which has length 0 and no data will be deserialized as a failure.

Actually, not only strings. `photon::rpc::array` with zero length should not take as failure.

NewbieOrange: I am a bit confused, if `x.size() == 0`, shouldn't the following code continue as normal as extract a zero size buffer (e.g. ptr is a valid pointer, and size is 0)?

and it seems according to C spec, memcpy / memmove from nullptr even with size == 0 is undefined behaviour.

https://stackoverflow.com/questions/5243012/is-it-guaranteed-to-be-safe-to-perform-memcpy0-0-0

NewbieOrange: also the other `process_field` didn't check empty case, so this just aligns the behaviour.
```cpp
void process_field(iovec_array& x)
{
            iovector_view v;
            ssize_t ret = _iov->extract_front(x.summed_size, &v);
            if (ret == (ssize_t)x.summed_size) {
                x.assign(v.iov, v.iovcnt);
            } else {
                failed = true;
            }
}
```

lihuiba: > and it seems according to C spec, memcpy / memmove from nullptr even with size == 0 is undefined behaviour.

That means ```memcpy()``` would access the address beyond the end of the buffer(s). I don't think this will actually happen in *reality*, unless ```memcpy()``` has a bug.

NewbieOrange: > That means `memcpy()` would access the address beyond the end of the buffer(s). I don't think this will actually happen in _reality_, unless `memcpy()` has a bug.

That is probably true tbh, but it makes more sense to have a valid pointer than null pointer. (and the other `process_field(iovec_array& x)` already did that)

It doesn't hurt if we have a valid pointer, and just makes more sense, I suppose. Like an empty std::string doesn't return nullptr from .data()

lihuiba: @NewbieOrange The C standard text referenced in the above stackoverflow link *all* describes general cases, not specific to ```memcpy()```.

NewbieOrange: > @NewbieOrange The C standard text reference in the above stackoverflow page _all_ describes general cases, not specific to `memcpy()`.

I know, and I agree what you said (its probably fine to have nullptr). My point is that - it doesn't cost anything for us to return a valid pointer (extract_front_continous with bytes = 0 is basically as free as an early return), so why not?

lihuiba: > Like an empty std::string doesn't return nullptr from .data()

An empty ```string_view```, however, does return nullptr from .data().
```
% cat x.cpp
#include <string_view>
#include <stdio.h>

int main() {
	std::string_view s;
	printf("%p\n", s.data());
	return 0;
}

% g++ x.cpp && ./a.out
0x0
```

The implementations of ```std::string``` usually store short strings inside the string objects themselves. That's why empty strings doesn't return nullptr from .data().

NewbieOrange: @lihuiba it's a bit different. I would expect if I declared a rpc::buffer and its .data() is nullptr, but not when its passed from rpc service call. Like what I said above, `process_field(iovec_array& x)` already assign the iov without checking if size is zero, so why should `process_field(buffer& x)` check the size, shouldn't it be consistent across?

lihuiba: For a zero-length ```string_view``` or ```rpc::buffer``` or something similar, their .data() should be ignored, and you must *NOT* access the addresses.

Using a constant ```nullptr``` in the implementation is easier for the compilers to optimize code,  than using a variable obtained from a function call.

rpc/serialize.h
```diff
@@ -477,7 +473,7 @@ namespace rpc
             } else {
                 failed = true;
             }
-            return t;
+            return failed ? nullptr : t;
```
lihuiba, line 476: you can further move the ```return``` statement into the ```if { } else { }``` block, to make it more clear.

PR #1019, opened 2025-10-21
Fix thread-local PooledStackAllocator init order

beef9999:
`use_pooled_stack_allocator`可能在程序很早的地方被调用，然而，`PooledStackAllocator`是一个thread-local变量，由于是按需使用alloc/dealloc，所以这个变量的初始化也很晚。

在程序结束的时候，`PooledStackAllocator`会优先于其他thread-local模块析构，于是造成了这些thread-local模块（已经使用了它分配的内存）产生coredump。

这个MR在`use_pooled_stack_allocator`的时候就初始化thread-local变量，并且把 `set_thread_stack_allocator` 放到 `thread.h` 模块，简化一些 `stack-allocator.h` 的代码

-------------

use_pooled_stack_allocator may be called very early in the program. However, PooledStackAllocator is a thread-local variable, and due to the on-demand use of alloc/dealloc, the initialization of this variable is also very late.​
At the end of the program, PooledStackAllocator will be destructed before other thread-local modules, which causes some thread-local modules (that have used the memory allocated by it) to generate coredumps.​
This MR initializes the thread-local variable when use_pooled_stack_allocator is called, and moves `set_thread_stack_allocator` to the `thread.h` module to simplify some code in `stack-allocator.h`.

### DIFF
--- a/thread/stack-allocator.cpp
+++ b/thread/stack-allocator.cpp
@@ -18,6 +18,7 @@ limitations under the License.
 #include <linux/mman.h>
 #endif
 #include <errno.h>
+#include <mutex>
 #include <photon/common/alog.h>
 #include <photon/common/utility.h>
 #include <photon/thread/arch.h>
@@ -173,8 +174,8 @@ size_t pooled_stack_trim_threshold(size_t x) {
 }
 
 void use_pooled_stack_allocator() {
+    get_pooled_stack_allocator();
     std::call_once(init_once_flag, [] {
-        get_pooled_stack_allocator();
         photon::set_thread_stack_allocator({&pooled_stack_alloc, nullptr},
                                            {&pooled_stack_dealloc, nullptr});
     });


photon.cpp
```diff
@@ -86,6 +86,8 @@ static int init_event_engine(uint64_t engine, uint64_t flags, const PhotonOption
 int __photon_init(uint64_t event_engine, uint64_t io_engine, const PhotonOptions& options) {
     if (options.use_pooled_stack_allocator) {
         use_pooled_stack_allocator();
+    } else {
+        use_default_stack_allocator();
```
Coldwings, line 90: each vcpu will call photon_init, and use_pooled_stack_allocator is a global config.
This change may leads to user set-up pooled-stack-allocator and then unset. (by initialize another vcpu without setting use_pooled_stack_allocator. That is highly risky.

thread/stack-allocator.cpp
```diff
@@ -169,8 +173,13 @@ size_t pooled_stack_trim_threshold(size_t x) {
     return get_pooled_stack_allocator().threshold(x);
 }
 
-size_t pooled_stack_trim_current_vcpu(size_t keep_size);
-size_t pooled_stack_trim_threshold(size_t x);
+void use_pooled_stack_allocator() {
+    get_pooled_stack_allocator();
+    std::call_once(init_once_flag, [] {
```
Coldwings, line 178: Since it is a one-way setter and always put same value, is it really need to use a `std::call_once`?

thread/thread.cpp
```diff
@@ -2095,6 +2093,10 @@ R"(
     }
 
     int vcpu_init(uint64_t flags) {
+        if (unlikely(!photon_thread_alloc || !photon_thread_dealloc)) {
```
Coldwings, line 2096: The only problem is here:
once vcpu_init called in a global constructor, it might be called before `photon_thread_alloc` construct or library load.

that may cause delegate `photon_thread_alloc` become an empty delegate and unable to be called, result in allocation failure and segment fault.

Coldwings (commented):
`use_pooled_stack_allocator`是需要严格要求首次photon init是配置的（或者在vcpu_init之前进行配置），并且不允许改回。因此这个配置的处理才是单向的（仅能开启，不能通过设置为false进行关闭）。

对于明确知道栈分配器配置的用户场景，使用单独的`photon::set_thread_stack_allocator`可以自由控制更改；对于非此类用户，应当避免在使用默认option去初始化主线程外的vcpu时，由于使用默认配置而改回，这样会造成pool泄漏、释放不正确等问题。

PR #1040, opened 2025-11-13
fix resume_threads()

lihuiba:


thread/thread.cpp
```diff
@@ -1209,7 +1209,7 @@ R"(
                 AtomicRunQ().insert_list_before(q);
             }
         }
-        if (sleepq.empty() || !if_update_now()) {
+        if (sleepq.empty() && !if_update_now()) {
             return count;
         }
         do {
```
ivanallen, line 1216: Maybe `sleepq.empty()` is true here, then `sleepq.front()` will crash.

lihuiba: I should move ```if_update_now()``` out of the ```if```.

ivanallen (approved):

beef9999 (approved):

PR #1045, opened 2025-11-18
Fix multiple heap buffer overflow vulnerabilities in URL parsing

ShadowRider09:
- Fix heap-buffer-overflow in URL::from_string() by using string_view
     - Fix strtol overflow by manual port number parsing
     - Fix rstring_view boundary check failures
     - Fix null pointer dereference in to_int64_check()
     - Fix stack buffer overflow in iov_iterator
     and the fixed_crashes are zipped inattach files.
[fixed_crash.zip](https://github.com/user-attachments/files/23598029/fixed_crash.zip)

### DIFF
--- a/thread/workerpool.cpp
+++ b/thread/workerpool.cpp
@@ -62,6 +62,7 @@ public:
         } else {
             while (vcpus.size()) std::this_thread::yield();
         }
+        worker_lock.lock();
     }
 
     void enqueue(Delegate<void> call, AutoContext = {}) {


common/iovector.cpp
```diff
@@ -259,7 +259,12 @@ class iov_iterator {
         assert(_iovcnt);
         assert(n <= _v.iov_len);
         if (n < _v.iov_len) { _v += n; }
-        else { _v = *++_iov; _iovcnt--; }
```
lihuiba, line 262: This is an internal helper class inside a .cpp file. Its usage ensures the validness of ```*++_iov```. If you are worried about this, add an ```assert``` instead.

net/http/url.cpp
```diff
@@ -64,15 +80,22 @@ void URL::from_string(std::string_view url) {
...
-        auto port = strtol(url.begin() + 1, &endp, 10);
-        auto port_len= endp - url.begin();
-        if (endp != url.begin() + 1) {
-            m_port = port;
-            if (need_optional_port(*this)) m_host_port = rstring_view16(m_host.offset(), m_host.length() + port_len);
+        url.remove_prefix(1); // 跳过 ':'
+        // 手动解析端口号，避免使用 strtol 导致越界读取
+        uint16_t port_val = 0;
+        size_t port_len = 0;
+        while (port_len < url.size() && url[port_len] >= '0' && url[port_len] <= '9') {
```
lihuiba: parse the number with estring_view::to_uint64_check()

net/http/url.cpp
```diff
@@ -92,7 +115,8 @@ void URL::from_string(std::string_view url) {
     m_path = rstring_view16(pos, p);
     m_target = rstring_view16(pos, url.size());
     pos += p+1;
-    m_query = rstring_view16(pos, url.size() - (p + 1));
+    auto query_len = (p + 1 <= url.size()) ? (url.size() - (p + 1)) : 0;
```
lihuiba: use ```photon::sat_sub()``` to avoid overflow.

net/http/url.h
```diff
@@ -39,7 +39,7 @@ inline int what_protocol(estring_view url) {
 
 class URL {
 protected:
-    const char* m_url = nullptr;
+    std::string_view m_url;
```
lihuiba: why string_view is necessary for m_url?

net/http/url.cpp
```diff
@@ -23,13 +23,29 @@ namespace net {
 namespace http {
 
 void URL::fix_target() {
```
lihuiba, line 25: This is actually a private helper function only used in ```URL::from_string()```. I would suggest move the function body into ```URL::from_string()```.

net/http/url.h
ShadowRider09: Looking at the code, you're absolutely right. The std::string_view m_url is NOT necessary. The original const char* m_url was fine because:
     - rstring_view::operator|(const char*) calls to_abs(s) which simply does s + _offset with
   the stored _length - no strlen() is involved
     - All the rstring_view16 members (m_host, m_path, etc.) already store both offset and length
     - The actual heap-buffer-overflow was likely caused by something else, not by the type of m_url

net/http/url.h
```diff
@@ -57,15 +57,14 @@ class URL {
     }
 
     std::string to_string() {
-        return m_url;
+        return std::string(m_url);
```
lihuiba: simply ```return m_url;``` is fine.

net/http/url.cpp
```diff
@@ -61,38 +50,72 @@ void URL::from_string(std::string_view url) {
...
+        url.remove_prefix(1); // Skip ':'
+        // Parse port number, find the end of digit sequence
+        size_t port_len = 0;
+        while (port_len < url.size() && url[port_len] >= '0' && url[port_len] <= '9') {
+            port_len++;
+        }
+        if (port_len > 0) {
+            uint64_t port_val = 0;
+            estring_view port_str(url.data(), port_len);
+            if (port_str.to_uint64_check(&port_val)) {
```
lihuiba: check for port_val > 65535

net/http/url.cpp
```diff
@@ -61,38 +50,72 @@ void URL::from_string(std::string_view url) {
...
-    }
-    p = url.find_first_of("?");
-    if (p == url.npos) {
-        m_path = rstring_view16(pos, url.size());
-        m_query = rstring_view16(0, 0);
-        m_target = m_path;
-        return;
+    
+    // Fix target if needed (was previously a separate fix_target() function)
+    estring_view t = m_url | m_target;
```
lihuiba: you can still use ```DEFER({ ... });``` to execute this code snippet on *all* returns. If that's ugly, define a lambda, and invoke it with DEFER.

net/http/url.cpp
```diff
@@ -17,29 +17,18 @@ limitations under the License.
 #include "url.h"
 #include <photon/common/alog.h>
 #include <photon/common/alog-stdstring.h>
+#include <photon/common/utility.h>
```
lihuiba: Is this necessary?

ShadowRider09: Yes, it's necessary. The file now uses:
     - DEFER macro (line 35) - defined in utility.h
     - sat_sub function (lines 102, 117) - also defined in utility.h

   Both are required for the code to compile.

common/estring.h
```diff
@@ -400,7 +400,11 @@ class rstring_view  // relative string_view, that stores values relative
     }
     estring_view operator | (const std::string_view& s) const
     {
-        assert(s.length() >= _offset + _length);
+        // Ensure offset and length are within bounds of the string_view
+        if (s.length() < _offset + _length) {
+            // Return empty view if out of bounds
+            return estring_view("", (size_t)0);
```
lihuiba: simply return {};

net/http/url.cpp
```diff
@@ -61,38 +63,61 @@ void URL::from_string(std::string_view url) {
...
-    if (url.front() == ':') {
-        char* endp;
-        auto port = strtol(url.begin() + 1, &endp, 10);
-        auto port_len= endp - url.begin();
-        if (endp != url.begin() + 1) {
-            m_port = port;
-            if (need_optional_port(*this)) m_host_port = rstring_view16(m_host.offset(), m_host.length() + port_len);
+    } else if (url.front() == ':') {
+        url.remove_prefix(1); // Skip ':'
+        // Parse port number, find the end of digit sequence
```
lihuiba: The logic of finding the digits are already included in ```to_uint64_check()```. You may not need to do it yourself?

net/http/url.cpp
```diff
@@ -61,38 +63,61 @@ void URL::from_string(std::string_view url) {
...
+        }
+        pos += port_len + 1; // +1 for ':'
+        url.remove_prefix(port_len);
+        
+        if (url.empty()) {
+            m_path = rstring_view16(pos, 0);
+            m_query = rstring_view16(0, 0);
+            m_target = m_path;
+        } else {
+            p = url.find_first_of("?");
```
lihuiba: do NOT repeat similar logics, such as ```find_first_of("?")``` and its related processing.

net/http/url.cpp
```diff
@@ -61,38 +63,61 @@ void URL::from_string(std::string_view url) {
...
+        while (port_len < url.size() && url[port_len] >= '0' && url[port_len] <= '9') {
+            port_len++;
+        }
+        if (port_len > 0) {
+            uint64_t port_val = 0;
+            estring_view port_str(url.data(), port_len);
+            if (port_str.to_uint64_check(&port_val) && port_val <= 65535) {
+                m_port = static_cast<uint16_t>(port_val);
+            } else {
+                m_port = m_secure ? 443 : 80;
```
lihuiba: consider moving the assignment ```m_port = m_secure ? 443 : 80;``` upward, immediately after ```m_secure``` is assigned, in order to avoid code duplication.

net/http/url.cpp
```diff
@@ -61,38 +63,61 @@ void URL::from_string(std::string_view url) {
...
+            }
+            if (need_optional_port(*this)) m_host_port = rstring_view16(m_host.offset(), m_host.length() + port_len + 1);
+        } else {
+            m_port = m_secure ? 443 : 80;
+        }
+        pos += port_len + 1; // +1 for ':'
+        url.remove_prefix(port_len);
+        
+        if (url.empty()) {
+            m_path = rstring_view16(pos, 0);
```
lihuiba: goto the then part of prior ```if (url.empty()) { ... }```, in order to avoid code duplication.

net/http/url.cpp
```diff
@@ -56,43 +59,44 @@ void URL::from_string(std::string_view url) {
...
         m_target = m_path;
-        return;
-    }
-    p = url.find_first_of("?");
-    if (p == url.npos) {
-        m_path = rstring_view16(pos, url.size());
-        m_query = rstring_view16(0, 0);
-        m_target = m_path;
-        return;
+    } else {
```
lihuiba: The original form of **linear** code structure (with multiple early ```return```s) is more preferable than  nesting ```else```s and ```{...}```s, as it is easier to read and understand. 

If there were vulnerabilities, please try to fix them without changing code structure, unless it's not proper.

net/http/url.cpp
```diff
@@ -56,43 +59,44 @@ void URL::from_string(std::string_view url) {
...
-        auto port = strtol(url.begin() + 1, &endp, 10);
-        auto port_len= endp - url.begin();
-        if (endp != url.begin() + 1) {
-            m_port = port;
-            if (need_optional_port(*this)) m_host_port = rstring_view16(m_host.offset(), m_host.length() + port_len);
+    if (!url.empty() && url.front() == ':') {
+        url.remove_prefix(1); // Skip ':'
+        pos += 1; // +1 for ':'
+        // Find end of port digits
+        auto port_end = url.find_first_not_of("0123456789");
```
lihuiba: You may not need to find the first non-digit, because ```estring_view::to_uint64_check()``` already does that. It stops at the first non-digit. 

So, just call it directly.

lihuiba:
Let me address the issue directly and push to your branch.

net/http/url.cpp
lihuiba: It turns out that ```to_uint64_check``` doesn't return the number of valid digits, so it's not possible to avoid ```find_first_not_of()``` at present. 

Leave it as is for now. I'll address this issue in another PR.

lihuiba:
@ShadowRider09 As you were not responding, we merged the PR #1057 in order to continue improving related code.

Thanks again for your contribution!

lihuiba:
#1058 is the follow on PR

PR #1062, opened 2025-12-03
provide more information about threading

lihuiba:
provide more information about threading with a function ```get_info()```

### DIFF
--- a/common/intrusive_list.h
+++ b/common/intrusive_list.h
@@ -452,5 +452,10 @@ public:
         node = node->prev();
         return x;
     }
+    size_t count_by_loop() const {
+        size_t n = 0;
+        for (auto x: (intrusive_list&)*this) (void)x, ++n;
+        return n;
+    }
 };
 
--- a/thread/thread.cpp
+++ b/thread/thread.cpp
@@ -340,12 +340,10 @@ namespace photon
 
     __attribute__((used)) static void _asan_start() { ASAN_START(); }
 
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Winvalid-offsetof"
+    #pragma GCC diagnostic ignored "-Winvalid-offsetof"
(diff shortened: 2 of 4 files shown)


common/intrusive_list.h
```diff
@@ -452,5 +452,10 @@ class intrusive_list
         node = node->prev();
         return x;
     }
+    size_t count_by_loop() const {
+        size_t n = 0;
+        for (auto x: (intrusive_list&)*this) (void)x, ++n;
```
ivanallen, line 457: Is this kind of traversal thread-safe? Especially when standbyq may involve cross-thread insertions and deletions.

lihuiba: Yes, you are right. I have added locking around invocation.

thread/thread.cpp
```diff
@@ -2145,8 +2145,30 @@ R"(
...
+                return _n_vcpu.load(std::memory_order_relaxed);
+            case INFO_THREAD_NUM:
+                return vcpu->nthreads;
+            case INFO_SLEEPING_THREAD_NUM:
+                return vcpu->sleepq.q.size();
+            case INFO_STANDBY_THREAD_NUM: {
+                SCOPED_LOCK(vcpu->standbyq.lock);
+                return vcpu->standbyq.count_by_loop(); }
+            case INFO_RUNNABLE_THREAD_NUM: {
+                int64_t n = vcpu->nthreads - vcpu->sleepq.q.size();
```
Coldwings, line 2165: Since load `nthreads` and `sleepq.q.size` is not atomic, get `INFO_RUNNABLE_THREAD_NUM` across vcpu may also get unreliable result ( and also may break the assert below)

lihuiba: There seems no efficient solution for this case. What is your suggestion?

lihuiba: Actually, ```nthreads``` and ```sleepq``` are located at a same cache line, so they are probably consistent.

PR #1088, opened 2025-12-19
ALPN Client/Server options

Coldwings:
Add ALPN extension for TLS Context and TLS Stream. add basic tests.


net/security-context/tls-stream.h
```diff
@@ -108,5 +114,17 @@ ISocketClient* new_tls_client(TLSContext* ctx, ISocketClient* base,
 
 void tls_stream_set_hostname(ISocketStream* stream, const char* hostname);
 
+estring_view tls_stream_get_alpn_selected(ISocketStream* stream);
+
+class TLSStream : public ISocketStream {
```
lihuiba: not used?

PR #1090, opened 2025-12-23
Fix on domain resolver

Coldwings:
1. FIX: The old code once resolver failed to resolve a domain, the result will be kept and never try to resolve again. Fix it by drop empty list when resolve failed.
2. Using two different clock to calculate timeout and sharing objects should always possible leads to race. Using shared_ptr to ResolveCtx, copy and keep shared data to keep it safe.

### DIFF
--- a/net/utils.cpp
+++ b/net/utils.cpp
@@ -275,44 +275,31 @@ protected:
     };
     IPAddr do_resolve(std::string_view host, Delegate<bool, IPAddr> filter) {
         auto ctr = [&]() -> IPAddrList* {
-            auto addrs = new IPAddrList();
-            std::shared_ptr<ResolveCtx> ctx = std::make_shared<ResolveCtx>();
-            ctx->addrs = addrs;
+            std::unique_ptr<IPAddrList> addrs(new IPAddrList());
+            auto ctx = std::make_shared<ResolveCtx>();
+            ctx->addrs = addrs.get();
             ctx->host = std::string(host);
             ctx->filter = filter;
             std::thread([ctx]() {
-                IPAddrList ret;
                 auto cb = [&](IPAddr addr) -> int {
                     SCOPED_LOCK(ctx->lock);
-                    if (ctx->filter && !ctx->filter.fire(addr))
-                        return 0;
-                    ret.push_back(new IPAddrNode(addr));
+                    if (ctx->filter && !ctx->filter.fire(addr)) return 0;
+                    if (ctx->addrs) {
+                        ctx->addrs->push_back(new IPAddrNode(addr));
+                    }
                     return 0;
                 };
                 _gethostbyname(ctx->host, cb);
-                {
-                    SCOPED_LOCK(ctx->lock);
-                    if (ctx->addrs) {
-                        ctx->addrs->push_back(std::move(ret));
-                        ctx->sem.signal(1);
-                    } else {
-                        LOG_ERROR("resolve timeout");
-                        while(!ret.empty())
-                            delete ret.pop_front();
-                    }
-                }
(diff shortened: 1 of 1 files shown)


net/utils.cpp
```diff
@@ -266,36 +266,59 @@ class DefaultResolver : public Resolver {
...
+        IPAddrList *addrs;
+        photon::semaphore sem;
+    };
     IPAddr do_resolve(std::string_view host, Delegate<bool, IPAddr> filter) {
         auto ctr = [&]() -> IPAddrList* {
             auto addrs = new IPAddrList();
-            photon::semaphore sem;
-            std::thread([&]() {
-                auto now = std::chrono::steady_clock::now();
+            std::shared_ptr<ResolveCtx> ctx = std::make_shared<ResolveCtx>();
```
lihuiba: auto ctx = ...

net/utils.cpp
```diff
@@ -266,36 +266,59 @@ class DefaultResolver : public Resolver {
...
-                                        std::chrono::steady_clock::now() - now).count();
-                if ((uint64_t)time_elapsed <= resolve_timeout_) {
-                    addrs->push_back(std::move(ret));
-                    sem.signal(1);
-                } else {
-                    LOG_ERROR("resolve timeout");
-                    while(!ret.empty())
-                        delete ret.pop_front();
+                _gethostbyname(ctx->host, cb);
+                {
```
lihuiba: why scope ```{ ... } ```?

net/utils.cpp
```diff
@@ -266,36 +266,59 @@ class DefaultResolver : public Resolver {
...
                 };
-                _gethostbyname(host, cb);
-                auto time_elapsed = std::chrono::duration_cast<std::chrono::microseconds>(
-                                        std::chrono::steady_clock::now() - now).count();
-                if ((uint64_t)time_elapsed <= resolve_timeout_) {
-                    addrs->push_back(std::move(ret));
-                    sem.signal(1);
-                } else {
-                    LOG_ERROR("resolve timeout");
-                    while(!ret.empty())
```
lihuiba, line 290: ~IPAddrList() will ```delete_all()```

net/utils.cpp
```diff
@@ -266,36 +266,59 @@ class DefaultResolver : public Resolver {
...
+            ctx->host = std::string(host);
+            ctx->filter = filter;
+            std::thread([ctx]() {
                 IPAddrList ret;
                 auto cb = [&](IPAddr addr) -> int {
-                    if (filter && !filter.fire(addr))
+                    SCOPED_LOCK(ctx->lock);
+                    if (ctx->filter && !ctx->filter.fire(addr))
                         return 0;
                     ret.push_back(new IPAddrNode(addr));
```
lihuiba: you can insert into ctx->addrs directly, and rely on the return value of ```ctx->sem.wait()``` to decide weather use it or not.

net/utils.cpp
```diff
@@ -266,36 +266,59 @@ class DefaultResolver : public Resolver {
...
-                    while(!ret.empty())
-                        delete ret.pop_front();
+                _gethostbyname(ctx->host, cb);
+                {
+                    SCOPED_LOCK(ctx->lock);
+                    if (ctx->addrs) {
+                        ctx->addrs->push_back(std::move(ret));
+                        ctx->sem.signal(1);
+                    } else {
+                        LOG_ERROR("resolve timeout");
```
lihuiba: LOG_ERROR_RETURN can be placed right after ```ctx->sem.wait()``` times out.

lihuiba (approved):

PR #1182, opened 2026-04-14
feat(cache): add idle LRU for full file cache to reduce memory usage

xiaoyang-hhh:
* Add idle LRU for closed cache file to reduce memory usage under large-file-count scenarios:
    * When default LRU exceeds demoteThreshold (default 1M), demote LRU tail entries (if not opened) to idle LRU
    * When file in idle LRU is re-accessed, promote it back to default LRU
    * Memory reduce points for idle LRU:
        * Replace LruEntry (24 bytes) with only lruIter (4 bytes)
        * Remove unused unique_ptr (8 bytes)
    * With 10 million files, memory usage drops about 300 MB
* Currently enabled only in non-quota mode.

### DIFF
--- a/fs/cache/full_file_cache/cache_pool.h
+++ b/fs/cache/full_file_cache/cache_pool.h
@@ -42,6 +42,8 @@ public:
     static const uint64_t kDiskBlockSize = 512; // stat(2)
     static const uint64_t kDeleteDelayInUs = 1000;
     static const uint32_t kWaterMarkRatio = 90;
+    // max inuse-lru entries before demoting tail to idle-lru (default: 100w)
+    static const uint32_t kDemoteThreshold = 1'000'000;
 
     void Init();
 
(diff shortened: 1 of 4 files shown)


fs/cache/full_file_cache/cache_pool.h
```diff
@@ -109,6 +111,22 @@ class FileCachePool : public photon::fs::ICachePool {
     LRUContainer lru_;
     // filename -> lruEntry
     FileNameMap fileIndex_;
+
+    uint32_t demoteThreshold_ = kDemoteThreshold;
+
+    // idleFileIndex_ stores filename -> lruContainer's iterator
+    // idleLru_ stores iterators into idleFileIndex_; std::map nodes are stable.
+    typedef map_string_key<uint32_t> IdleFileNameMap;
```
Moonpuck, line 119: Idle -> Inactive?

EricHuangqx (approved):

PR #1186, opened 2026-04-20
alog: make LogBuilder destructor noinline to avoid 4KB stack leak

hulexwang:
### Problem

When `LogBuilder::~LogBuilder()` is marked `__INLINE__`, the compiler inlines the destructor into every caller. Since the destructor invokes the `noinline, cold` lambda which returns a `STFMTLogBuffer` (containing a 4KB `char buf[LOG_BUFFER_SIZE]`), the return value must be allocated in the **caller's** stack frame (the inlined destructor has no frame of its own).

This means any function that uses `LOG_WARN`, `LOG_ERROR`, etc. unconditionally allocates **4KB+** on the stack — even when the log is never written because the level check fails.

**Example**: a simple function with one `LOG_DEBUG`:

```cpp
int hot_function(int x) {
    LOG_DEBUG("x=`", x);
    return x * 2 + 1;
}
```

With `__INLINE__` destructor (before):
```asm
hot_function(int):
    sub    rsp, 0x1028          ; 4KB+ stack allocation, unconditional
    ...
```

### Fix

Change `~LogBuilder()` from `__INLINE__` to `__attribute__((noinline))`:

```cpp
-    ~LogBuilder() __INLINE__ {
+    __attribute__((noinline)) ~LogBuilder() {
```

### Verification via objdump

Test code compiled with `c++ -std=c++17 -O2`, inspected via `objdump -d -C --no-show-raw-insn`.

**Note on stack probes**: On Ubuntu/Debian, GCC enables `-fstack-clash-protection` by default (via `gcc -dumpspecs`, in `*distro_defaults`). This inserts `orq $0x0,(%rsp)` stack probes that touch memory on every 4KB allocation. On macOS/Clang or other platforms without this default, the `orq` is absent — but the 4KB RSP shift and spatial locality issue remain.

#### Before (`__INLINE__` destructor)

With `-fstack-clash-protection` (Ubuntu/Debian default):
```asm
hot_function(int):                        ; even LOG_DEBUG causes 4KB stack
    sub    $0x1000,%rsp                   ; <-- 4096 bytes unconditional!
    orq    $0x0,(%rsp)                    ; stack probe — touches memory!
    sub    $0x38,%rsp                     ; + 56 bytes = 4152 total
    ...
    test   %eax,%eax                      ; level check (inlined destructor)
    je     .skip
    ...
.skip:
    add    $0x1038,%rsp
    ret
```

Without `-fstack-clash-protection` (macOS/Clang):
```asm
hot_function(int):
    pushq  %rbp
    movq   %rsp, %rbp
    subq   $0x1030, %rsp                  ; 4KB+ allocated, no probe
    movl   default_logger+8(%rip), %eax   ; load log level
    movl   %edi, -0x1024(%rbp)            ; x stored 4KB below frame base!
    testl  %eax, %eax
    je     .cold
    movl   -0x1024(%rbp), %eax            ; load x back from 4KB offset
    leave
    leal   0x1(%rax,%rax), %eax           ; x*2+1
    retq
```

In both cases, the caller's stack frame is 4KB+. Note that even without the stack probe, the parameter `x` is stored at `-0x1024(%rbp)` — 4KB away from the caller's frame, degrading spatial locality.

Another example — a function where only the error branch logs, but the fast path still pays 4KB:

```cpp
int hot_function_error(int x) {
    if (x < 0) {
        LOG_ERROR("negative x=`", x);
        return -1;
    }
    return x + 1;  // fast path: no logging
}
```

```asm
hot_function_error(int):
    subq   $0x1030, %rsp           ; 4KB allocated BEFORE any branch
    test   %edi, %edi
    js     .do_log                  ; x < 0: error path with LOG_ERROR
    add    $0x1, %eax              ; x >= 0: fast path — never logs, but 4KB already on stack
    add    $0x1038, %rsp
    ret
```

The 4KB stack frame is allocated at function entry, unconditionally — even though the `x >= 0` fast path never reaches the LOG_ERROR.

#### After (`noinline` destructor)

**`.text` section (hot path)** — the caller is compact:

```asm
hot_function(int):
    sub    $0x48,%rsp                     ; <-- only 72 bytes!
    ; construct LogBuilder on stack (a few MOVs, no formatting)
    movl   $0x0,0x10(%rsp)               ; level = ALOG_DEBUG
    movb   $0x0,0x28(%rsp)               ; done = false
    call   ~LogBuilder()                  ; noinline: level check in separate frame
    lea    0x1(%rax,%rax,1),%eax          ; business logic: x*2+1
    add    $0x48,%rsp
    ret

hot_function_error(int):
    sub    $0x48,%rsp                     ; <-- only 72 bytes!
    test   %edi,%edi
    js     .cold_branch                   ; unlikely: x < 0
    add    $0x1,%eax                      ; fast path: x + 1, compact stack
    add    $0x48,%rsp
    ret
.cold_branch:
    ; construct LogBuilder, call ~LogBuilder
    mov    $0xffffffff,%eax
    jmp    .ret
```

**`LogBuilder::~LogBuilder()`** — level check, early return if not met:

```asm
LogBuilder::~LogBuilder():
    cmp    level, [logger->log_level]
    jbe    .Lreturn             ; fast path: log level not met, return
    ; fall through to cold clone...

LogBuilder::~LogBuilder() [clone .cold]:       ; in .text.unlikely
    call   {lambda}::operator()                ; 4KB buffer allocated here
    call   ILogOutput::write()
    ret
```

**`.text.unlikely` section (cold path)** — all formatting code is isolated:

```asm
{lambda}::operator()(ILogOutput*):
    sub    rsp, 0x1028          ; 4KB buffer allocated in cold path only
    ; Prologue formatting, string formatting, write
    ret
```

#### Stack frame comparison

| Function | Before (`__INLINE__`) | After (`noinline`) | Reduction |
|----------|----------------------|-------------------|-----------|
| `hot_function` | 0x1038 (4152 bytes) | 0x48 (72 bytes) | **98.3%** |
| `hot_function_error` | 0x1038 (4152 bytes) | 0x48 (72 bytes) | **98.3%** |

### Performance improvement

Below are the maximum latency statistics of a live production project. After optimization, the tail latency has been reduced from 30+ms to 2ms.

image

### Impact

- **Stack usage**: Reduces per-function stack consumption by **~4KB** for every function containing a LOG_* call (4152 → 72 bytes, **98.3% reduction**). In coroutine/fiber environments like Photon (default stack 64KB–8MB), each LOG_* site unconditionally reserves 4KB on the caller's stack. A 3-level call chain with 4 LOG_* sites wastes 16KB — 25% of a 64KB coroutine stack — for log buffers that are never used, increasing the risk of stack overflow in deep call chains.
- **Stack spatial locality**: The `sub rsp, 0x1020` shifts the stack pointer down by 4KB. As shown in the objdump above, the parameter `x` is stored at `-0x1024(%rbp)` — 4KB away from the caller's frame. All subsequent callee stack frames end up 4KB further from the caller's data. In a deep call chain, this accumulates: with N LOG_* sites, the actual working data is spread across N×4KB of extra address space, degrading cache line and TLB hit rates for stack accesses.
- **Stack probe on Linux**: On Ubuntu/Debian (and many production Linux environments), GCC enables `-fstack-clash-protection` by default (`gcc -dumpspecs` → `*distro_defaults`). This inserts `orq $0x0,(%rsp)` that touches memory on each 4KB allocation, adding real memory access overhead even when the log is never written.
- **Overhead**: One noinline function call (~1ns) per LOG_* site, negligible compared to the microsecond-scale cost of actual log I/O.

### DIFF
--- a/common/alog.h
+++ b/common/alog.h
@@ -442,7 +442,7 @@ struct LogBuilder {
         // so just make sure the other one will not output
         rhs.done = true;
     }
-    ~LogBuilder() __INLINE__ {
+    __attribute__((noinline)) ~LogBuilder() {
         if (!done && level >= logger->log_level) {
             builder(logger->log_output);
             done = true;


lihuiba:
What are the 2 ```orq``` s doing? I don't have those in my machine. That instruction will access stack memory, and it incurs a much bigger overhead compared to the payload ```x * 2 + 1```.
image

lihuiba:
What do you mean by "the x>=0 fast path"?
image
```cpp
int hot_function(int x) {
    LOG_DEBUG("x=`", x);
    return x * 2 + 1;
}
```

lihuiba:
Take test_log2() in test_alog.cpp for example:
```cpp
__attribute__((noinline))
void test_log2(int x)
{
    LOG_DEBUG(ALogStringL("asdf:"), 1, ", jkl:", 2, ", and: ", 3,4,5);
}
```
It compiles to 
```asm
test-alog[0x4108c0] <+0>:  pushq  %rbp
test-alog[0x4108c1] <+1>:  movq   %rsp, %rbp
test-alog[0x4108c4] <+4>:  subq   $0x1020, %rsp             ; imm = 0x1020
test-alog[0x4108cb] <+11>: movl   0x8e2d7(%rip), %eax       ; default_logger + 8
test-alog[0x4108d1] <+17>: testl  %eax, %eax
test-alog[0x4108d3] <+19>: je     0x40b8d5                  ; test_log2(int) (.cold)
test-alog[0x4108d9] <+25>: leave
test-alog[0x4108da] <+26>: retq
```
The total stack space is subtracted with one single instruction, and there it doesn't access stack memory.

lihuiba:
and the ```hot_function()``` compiles to following instructions in my machine:
```asm
(lldb) disassemble -n hot_function
test-alog`hot_function:
test-alog[0x4109c0] <+0>:  pushq  %rbp
test-alog[0x4109c1] <+1>:  movq   %rsp, %rbp
test-alog[0x4109c4] <+4>:  subq   $0x1030, %rsp             ; imm = 0x1030
test-alog[0x4109cb] <+11>: movl   0x8e1d7(%rip), %eax       ; default_logger + 8
test-alog[0x4109d1] <+17>: movl   %edi, -0x1024(%rbp)
test-alog[0x4109d7] <+23>: testl  %eax, %eax
test-alog[0x4109d9] <+25>: je     0x40b9b5                  ; hot_function(int) (.cold)
test-alog[0x4109df] <+31>: movl   -0x1024(%rbp), %eax
test-alog[0x4109e5] <+37>: leave
test-alog[0x4109e6] <+38>: leal   0x1(%rax,%rax), %eax
test-alog[0x4109ea] <+42>: retq
```

hulexwang:
Thanks for the detailed review @lihuiba!

## Regarding `orq` (stack probe)

The `orq $0x0,(%rsp)` in my objdump output comes from `-fstack-clash-protection`. On Ubuntu/Debian with GCC, this flag is **enabled by default** via the compiler specs (`%{!fno-stack-clash-protection:-fstack-clash-protection}`). On macOS/Clang or other distros it may not be default, which explains the difference in our outputs.

With `-fstack-clash-protection` enabled (common in Linux production environments), the 4KB stack allocation does trigger a stack probe that touches memory. Without it, `sub rsp, 0x1020` is just a register adjustment — but the spatial locality issue still applies (see below).

## Regarding "x>=0 fast path"

That referred to `hot_function_error()` in my test code (not `hot_function()`), which has an early return for `x >= 0`. The point was that even the non-logging branch still pays the 4KB stack cost. I'll clarify this in the description.

## The spatial locality argument

Looking at your `hot_function()` disassembly:

```asm
test-alog[0x4109c4] <+4>:  subq   $0x1030, %rsp
test-alog[0x4109d7] <+23>: movl   %edi, -0x1024(%rbp)   ; x stored 4KB below frame base
...
test-alog[0x4109df] <+31>: movl   -0x1024(%rbp), %eax   ; x loaded back from 4KB offset
test-alog[0x4109e6] <+38>: leal   0x1(%rax,%rax), %eax  ; x*2+1
```

The parameter `x` is stored at `-0x1024(%rbp)` — 4KB below the frame base. This means even though the 4KB log buffer itself is never touched, the function's actual working data (parameter `x`) is placed 4KB away from the caller's stack frame.

In a deep call chain with multiple LOG_* sites, this accumulates. Consider a typical 3-level chain:

```
caller()          ; RSP = base
  outer_handler() ; RSP = base - 4KB  (LOG_DEBUG here)
    mid_lookup()  ; RSP = base - 8KB  (LOG_DEBUG here)
      inner_op()  ; RSP = base - 12KB (LOG_DEBUG here)
```

The actual working data — function arguments, local variables — is spread across 12KB+ of address space, instead of fitting within a few cache lines (~300 bytes without the log gaps). Each function's parameters end up on a different page, reducing spatial locality for stack accesses.

With the noinline fix, the stack frames are compact (~72 bytes each), so the entire 3-level call chain's working data stays within 1-2 cache lines.

## Coroutine stack overflow

In Photon coroutine environments with limited stacks (e.g. 64KB), each LOG_* site unconditionally reserves ~4KB of stack. A 3-level call chain with 4 LOG_* calls consumes 16KB — 25% of a 64KB coroutine stack — for log buffers that are never used when the log level check fails.

## Overhead

The noinline destructor adds ~1ns per LOG_* site (one function call for the level check). This is negligible compared to the microsecond-scale cost of actual log I/O when logging does occur.

hulexwang:
PR description updated.

lihuiba:
The noinline destructor of ~LogBuilder makes hot_function() compiled to:
```asm
hot_function(int):
    sub    $0x48,%rsp                     ; <-- only 72 bytes!
    ; construct LogBuilder on stack (a few MOVs, no formatting)
    movl   $0x0,0x10(%rsp)               ; level = ALOG_DEBUG
    movb   $0x0,0x28(%rsp)               ; done = false
    call   ~LogBuilder()                  ; noinline: level check in separate frame
    lea    0x1(%rax,%rax,1),%eax          ; business logic: x*2+1
    add    $0x48,%rsp
    ret
```

The check for whether doing the log output is buried in ~LogBuilder(), and it is most likely to be false, so it's best to get inlined to avoid an invocation in most cases. 

I'll try to fix the buffer issue with another approach.

hulexwang:
Got it. The current fix is just for minimal code changes, not the optimal one. I will close this PR for later

lihuiba:
@hulexwang Please review #1189

PR #1197, opened 2026-04-27
[feat.] remove O_RDONLY flag in httpfs_v2

BigVan:
O_RDONLY flag in httpfs_v2 shouldn't be set when people try to
upload data chunks to remote HTTP Server (eg. OSS)


Coldwings (approved):
LGTM.

Details of prevent open with RDONLY then calling write should add into repo later.

PR #1201, opened 2026-04-27
feat(http): support per-operation proxy URL for HTTP client

pakkah:
## Summary

Add `Operation::set_proxy()` / `clear_proxy()` to allow individual operations
to specify a proxy URL that takes precedence over the client-level proxy.

This enables use cases where a single `Client` sends requests through different
proxies.

## Changes

- **`client.h`**: Add `proxy_url` fields and `set_proxy()` / `get_proxy()` /
  `clear_proxy()` methods on `Operation`. `set_proxy()` automatically enables
  proxy and rebuilds the request line from relative path to absolute URI format
  (required by HTTP proxy protocol) when proxy was not
  previously enabled.
- **`client.cpp`**: Update `do_roundtrip()` routing to check operation-level
  proxy (`enable_proxy && !proxy_url.empty()`) before client-level proxy.

## Design Notes

- `enable_proxy` controls **request line format** (relative vs absolute URI).
- `proxy_url` controls **routing decision** (which proxy to connect).
- These are decoupled: changing the proxy target does not require rebuilding the request line.
- `clear_proxy()` keeps `enable_proxy=true` so fallback to client-level proxy
  uses the correct absolute URI format.

### DIFF
--- a/net/http/client.h
+++ b/net/http/client.h
@@ -83,6 +83,24 @@ public:
         void set_enable_proxy(bool enable) {
             enable_proxy = enable;
         }
+        // Set per-operation proxy URL, takes precedence over client-level proxy.
+        // Automatically enables proxy and rebuilds request line to absolute URI format.
+        void set_proxy(std::string_view proxy) {
+            proxy_url.from_string(proxy);
+            if (!enable_proxy) {
+                // redirect() handles scheme+host completion and absolute URI rebuild
+                req.redirect(req.verb(), req.target(), true);
+                enable_proxy = true;
+            }
+        }
+        const StoredURL* get_proxy() {
+            return &proxy_url;
+        }
+        // Clear per-operation proxy, fall back to client-level proxy.
+        // Does not change enable_proxy or request line format.
+        void clear_proxy() {
+            proxy_url.clear();
+        }
         int call() {
             if (!_client) return -1;
             return _client->call(this);
@@ -105,6 +123,8 @@ public:
 
     protected:
         Client* _client;
+        StoredURL proxy_url; // Per-operation proxy URL (takes precedence over
+                             // client-level proxy)
         const void *body_buffer = nullptr;
         size_t body_buffer_size = 0;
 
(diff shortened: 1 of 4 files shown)


net/http/client.h
```diff
@@ -62,6 +62,8 @@ class Client : public Object {
         bool enable_proxy = false;
         std::string_view uds_path;                // If set, Unix Domain Socket will be used instead of TCP.
                                                   // URL should still be the format of http://localhost/xxx
+        StoredURL proxy_url;                      // Per-operation proxy URL (takes precedence over client-level proxy)
```
lihuiba: put these in ```protected``` section, given that we have setters.

pakkah: done, thanks!

net/http/client.h
```diff
@@ -83,6 +83,24 @@ class Client : public Object {
...
+        // Automatically enables proxy and rebuilds request line to absolute URI format.
+        void set_proxy(std::string_view proxy) {
+            proxy_url.from_string(proxy);
+            if (!enable_proxy) {
+                // redirect() handles scheme+host completion and absolute URI rebuild
+                req.redirect(req.verb(), req.target(), true);
+                enable_proxy = true;
+            }
+        }
+        StoredURL* get_proxy() {
```
lihuiba: const URL*

net/http/url.h
```diff
@@ -106,6 +106,12 @@ class StoredURL : public URL {
         auto u = strndup(url.data(), url.size());
         URL::from_string({u, url.size()});
     }
+    void clear() {
+        if (m_url) {
```
lihuiba: no need ```if```, free() accepts nullptr

net/http/url.h
lihuiba: no need if, free() accepts nullptr

net/http/client.h
pakkah: done

net/http/url.h
pakkah: done

net/http/url.h
pakkah: done

lihuiba (approved):

PR #1206, opened 2026-05-06
fix(url): move `remove_prefix` inside scheme check to fix relative path parsing

pakkah:
In commit 9600530 (#1058), `url.remove_prefix(p + 3)` was moved outside the `if (p != url.npos)` block during refactoring. When `://` is not found (URLs that do not start with `http[s]://`), `p == npos` and `p + 3` overflows to `2`, causing `remove_prefix(2)` to strip the first two characters and corrupt subsequent parsing.
Move `remove_prefix(p + 3)` back inside the if block, consistent with the pre-#1058 behavior, so that relative paths are parsed correctly.

### DIFF
--- a/net/http/test/client_function_test.cpp
+++ b/net/http/test/client_function_test.cpp
@@ -712,6 +712,47 @@ TEST(url, path_fix) {
     EXPECT_EQ(u2.query(), "a=b");
 }
 
+TEST(url, no_scheme) {
+    // Relative paths
+    URL u0("/");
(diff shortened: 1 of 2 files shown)


lihuiba:
add some test to prevent similar bugs

pakkah:
> add some test to prevent similar bugs

done, added a test to cover URLs without http[s]:// prefix

lihuiba (approved):

PR #1230, opened 2026-05-21
feat(cache): probe fiemap support and track ranges as fallback

xiaoyang-hhh:
Some media filesystems (notably tmpfs) do not support fiemap, which the `full_file_cache` relies on to detect holes during refill. Probe fiemap once per pool via an `O_TMPFILE` on the media FS; when unsupported, each `FileCacheStore` tracks filled byte ranges in an in-memory RangeModule rebuilt on open via `SEEK_DATA/SEEK_HOLE`, and `queryRefillRange` consults it instead of calling fiemap.

Note on `SEEK_HOLE/SEEK_DATA` on tmpfs: tmpfs implements these by linearly scanning the page cache, which is significantly slower than fiemap on disk-backed filesystems and degrades further with file size and sparseness.

Also route pool-level evictions through cacheStore->evict() so `filledRanges_` stays consistent with the file, and add tests covering the fallback path.


fs/cache/full_file_cache/cache_pool.cpp
```diff
@@ -70,9 +71,38 @@ FileCachePool::~FileCachePool() {
 
 void FileCachePool::Init() {
   traverseDir("/");
+  probeFiemap();
   timer_ = new photon::Timer(periodInUs_, {this, FileCachePool::timerHandler}, true, 8UL * 1024 * 1024);
 }
 
+void FileCachePool::probeFiemap() {
+  auto probe = mediaFs_->open("/", O_TMPFILE | O_RDWR, 0600);
```
lihuiba: using a special rare file name is more universal than O_TMPFILE

fs/cache/full_file_cache/cache_pool.h
```diff
@@ -102,6 +108,9 @@ class FileCachePool : public photon::fs::ICachePool {
 
     bool isFull_;
 
+    std::atomic<bool> fiemapSupported_{true};
```
lihuiba: why atomic? it is written only once at init.

fs/cache/full_file_cache/range_module.h
```diff
@@ -0,0 +1,152 @@
...
+#include <utility>
+
+namespace photon {
+namespace fs {
+
+// A disjoint interval set that supports add/remove/query operations.
+// Internally stored as a sorted map: key = interval start, value = interval end.
+// All intervals are half-open [left, right). Overlapping/adjacent intervals are
+// automatically merged on insertion.
+class RangeModule {
```
lihuiba, line 30: The handling of intervals seems verbose. Can we share some code form ```RangeLock```?

fs/cache/full_file_cache/cache_pool.cpp
xiaoyang-hhh: Done

fs/cache/full_file_cache/cache_pool.h
xiaoyang-hhh: atomic removed

fs/cache/full_file_cache/range_module.h
xiaoyang-hhh: `RangeLock` can't handle overlapping ranges, so it can't express merge/split. But I simplify the implementation, and it may be a bit cleaner now.

lihuiba:
LGTM
@EricHuangqx please review the code and perform approve / merge if it's ok.

EricHuangqx (approved):

PR #1238, opened 2026-06-02
fix(ecosystem/rapidjson): add RawBool handler to support kParseBoolsAsStringFlag

yyaoling:


beef9999:
@lihuiba Please take a look at this PR

lihuiba:
@yyaoling @beef9999 what does this PR try to fix?

yyaoling:
> @yyaoling @beef9999 what does this PR try to fix?

Background
  Photon's JSON DOM (ecosystem/simple_dom.cpp) needs to store JSON true/false as raw string tokens in JNode (consistent with how RawNumber is handled). To enable this, an earlier patch that
  introduced the kParseBoolsAsStringFlag feature for RapidJSON modified ParseTrue/ParseFalse in reader.h so that they invoke handler.RawBool(...) via a ternary expression.

  There are two kinds of handler implementations here:

  - The only real custom implementation is Photon's JHandler (simple_dom.cpp:221), which overrides RawBool to emplace the token into the DOM as a JNode::BOOLEAN ? this is the sole place where the
  feature actually takes effect and gets called;
  - The base class BaseReaderHandler provides only a default-forwarding implementation (RawBool ? Default()), serving as a fallback for handlers that derive from it and override callbacks
  selectively, so they still compile even when they don't care about bool tokens.

  Because both branches of a ternary expression must compile during template instantiation (regardless of which branch is actually taken at runtime), every class used as a Handler is now required
  to provide a RawBool member function. However, the original patch only added the default-forwarding method to BaseReaderHandler in reader.h, and missed the actual handlers that are complete 
  handlers in their own right and do not forward through the base class: GenericDocument, Writer, PrettyWriter, and SchemaValidator.
  
  As a result, any downstream project that uses Document::Parse() to parse JSON hits a compilation failure:

  build/_deps/rapidjson-src/include/rapidjson/reader.h:734:25: error: 'class rapidjson::GenericDocument<rapidjson::UTF8<> >' has no member named 'RawBool'; did you mean 'Bool'?
    734 |                 handler.RawBool(begin, 4, copy) : handler.Bool(true);
        |                 ~~~~~~~~^~~~~~~
        |                 Bool
  build/_deps/rapidjson-src/include/rapidjson/reader.h:751:25: error: 'class rapidjson::GenericDocument<rapidjson::UTF8<> >' has no member named 'RawBool'; did you mean 'Bool'?
    751 |                 handler.RawBool(begin, 5, copy) : handler.Bool(false);
        |                 ~~~~~~~~^~~~~~~
        |                 Bool

  Fix

  Add the RawBool method to the Handler classes in the following files within ecosystem/patches/rapidjson.patch, uniformly treating the raw bool token as a string value: document.h, writer.h,
  prettywriter.h, and schema.h. Like the base class, these added methods carry generic fallback semantics (forwarding to String()), distinct from JHandler's implementation that actually consumes
  the bool token. After the fix, all handlers compile correctly whether or not the flag is enabled; the method is only invoked at runtime when kParseBoolsAsStringFlag is explicitly set, so the
  default parsing behavior is unchanged.


@lihuiba

lihuiba (approved):

PR #1250, opened 2026-06-03
Cap HTTP client retry backoff at 5 seconds

jfeng18:
## Summary

Fixes unbounded exponential backoff in HTTP client retry logic. The interval is now capped at 5 seconds to prevent excessive tail latency on transient failures.

## Changes

| File | Change |
|------|--------|
| `net/http/client.cpp:320` | Added `std::min(..., 5000UL)` cap to sleep_interval |

## Test plan

- [ ] Existing HTTP client tests pass
- [ ] Manual: verify retry intervals don't exceed 5s under repeated failures

### DIFF
--- a/net/http/client.cpp
+++ b/net/http/client.cpp
@@ -317,7 +317,7 @@ public:
             switch (ret) {
                 case ROUNDTRIP_NEED_RETRY:
                     photon::thread_usleep(sleep_interval * 1000UL);
-                    sleep_interval = (sleep_interval + 500) * 2;
+                    sleep_interval = std::min((sleep_interval + 500) * 2, 5000UL);
                     ++retry;
                     break;
                 case ROUNDTRIP_FAST_RETRY:
--- a/net/http/client.cpp
+++ b/net/http/client.cpp
@@ -316,8 +316,10 @@ public:
             if (ret == ROUNDTRIP_SUCCESS || ret == ROUNDTRIP_FAILED) break;
             switch (ret) {
                 case ROUNDTRIP_NEED_RETRY:
-                    photon::thread_usleep(sleep_interval * 1000UL);
-                    sleep_interval = std::min((sleep_interval + 500) * 2, 5000UL);
+                    // never sleep past the remaining deadline, otherwise a single
+                    // backoff can overshoot the operation timeout by a wide margin
+                    photon::thread_usleep(std::min(sleep_interval * 1000UL, tmo.timeout()));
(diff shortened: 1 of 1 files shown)


net/http/client.cpp
```diff
@@ -317,7 +317,7 @@ class ClientImpl : public Client {
             switch (ret) {
                 case ROUNDTRIP_NEED_RETRY:
                     photon::thread_usleep(sleep_interval * 1000UL);
-                    sleep_interval = (sleep_interval + 500) * 2;
+                    sleep_interval = std::min((sleep_interval + 500) * 2, 5000UL);
```
lihuiba: you need it to retry many times?

jfeng18: Good point. Dropped the arbitrary 5s cap and instead clamp each backoff to `std::min(sleep_interval*1000, tmo.timeout())`. This bounds total retry wall-clock to the operation's own deadline rather than a magic constant — the client never sleeps past the timeout it was given, so the answer to "retry many times?" is now self-bounding: regardless of how many retries or how the interval grows, the cumulative sleep can't exceed `tmo`. `Timeout::timeout()` uses saturating subtraction, so near the deadline the clamp goes to 0 and the existing `tmo.timeout()==0` guard returns `ETIMEDOUT`. The exponential growth is left uncapped but harmless since every sleep is clamped.

net/http/client.cpp
```diff
@@ -316,7 +316,9 @@ class ClientImpl : public Client {
             if (ret == ROUNDTRIP_SUCCESS || ret == ROUNDTRIP_FAILED) break;
             switch (ret) {
                 case ROUNDTRIP_NEED_RETRY:
-                    photon::thread_usleep(sleep_interval * 1000UL);
+                    // never sleep past the remaining deadline, otherwise a single
+                    // backoff can overshoot the operation timeout by a wide margin
+                    photon::thread_usleep(std::min(sleep_interval * 1000UL, tmo.timeout()));
```
lihuiba: limit it to remaining timeout is good, but  wait, why should we multiply sleep_interval by 1000 here? I believe we can define its unit as us at first.

jfeng18: Good catch. Changed sleep_interval to microseconds directly — removed the `* 1000` and adjusted the increment to `500'000UL`. Now both sleep_interval and tmo.timeout() are in the same unit, no conversion needed.

lihuiba (approved):

PR #1252, opened 2026-06-03
Fix crash in threads_create_join on partial failure

jfeng18:
## Summary

Track the actual number of threads created and only join those, preventing segfault when thread_create() fails mid-batch.

## Changes

| File | Change |
|------|--------|
| `thread/thread.cpp:1580-1589` | Added `created` counter; join loop bounded by `created` instead of `n` |

## Test plan

- [ ] Existing thread tests pass
- [ ] ctest full suite passes

### DIFF
--- a/thread/thread.cpp
+++ b/thread/thread.cpp
@@ -1577,14 +1577,16 @@ insert_list:
             _threads.resize(n);
             pthreads = &_threads[0];
         }
+        uint64_t created = 0;
         for (uint64_t i = 0; i < n; ++i)
         {
             auto th = thread_create(start, arg, stack_size);
             if (!th) break;
             thread_enable_join(th);
             pthreads[i] = th;
+            created++;
         }
-        for (uint64_t i = 0; i < n; ++i) {
+        for (uint64_t i = 0; i < created; ++i) {
             thread_join(pthreads[i]);
         }
     }
--- a/thread/thread.cpp
+++ b/thread/thread.cpp
@@ -1565,10 +1565,10 @@ insert_list:
         return 0;
     }
 
-    void threads_create_join(uint64_t n,
+    uint64_t threads_create_join(uint64_t n,
         thread_entry start, void* arg, uint64_t stack_size)
     {
-        if (n == 0) return;
+        if (n == 0) return 0;
         thread* threads[32];
         thread** pthreads = threads;
         std::vector<thread*> _threads;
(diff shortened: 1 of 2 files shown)


thread/thread.cpp
```diff
@@ -1577,14 +1577,16 @@ R"(
             _threads.resize(n);
             pthreads = &_threads[0];
         }
+        uint64_t created = 0;
         for (uint64_t i = 0; i < n; ++i)
         {
             auto th = thread_create(start, arg, stack_size);
             if (!th) break;
             thread_enable_join(th);
             pthreads[i] = th;
```
lihuiba: pthreads[created++] = th;

thread/thread.cpp
```diff
@@ -1577,14 +1577,16 @@ R"(
...
+            created++;
         }
-        for (uint64_t i = 0; i < n; ++i) {
+        for (uint64_t i = 0; i < created; ++i) {
             thread_join(pthreads[i]);
         }
```
lihuiba, line 1590: should it return actual # of threads created?

thread/thread.cpp
jfeng18: Done — applied `pthreads[created++] = th;` and removed the separate `created++` line. Equivalent to the old `pthreads[i]` since the loop breaks (not continues) on the first failure, so `created == i` at every assignment.

thread/thread.cpp
jfeng18: Agreed — changed the return type to `uint64_t` and it now returns `created`, the number of threads actually created (and therefore joined). On the success path this is `n`; on partial `thread_create` failure it's `< n`, letting callers detect it. Updated the header decl + comment accordingly. All in-repo callers discard the return value, so void→uint64_t is source/ABI-compatible (mangled name and SysV calling convention unchanged).

thread/thread.cpp
lihuiba: no they are not equivalent. When you use i, the thread pointers are stored sparsely. But you use them later as dense records in the join loop.

lihuiba (approved):

thread/thread.cpp
jfeng18: Right — with `created++` the filled slots are always contiguous `[0, created)`, which is exactly what the join loop needs. Thanks for the approval.

PR #1254, opened 2026-06-04
Fix resource leak in photon::init() on partial failure

jfeng18:
## Summary

Clean up already-initialized resources when photon::init() fails partway through.

## Changes

| File | What changed | What was NOT changed |
|------|-------------|---------------------|
| `photon.cpp` | Error paths now `goto fail` → `fini()` instead of bare `return -1`; g_event_engine/g_io_engine bits set incrementally as each engine succeeds | Success path behavior unchanged (g_event_engine/g_io_engine fully assigned at end) |

## Design

Uses the `INIT_IO_OR_FAIL` macro (file-local, `#undef`'d after use) to replace `INIT_IO` in the init function. Each successful init `|=` its bit into `g_io_engine`/`g_event_engine`. On failure, `fini()` checks these bits and only tears down what was actually initialized.

## Test plan

- [ ] All existing tests pass (the success path is unchanged)
- [ ] Verified: photon.cpp compiles cleanly with gcc 10.2.1

### DIFF
--- a/photon.cpp
+++ b/photon.cpp
@@ -101,26 +101,40 @@ int __photon_init(uint64_t event_engine, uint64_t io_engine, const PhotonOptions
     if (event_engine & ALL_ENGINES) {
         for (auto x : recommended_order) {
             if ((x & event_engine) && init_event_engine(x, event_engine, options) == 0) {
+                g_event_engine |= x;
                 goto next;
             }
         }
-        LOG_ERROR_RETURN(0, -1, "All master engines init failed");
+        LOG_ERROR("All master engines init failed");
+        goto fail;
     }
 next:
-    if ((INIT_EVENT_SIGNAL & event_engine) && sync_signal_init() < 0)
-        return -1;
+    if (INIT_EVENT_SIGNAL & event_engine) {
+        if (sync_signal_init() < 0)
+            goto fail;
+        g_event_engine |= INIT_EVENT_SIGNAL;
+    }
+
+#define INIT_IO_OR_FAIL(name, prefix, ...) \
+    if (INIT_IO_##name & io_engine) { \
+        if (prefix##_init(__VA_ARGS__) < 0) goto fail; \
+        g_io_engine |= INIT_IO_##name; \
+    }
 
 #ifdef ENABLE_FSTACK_DPDK
-    INIT_IO(FSTACK_DPDK, fstack_dpdk);
+    INIT_IO_OR_FAIL(FSTACK_DPDK, fstack_dpdk);
 #endif
-    INIT_IO(EXPORTFS, exportfs)
+    INIT_IO_OR_FAIL(EXPORTFS, exportfs)
 #ifdef ENABLE_CURL
-    INIT_IO(LIBCURL, libcurl)
+    INIT_IO_OR_FAIL(LIBCURL, libcurl)
 #endif
 #ifdef __linux__
-    INIT_IO(LIBAIO, libaio_wrapper, options.libaio_queue_depth)
-    INIT_IO(SOCKET_EDGE_TRIGGER, et_poller)
+    INIT_IO_OR_FAIL(LIBAIO, libaio_wrapper, options.libaio_queue_depth)
+    INIT_IO_OR_FAIL(SOCKET_EDGE_TRIGGER, et_poller)
 #endif
+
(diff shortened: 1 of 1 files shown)


photon.cpp
```diff
@@ -101,26 +101,40 @@ int __photon_init(uint64_t event_engine, uint64_t io_engine, const PhotonOptions
     if (event_engine & ALL_ENGINES) {
         for (auto x : recommended_order) {
             if ((x & event_engine) && init_event_engine(x, event_engine, options) == 0) {
+                g_event_engine |= x;
```
lihuiba: how do you use g_event_engine?

photon.cpp
```diff
@@ -101,26 +101,40 @@ int __photon_init(uint64_t event_engine, uint64_t io_engine, const PhotonOptions
...
+        LOG_ERROR("All master engines init failed");
+        goto fail;
     }
 next:
-    if ((INIT_EVENT_SIGNAL & event_engine) && sync_signal_init() < 0)
-        return -1;
+    if (INIT_EVENT_SIGNAL & event_engine) {
+        if (sync_signal_init() < 0)
+            goto fail;
+        g_event_engine |= INIT_EVENT_SIGNAL;
```
lihuiba: how do you use g_event_engine?

photon.cpp
```diff
@@ -129,6 +143,10 @@ int __photon_init(uint64_t event_engine, uint64_t io_engine, const PhotonOptions
         reset_handle_registed = true;
     }
     return 0;
+
+fail:
+    fini();
```
lihuiba: is it possible to use ```DEFER``` like 
```cpp
DEFER(if (!ok) fini());
```

lihuiba: instead of the ```goto fail;``` idiom.

lihuiba (approved):

photon.cpp
jfeng18: Revised — now g_event_engine and g_io_engine are assigned upfront (the full requested bitmask), so fini() can clean up whatever was actually initialized. The incremental |= tracking is gone.

photon.cpp
jfeng18: Done — switched to `DEFER(if (!ok) fini())` as suggested. Much cleaner: reuses the original `INIT_IO` macro unchanged, no `goto`, no custom macro. The `CURRENT` null guard in `fini()` stays to make it idempotent.

photon.cpp
jfeng18: Same as above — these incremental `|=` assignments are gone in the latest revision. Now g_event_engine is assigned the full bitmask upfront, and DEFER handles cleanup on any failure.

photon.cpp
```diff
@@ -94,19 +95,27 @@ int __photon_init(uint64_t event_engine, uint64_t io_engine, const PhotonOptions
...
     const uint64_t ALL_ENGINES =
             INIT_EVENT_EPOLL   | INIT_EVENT_EPOLL_NG |
             INIT_EVENT_IOURING | INIT_EVENT_KQUEUE |
             INIT_EVENT_SELECT  | INIT_EVENT_IOCP;
     if (event_engine & ALL_ENGINES) {
+        bool engine_ok = false;
         for (auto x : recommended_order) {
             if ((x & event_engine) && init_event_engine(x, event_engine, options) == 0) {
-                goto next;
+                engine_ok = true;
```
lihuiba: ```engine_ok``` seems not necessary, as long as you revert the code back.

photon.cpp
```diff
@@ -94,19 +95,27 @@ int __photon_init(uint64_t event_engine, uint64_t io_engine, const PhotonOptions
     if (vcpu_init() < 0)
         return -1;
 
+    bool ok = false;
+    g_event_engine = event_engine;
```
lihuiba: why move the assignments up?

photon.cpp
jfeng18: Agreed — removed `engine_ok`, restored the original `goto next` structure.

photon.cpp
jfeng18: Moved them back to the original position (end of success path, before `ok = true`). On early failure fini() sees g_io_engine=0 so it only cleans up vcpu + event engine — which is correct.

lihuiba (approved):

PR #1263, opened 2026-06-06
Fix chunked encoding 64-bit size truncation to 32-bit

jfeng18:
## Summary
- In ChunkedBodyWriter, chunk sizes were cast to `(unsigned)` 32-bit with `%x` format, silently truncating transfers >4GB. The buffer was also too small (10 bytes) for 64-bit hex values.

## Changes
- `net/http/body.cpp`: Use `%zx` format with `(size_t)` cast, increase buffer from 10 to 20 bytes in both `write()` and `writev()`

## Verification
- [x] Compilation verified (URING ON/OFF)
- [x] Full test suite: no regressions
- [x] 3-reviewer adversarial code review: unanimous APPROVE

### DIFF
--- a/net/http/body.cpp
+++ b/net/http/body.cpp
@@ -288,8 +288,8 @@ public:
     }
 
     virtual ssize_t write(const void *buf, size_t count) override {
-        char chunk_size[10];
-        auto size = snprintf(chunk_size, sizeof(chunk_size), "%x\r\n", (unsigned)count);
+        char chunk_size[20];
+        auto size = snprintf(chunk_size, sizeof(chunk_size), "%zx\r\n", count);
         if (size <= 0) return -1;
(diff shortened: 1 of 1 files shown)


jfeng18:
The gcc14 CI failure is unrelated to this change. 9 tests across completely unrelated modules (test-exportfs, test-socket, test-curl, test-vdma, client_function_test, server_function_test, websocket_test, test-rpc, test-thread) all SEGFAULT simultaneously — this indicates a CI environment issue (likely ASAN shadow memory mapping or host memory pressure), not a code regression.

Evidence: 8 out of 10 PRs submitted in the same batch passed gcc14; only this PR and #1264 hit the failure. PR #1271, which modifies the same file (kernel_socket.cpp), passed gcc14 cleanly.

Could you please re-run the gcc14 job? Thank you.

lihuiba (approved):

PR #1276, opened 2026-06-08
extfs: add resize_extfs interface for ext4 filesystem resize

OsentryO:
Description:
  ## Summary

  Add resize_extfs(IFile*, uint64_t new_size, int flags) interface for resizing ext4 filesystems, following the same pattern as make_extfs.
  
  ### Files
  - fs/extfs/resize.cpp — new, implements resize_extfs
  - fs/extfs/extfs.h — add declaration and #include <cstdint>


BigVan:
@yuchen0cc @WaberZhuang PTAL

fs/extfs/resize.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+static errcode_t refcounted_close(io_channel channel) {
+    if (--channel->refcount > 0)
+        return 0;
+    if (channel->private_data)
+        ext2fs_free_mem(&channel->private_data);
+    if (channel->name)
+        ext2fs_free_mem(&channel->name);
+    return ext2fs_free_mem(&channel);
+}
```
WaberZhuang: No need to make this a separate option, just replace the existing extfs_close in extfs_io.cpp directly.

WaberZhuang (approved):
Please add test for resize.

WaberZhuang (changes requested):
Please add test for resize.

WaberZhuang (approved):
LGTM. We'll add a resize test on the caller side.

PR #1278, opened 2026-06-08
Fix fstack recv direction, stale event leak, and TCP_NODELAY

jfeng18:
## Summary
- Fix fstack_recv/fstack_recvmsg waiting on writable instead of readable (copy-paste bug from send)
- Uncomment EV_DELETE after timeout to clean up stale kqueue events (matches kqueue.cpp)
- Replace ff_ioctl(TCP_NODELAY) with ff_setsockopt (TCP_NODELAY is a socket option, not ioctl)

## Verification
- E2E test: not feasible (requires f-stack DPDK environment)

### DIFF
--- a/io/fstack-dpdk.cpp
+++ b/io/fstack-dpdk.cpp
@@ -114,7 +114,7 @@ public:
             return 0;  // event arrived
         }
 
-        // enqueue(fd, ev, EV_DELETE, 0, CURRENT, true); // immediately
+        enqueue(fd, ev, EV_DELETE, 0, CURRENT, true); // immediately
         errno = (ret == 0) ? ETIMEDOUT : err.no;
         return -1;
     }
@@ -239,7 +239,7 @@ int fstack_socket(int domain, int type, int protocol) {
     int val = 1;
     if (ff_ioctl(fd, FIONBIO, &val) < 0)
         LOG_WARN("failed to set socket non-blocking");
-    if (ff_ioctl(fd, TCP_NODELAY, &val) < 0)
+    if (ff_setsockopt(fd, IPPROTO_TCP, TCP_NODELAY, &val, sizeof(val)) < 0)
         LOG_WARN("failed to set TCP_NODELAY");
     return fd;
 }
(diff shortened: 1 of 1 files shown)


lihuiba (approved):

io/fstack-dpdk.cpp
```diff
@@ -114,7 +114,7 @@ class FstackDpdkEngine : public MasterEventEngine, public CascadingEventEngine,
             return 0;  // event arrived
         }
 
-        // enqueue(fd, ev, EV_DELETE, 0, CURRENT, true); // immediately
+        enqueue(fd, ev, EV_DELETE, 0, CURRENT, true); // immediately
```
beef9999, line 117: Does EV_ONESHOT really need EV_DELETE? Not sure about it

io/fstack-dpdk.cpp
```diff
@@ -239,7 +239,7 @@ int fstack_socket(int domain, int type, int protocol) {
     int val = 1;
     if (ff_ioctl(fd, FIONBIO, &val) < 0)
         LOG_WARN("failed to set socket non-blocking");
-    if (ff_ioctl(fd, TCP_NODELAY, &val) < 0)
+    if (ff_setsockopt(fd, IPPROTO_TCP, TCP_NODELAY, &val, sizeof(val)) < 0)
```
beef9999, line 242: Should be manually tested

io/fstack-dpdk.cpp
```diff
@@ -307,12 +307,12 @@ ssize_t fstack_sendmsg(int sockfd, const struct msghdr* message, int flags, Time
...
     return photon::net::DOIO_ONCE(ff_recv(sockfd, buf, count, flags),
-        g_engine->wait_for_fd_writable(sockfd, timeout));
+        g_engine->wait_for_fd_readable(sockfd, timeout));
 }
 
 ssize_t fstack_recvmsg(int sockfd, struct msghdr* message, int flags, Timeout timeout) {
     return photon::net::DOIO_ONCE(ff_recvmsg(sockfd, message, flags),
-        g_engine->wait_for_fd_writable(sockfd, timeout));
+        g_engine->wait_for_fd_readable(sockfd, timeout));
 }
```
beef9999, line 316: Can you make a new PR with only these two lines of changes?

io/fstack-dpdk.cpp
jfeng18: Yes. EV_ONESHOT disables the event after it fires, but does NOT remove it from kqueue. On timeout, the event has not fired, so it remains active with stale udata (thread pointer). If the fd later becomes ready, kqueue delivers the event with the old udata — use-after-free.

The matching implementation in io/kqueue.cpp:118 does issue EV_DELETE on timeout for exactly this reason.

io/fstack-dpdk.cpp
jfeng18: Agreed. We do not have an f-stack environment to test. The change follows the same pattern as other fstack_setsockopt calls in this file (e.g., FstackDpdkSocketServer::setsockopt).

io/fstack-dpdk.cpp
jfeng18: Sure, I can split the recv fix into a separate PR. Which two lines do you mean — the recv writable→readable changes (lines 310 and 315)?

beef9999 (approved):

PR #1279, opened 2026-06-08
Fix SubFile pwritev offset and path buffer overflow

jfeng18:
## Summary
- Fix SubFile::pwritev passing bare offset instead of offset+m_offset (data corruption on vectored writes to sub-files)
- Fix mkdir_recursive missing PATH_MAX bounds check before memcpy (stack overflow on long paths)
- Fix Walker::path_push_back assert-only bounds check (overflow in release builds)

## Verification
- E2E test: not feasible (requires specific sub-file and long-path test fixtures)

### DIFF
--- a/fs/path.cpp
+++ b/fs/path.cpp
@@ -99,8 +99,7 @@ namespace fs
 
         char path[PATH_MAX];
         if (len >= PATH_MAX - 1) {
-            errno = ENAMETOOLONG;
-            return -1;
+            LOG_ERROR_RETURN(ENAMETOOLONG, -1, "pathname too long");
         }
         if (pathname[0] != '/') {
             *path = '/';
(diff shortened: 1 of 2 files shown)


fs/path.cpp
```diff
@@ -98,6 +98,10 @@ namespace fs
         if (len == 0) return 0;
 
         char path[PATH_MAX];
+        if (len >= PATH_MAX - 1) {
+            errno = ENAMETOOLONG;
```
lihuiba: ```LOG_ERROR_RETURN``` is preferred

jfeng18: Done. Changed to LOG_ERROR_RETURN.

lihuiba (approved):

PR #1280, opened 2026-06-08
Fix cache UAF, infinite recursion, and null pointer issues

jfeng18:
## Summary
- Fix use-after-free in async_refill: save vcpu pointer before release() which may delete the store object
- Fix infinite recursion in PersistentCacheFs::open 3-arg overload calling itself
- Fix null pointer dereference in CachedFile::vioctl when source filesystem is absent
- Fix undefined behavior from nullptr in quota_pool set_quota/evict/stat
- Fix string_view size ignored in pool_store set_src_name/set_store_key

## Verification
- E2E test: not feasible (requires cache layer test fixtures)

### DIFF
--- a/fs/cache/pool_store.h
+++ b/fs/cache/pool_store.h
@@ -214,9 +214,9 @@ namespace fs
         }
 
         std::string_view get_src_name() { return src_name_; }
-        void set_src_name(std::string_view pathname) { src_name_ = pathname.data(); }
+        void set_src_name(std::string_view pathname) { src_name_.assign(pathname.data(), pathname.size()); }
         std::string_view get_store_key() { return store_key_; }
-        void set_store_key(std::string_view pathname) { store_key_ = pathname.data(); }
+        void set_store_key(std::string_view pathname) { store_key_.assign(pathname.data(), pathname.size()); }
         void set_pool(ICachePool* pool) { pool_ = pool; }
         void set_cached_size(off_t cached_size, int flags = 0);
         off_t get_actual_size() { return actual_size_; }
--- a/fs/cache/store.cpp
+++ b/fs/cache/store.cpp
@@ -191,8 +191,9 @@ void* ICacheStore::async_refill(void* args) {
 
     ctx->store->pool_->m_refilling.fetch_sub(1, std::memory_order_relaxed);
     ctx->store->range_lock_.unlock(ctx->refill_off, ctx->refill_size);
+    auto vcpu = static_cast<photon::vcpu_base*>(ctx->store->pool_->m_vcpu);
     ctx->store->release();
-    photon::thread_migrate(photon::CURRENT, static_cast<photon::vcpu_base*>(ctx->store->pool_->m_vcpu));
+    photon::thread_migrate(photon::CURRENT, vcpu);
     delete ctx;
     return nullptr;
 }
--- a/fs/cache/cached_fs.cpp
+++ b/fs/cache/cached_fs.cpp
@@ -371,6 +371,7 @@ public:
     int vioctl(int request, va_list args) override {
         IFile *src_rwfile = nullptr;
         if (cache_store_->get_src_file(&src_rwfile, O_RDWR) != 0) return -1;
+        if (!src_rwfile) LOG_ERROR_RETURN(0, -1, "source rw file is null");
         return src_rwfile->vioctl(request, args);
     }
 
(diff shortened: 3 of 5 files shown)


fs/cache/pool_store.h
```diff
@@ -214,9 +214,9 @@ namespace fs
         }
 
         std::string_view get_src_name() { return src_name_; }
-        void set_src_name(std::string_view pathname) { src_name_ = pathname.data(); }
```
lihuiba, line 217: why not using the assignment operator?

fs/cache/store.cpp
```diff
@@ -191,8 +191,9 @@ void* ICacheStore::async_refill(void* args) {
 
     ctx->store->pool_->m_refilling.fetch_sub(1, std::memory_order_relaxed);
     ctx->store->range_lock_.unlock(ctx->refill_off, ctx->refill_size);
+    auto vcpu = static_cast<photon::vcpu_base*>(ctx->store->pool_->m_vcpu);
```
lihuiba, line 194: is this necessary?

EricHuangqx (approved):

fs/cache/pool_store.h
jfeng18: Good point. `src_name_ = pathname` would work since `std::string::operator=(std::string_view)` is available in C++17. Will simplify to that.

fs/cache/store.cpp
jfeng18: Yes. `release()` can trigger `delete this` when `detached_==true` and `ref_` reaches 0 (pool_store.h PooledCacheStore destructor). After that, `ctx->store->pool_->m_vcpu` dereferences freed memory. Saving `m_vcpu` to a local before `release()` avoids the use-after-free.

fs/cache/pool_store.h
jfeng18: Actually, `src_name_ = pathname` does not compile in C++14 — `std::string::operator=(string_view)` was added in C++17, and the project defaults to C++14 where string_view is `std::experimental::string_view`.

I can change to `src_name_ = std::string(pathname)` which uses the explicit conversion + move assignment. Or keep `.assign(data, size)` which avoids the temporary. Which do you prefer?

fs/cache/store.cpp
jfeng18: Yes, here is the concrete path:

1. `CachedFs::unlink()` (cached_fs.cpp:89) calls `cache_store->release(true)` — sets `detached_=true` but does not delete (async_refill holds a ref)
2. `do_refill_range` (store.cpp:268) bumped `ref_` before spawning async_refill
3. When async_refill calls `release()` here, `ref_` drops to 0 with `detached_==true`
4. `release()` at pool_store.h:198-201: `should_delete = true` → `delete this`
5. Original code then accesses `ctx->store->pool_->m_vcpu` on the deleted object — use-after-free

The fix saves `m_vcpu` to a local before `release()` to avoid the dangling dereference.

fs/cache/pool_store.h
lihuiba: It actually compiles successfully in c++14

jfeng18: I tried `src_name_ = pathname` but it fails to compile in C++14 — the project uses `std::experimental::string_view` from C++14, which does not have an implicit conversion to `std::string`. The assignment operator `std::string::operator=(std::string_view)` was only added in C++17. Keeping `.assign(data, size)` for C++14 compatibility.

lihuiba: What's your compiler?
the default c++ standard used in CMakeLists.txt is 14, and the project compiles nicely in our dev environments, as long as the CI, which includes major version of gcc and clang.

jfeng18: GCC 10.2 on our dev machine. We already pushed the change to `= pathname` and it compiled fine. The earlier failure may have been a stale build. Thanks for confirming it works in C++14.

lihuiba (approved):

fs/cache/pool_store.h
lihuiba: Oh I made a mistake. We need to use ```assign()``` in C++14 mode. Please go back to your original proposal.

The reason why the code is compiled successfully with ```=``` lies in the trailing ```.data()```, which incurs an extra ```strlen()``` invoke.

lihuiba (approved):

PR #1281, opened 2026-06-08
Fix extfs handle leak, memory leak, and fiemap overflow

jfeng18:
## Summary
- Remove unused ext2fs_file_open in readdir that leaked a handle on every call
- Add delete for heap-allocated ext2_ino_t in get_inode follow path
- Clamp fiemap write loop to fm_extent_count to prevent heap overflow

## Verification
- Compilation: verified
- E2E test: not feasible (requires ext2 filesystem fixture)

### DIFF
--- a/fs/extfs/extfs.cpp
+++ b/fs/extfs/extfs.cpp
@@ -23,6 +23,7 @@ limitations under the License.
 #include <sstream>
 #include <sys/types.h>
 #include <unistd.h>
+#include <algorithm>
 #include <vector>
 #include <sys/sysmacros.h>
 #include <sys/statvfs.h>
@@ -710,14 +711,7 @@ static int do_ext2fs_readdir(ext2_filsys fs, const char *path, std::vector<::dir
     ext2_ino_t ino = string_to_inode(fs, path, 1);
     if (!ino) return -ENOENT;
 
-    ext2_file_t file;
-    errcode_t ret = ext2fs_file_open(
-        fs,
-        ino,  // inode,
-        0,    // flags TODO
-        &file);
-    if (ret) return parse_extfs_error(fs, ino, ret);
-    ret = ext2fs_check_directory(fs, ino);
+    errcode_t ret = ext2fs_check_directory(fs, ino);
     if (ret) return parse_extfs_error(fs, ino, ret);
     auto block_buf = (char *)malloc(fs->blocksize);
     ret = ext2fs_dir_iterate(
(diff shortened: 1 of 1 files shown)


fs/extfs/extfs.cpp
yuchen0cc: Respect fm_extent_count: when 0, only report the total count; otherwise write at most fm_extent_count entries and set fm_mapped_extents to the number actually written.

```
photon::fs::fiemap_extent *ext_buf = &map->fm_extents[0];

if (map->fm_extent_count == 0) {
    // probe mode: just report the count, don't touch fm_extents[]
    map->fm_mapped_extents = blocks.size();
    return 0;
}

size_t count = std::min(blocks.size(), (size_t)map->fm_extent_count);
map->fm_mapped_extents = count;
for (size_t i = 0; i < count; i++) {
    LOG_DEBUG("find block ` `", blocks[i].first * fs->blocksize,
                                  blocks[i].second * fs->blocksize);
    ext_buf[i].fe_physical = blocks[i].first * fs->blocksize;
    ext_buf[i].fe_length   = blocks[i].second * fs->blocksize;
}
return 0;
```

jfeng18: Good suggestion — handling fm_extent_count==0 as probe mode is more correct per the fiemap API contract. Will update.

yuchen0cc (approved):
LGTM

fs/extfs/extfs.cpp
```diff
@@ -954,9 +948,14 @@ static int do_ext2fs_fiemap(ext2_filsys fs, ext2_ino_t ino, struct photon::fs::f
...
-    map->fm_mapped_extents = blocks.size();
+    if (map->fm_extent_count == 0) {
+        map->fm_mapped_extents = blocks.size();
+        return 0;
+    }
     photon::fs::fiemap_extent *ext_buf = &map->fm_extents[0];
-    for (size_t i = 0; i < blocks.size(); i++) {
+    size_t count = std::min(blocks.size(), (size_t)map->fm_extent_count);
+    map->fm_mapped_extents = count;
+    for (size_t i = 0; i < count; i++) {
```
WaberZhuang, line 958: Shall we set flag `FIEMAP_EXTENT_LAST` according to https://www.kernel.org/doc/Documentation/filesystems/fiemap.txt ?

jfeng18: Good point. The original code did not set FIEMAP_EXTENT_LAST either, so adding it would be a behavior change beyond this fix scope. Worth tracking as a follow-up improvement.

WaberZhuang: Makes sense, we will revisit that later. Thanks for the fix!

jfeng18:
The `arm-linux-build-and-test` failure is in the Test step — all builds passed. This PR only modifies `fs/extfs/extfs.cpp`. The same ARM Test flaky pattern appears on PR #1307 (which modifies `rpc/rpc.cpp`). Could you please re-run?

PR #1282, opened 2026-06-08
Fix FUSE splice fd leak and httpfs prefix matching

jfeng18:
## Summary
- Fix fuse_pipe::flush leaking /dev/null fd on every call
- Fix strciprefix implementing strcmp instead of prefix check

## Verification
- Compilation: verified
- E2E test: not feasible (requires FUSE mount)

### DIFF
--- a/fs/fuse_adaptor/session_loop.cpp
+++ b/fs/fuse_adaptor/session_loop.cpp
@@ -56,11 +56,9 @@ struct fuse_pipe {
     int pipefd[2] = {-1, -1};
     ssize_t flush() {
         int nuldev_fd = open("/dev/null", O_WRONLY);
-        if (nuldev_fd < 0) {
-            LOG_ERROR("fuse-adaptor failed to open null device: `", strerror(errno));
-            return -1;
-        }
-
+        if (nuldev_fd < 0)
+            LOG_ERRNO_RETURN(0, -1, "fuse-adaptor failed to open null device");
+        DEFER(close(nuldev_fd));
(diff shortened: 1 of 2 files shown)


fs/fuse_adaptor/session_loop.cpp
lihuiba: ```LOG_ERRNO_RETURN``` is preferred

fs/fuse_adaptor/session_loop.cpp
```diff
@@ -61,7 +61,9 @@ struct fuse_pipe {
             return -1;
         }
 
-        return splice(pipefd[0], NULL, nuldev_fd, NULL, size, SPLICE_F_MOVE);
```
lihuiba, line 64: ```DEFER(close(nuldev_fd));``` is preferred

fs/httpfs/httpfs.cpp
lihuiba: ```estring_view::istarts_with``` is preferred over ```strciprefix```

fs/fuse_adaptor/session_loop.cpp
jfeng18: Will fix, thanks.

fs/fuse_adaptor/session_loop.cpp
jfeng18: Will fix, thanks.

fs/httpfs/httpfs.cpp
jfeng18: Will fix, thanks.

lihuiba (approved):

PR #1283, opened 2026-06-08
Fix alog snprintf overflow and iovector slice/iterator bugs

jfeng18:
## Summary
- Fix LogFormatter snprintf return value exceeding buf.size on truncation, causing stack buffer overflow
- Fix iovector_view::slice() not advancing cumulative position for skipped elements
- Fix iov_iterator::operator+= reading past array end on last element

## Verification
- Compilation: verified
- E2E test: test-alog PASS on ECS

### DIFF
--- a/common/iovector.cpp
+++ b/common/iovector.cpp
@@ -78,27 +78,33 @@ ssize_t iovector_view::slice(size_t count, off_t offset, iovector_view* /*OUT*/ 
         return -1;
     }
     int cnt = 0;
-    off_t pos = 0;
     auto ptr = iov->iov;
     ssize_t ret = 0;
-    for (auto &p : *this) {
-        if (count && (pos + (off_t)p.iov_len > offset)) {
-            if (offset > pos) {
-                ptr[cnt].iov_base = (char*)p.iov_base + (offset - pos);
-                ptr[cnt].iov_len = p.iov_len - (offset - pos);
-            } else {
-                ptr[cnt].iov_base = p.iov_base;
-                ptr[cnt].iov_len = p.iov_len;
-            }
-            if (count < ptr[cnt].iov_len)
-                ptr[cnt].iov_len = count;
-            pos += p.iov_len;
-            ret += ptr[cnt].iov_len;
-            count -= ptr[cnt].iov_len;
-            cnt ++;
-            if (cnt == iov->iovcnt)
-                break;
+    auto it = begin();
+    auto e = end();
+    // skip iovec elements before offset
+    off_t pos = 0;
+    for (; it != e; ++it) {
+        if (pos + (off_t)it->iov_len > offset)
+            break;
+        pos += it->iov_len;
+    }
+    for (; it != e && count; ++it) {
+        if (offset > pos) {
+            ptr[cnt].iov_base = (char*)it->iov_base + (offset - pos);
+            ptr[cnt].iov_len = it->iov_len - (offset - pos);
+        } else {
+            ptr[cnt].iov_base = it->iov_base;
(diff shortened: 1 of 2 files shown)


jfeng18:
The gcc921 `Test io_uring` failure is unrelated to this change. This PR only modifies `common/alog.h` (snprintf return value capping) and `common/iovector.cpp` (slice position tracking and iterator bounds) — no io_uring code is touched.

16 other PRs submitted in the same batch all passed gcc921. This is a CI environment flaky.

Could you please re-run the gcc921 job? Thank you.

common/iovector.cpp
```diff
@@ -98,6 +98,8 @@ ssize_t iovector_view::slice(size_t count, off_t offset, iovector_view* /*OUT*/
             cnt ++;
             if (cnt == iov->iovcnt)
                 break;
+        } else {
```
lihuiba: should we use a prio loop to skip enough iovec elements such that the first remaining one must be used?

jfeng18: Good idea. A prior skip loop would be cleaner — the first element in the main loop is then always used. Will refactor.

common/iovector.cpp
```diff
@@ -78,27 +78,33 @@ ssize_t iovector_view::slice(size_t count, off_t offset, iovector_view* /*OUT*/
...
+    auto e = end();
+    // skip iovec elements before offset
+    off_t pos = 0;
+    for (; it != e; ++it) {
+        if (pos + (off_t)it->iov_len > offset)
+            break;
+        pos += it->iov_len;
+    }
+    for (; it != e && count; ++it) {
+        if (offset > pos) {
```
lihuiba: process the first element before loop

jfeng18: Done. Now the skip loop only skips complete elements, first partial element is processed before the main loop, and the main loop has no branching.

common/iovector.cpp
```diff
@@ -78,27 +78,37 @@ ssize_t iovector_view::slice(size_t count, off_t offset, iovector_view* /*OUT*/
...
+    auto e = end();
+    // skip iovec elements before offset
+    off_t pos = 0;
+    for (; it != e; ++it) {
+        if (pos + (off_t)it->iov_len > offset)
+            break;
+        pos += it->iov_len;
+    }
+    // process the first element (may be partial due to offset)
+    if (it != e && count) {
```
lihuiba: if (unlikely(count == 0)), we can return eariler.

common/iovector.cpp
```diff
@@ -78,27 +78,37 @@ ssize_t iovector_view::slice(size_t count, off_t offset, iovector_view* /*OUT*/
...
+        ptr[cnt].iov_len = it->iov_len - (offset - pos);
+        if (count < ptr[cnt].iov_len)
+            ptr[cnt].iov_len = count;
+        ret += ptr[cnt].iov_len;
+        count -= ptr[cnt].iov_len;
+        cnt++;
+        ++it;
+    }
+    // process remaining full elements
+    for (; it != e && count && cnt < iov->iovcnt; ++it) {
```
lihuiba: given that we have processed the first element specifically, we'd better processed the last element specifically too.

common/iovector.cpp
jfeng18: Done. Added early return for count==0.

common/iovector.cpp
jfeng18: Done. Both first element and last element (count exhausted) now handled with early return, so the main loop body only processes full elements.

lihuiba (approved):

PR #1284, opened 2026-06-08
Fix lockfree queue init and ObjectCache move/lock bugs

jfeng18:
## Summary
- Fix LockfreeBatchMPMCRingQueue write_head/read_tail uninitialized (add {0})
- Fix ObjectCacheV2::Borrow default move causing double-release (explicit move ops with source nullification and old-state release including LRU list requeue)
- Fix ObjectCacheBase::release() iterator used after lock release (hold lock across find and dereference)

## Verification
- Code review: adversarial review passed (3 independent reviewers)
- Compilation: verified
- E2E test: test-objcache PASS on ECS

### DIFF
--- a/common/expirecontainer.cpp
+++ b/common/expirecontainer.cpp
@@ -159,7 +159,12 @@ void* ObjectCacheBase::ref_release(ItemPtr item, bool recycle, bool destroy) {
 // the argument `key` plays the roles of (type-erased) key
 void* ObjectCacheBase::release(const Item& key_item, bool recycle,
                                bool destroy) {
-    auto item = ExpireContainerBase::TypedIterator<Item>(Base::find(key_item));
-    if (item == end()) return nullptr;
-    return ref_release(*item, recycle, destroy);
+    ItemPtr item;
+    {
+        SCOPED_LOCK(_lock);
+        auto it = ExpireContainerBase::TypedIterator<Item>(Base::__find_prelock(key_item));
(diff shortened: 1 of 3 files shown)


Coldwings (approved):

lihuiba:
> * Fix ObjectCacheBase::release() iterator used after lock release (hold lock across find and dereference)

@jfeng18 wait, it seems that you didn't hold lock across find and dereference.

jfeng18:
Thanks for the second look! I believe the lock does cover both: in the merged `release()`, `__find_prelock()`, the `it == end()` check, and `item = *it` are all inside one `SCOPED_LOCK(_lock)` block (`expirecontainer.cpp:164–168`) — only the copied raw `Item*` is used afterward, so the old "dereference the iterator after `find()` already unlocked" hazard is gone.

The one out-of-lock step is `ref_release(item, …)` on that raw pointer (it re-acquires `_lock` internally). It stays valid because `release(key)` decrements the caller's own outstanding `acquire(key)` reference, so `_refcnt ≥ 1` across the gap — and a referenced item is popped off `_list`, so `expire()` skips it and nothing frees it until the count reaches 0. This find→ref_release gap is unchanged from before this PR.

Did you have a specific interleaving in mind? Happy to dig in further if I'm missing something.

lihuiba:
Oh I see.
As ref_release() has another lock, I thought you were trying to merge them.
Now I see that the assignment were not protected by the lock.

PR #1285, opened 2026-06-08
Fix TLS null ctx crash, OpenSSL leaks, and curl init

jfeng18:
## Summary
- Fix NULL dereference when SSL_CTX_new fails (add early return)
- Fix X509 cert leak in set_cert (add X509_free)
- Fix EVP_PKEY leak in set_pkey (add EVP_PKEY_free)
- Move BIO_meth_set_* calls into static initializer (fix data race)
- Fix curl init partial failure leaking resources and preventing retry

## Verification
- Compilation: verified
- E2E test: not feasible (TLS tests crash on main too, pre-existing)


net/security-context/tls-stream.cpp
```diff
@@ -210,6 +210,7 @@ class TLSContextImpl : public TLSContext {
         if (ctx == nullptr) {
             ERR_error_string_n(ERR_get_error(), errbuf, MAX_ERRSTRING_SIZE);
             LOG_ERROR(0, -1, "Failed to initial TLS: ", errbuf);
+            return;
```
lihuiba: LOG_ERROR_RETURN is preferred

net/curl.cpp
```diff
@@ -231,6 +231,18 @@ class CurlResetHandle : public ResetHandle {
 };
 static thread_local CurlResetHandle *reset_handler = nullptr;
 
+static void libcurl_init_cleanup() {
```
lihuiba: use a DEFER({...}); inside libcurl_init() instead.

lihuiba (approved):

lihuiba:
@jfeng18 please see the comments above

net/security-context/tls-stream.cpp
jfeng18: Done. Changed to LOG_ERROR_RETURN.

net/curl.cpp
jfeng18: Done. Now using DEFER with a success flag to handle cleanup on any early return.

net/curl.cpp
```diff
@@ -239,18 +239,37 @@ int libcurl_init(long flags, long pipelining, long maxconn) {
         cctx.g_loop->start();
         cctx.g_timer =
             new photon::Timer(-1UL, {nullptr, &on_timer}, true, 8UL * 1024 * 1024);
-        if (!cctx.g_timer)
+
+        bool ok = false;
+        DEFER({
+            if (!ok) {
```
lihuiba: ```ok``` seems not necessary

jfeng18: You mean remove the `ok` flag and let DEFER always clean up, with the success path nullifying the pointers before return? Like:
```cpp
DEFER({
    delete cctx.g_timer; cctx.g_timer = nullptr;
    ...
});
// on success, set pointers to nullptr so DEFER is no-op
cctx.g_timer = nullptr; // already owned
return 0;
```
Or do you prefer no DEFER at all and just clean up in each error path?

net/curl.cpp
```diff
@@ -239,18 +239,37 @@ int libcurl_init(long flags, long pipelining, long maxconn) {
...
+                delete cctx.g_timer;
+                cctx.g_timer = nullptr;
+                if (cctx.g_loop) {
+                    cctx.g_loop->stop();
+                    delete cctx.g_loop;
```
lihuiba: we may simply use ```DEFER(libcurl_fini());```, if we let cctx.g_loop's dtor invoke stop().

net/curl.cpp
jfeng18: will fix

lihuiba (approved):

lihuiba:
FstackDpdkSocketServer compilation failed:
```
/__w/PhotonLibOS/PhotonLibOS/net/kernel_socket.cpp: In member function 'virtual photon::net::ISocketStream* photon::net::FstackDpdkSocketServer::accept(photon::net::EndPoint*)':
/__w/PhotonLibOS/PhotonLibOS/net/kernel_socket.cpp:766:63: error: no matching function for call to 'photon::net::FstackDpdkSocketServer::do_accept(photon::net::EndPoint&)'
  766 |         int cfd = remote_endpoint ? do_accept(*remote_endpoint) : do_accept();
      |                                                               ^
/__w/PhotonLibOS/PhotonLibOS/net/kernel_socket.cpp:784:9: note: candidate: 'virtual int photon::net::FstackDpdkSocketServer::do_accept(sockaddr*, socklen_t*)'
  784 |     int do_accept(struct sockaddr* addr, socklen_t* addrlen) override {
      |         ^~~~~~~~~
/__w/PhotonLibOS/PhotonLibOS/net/kernel_socket.cpp:784:9: note:   candidate expects 2 arguments, 1 provided
/__w/PhotonLibOS/PhotonLibOS/net/kernel_socket.cpp:766:77: error: no matching function for call to 'photon::net::FstackDpdkSocketServer::do_accept()'
  766 |         int cfd = remote_endpoint ? do_accept(*remote_endpoint) : do_accept();
      |                                                                             ^
/__w/PhotonLibOS/PhotonLibOS/net/kernel_socket.cpp:784:9: note: candidate: 'virtual int photon::net::FstackDpdkSocketServer::do_accept(sockaddr*, socklen_t*)'
  784 |     int do_accept(struct sockaddr* addr, socklen_t* addrlen) override {
      |         ^~~~~~~~~
/__w/PhotonLibOS/PhotonLibOS/net/kernel_socket.cpp:784:9: note:   candidate expects 2 arguments, 0 provided
```

net/curl.cpp
```diff
@@ -239,18 +240,24 @@ int libcurl_init(long flags, long pipelining, long maxconn) {
         cctx.g_loop->start();
         cctx.g_timer =
             new photon::Timer(-1UL, {nullptr, &on_timer}, true, 8UL * 1024 * 1024);
-        if (!cctx.g_timer)
+
+        DEFER(if (!cctx.g_libcurl_multi) libcurl_fini());
```
lihuiba: put the ```DEFER``` after the following ```if```, which doesn't need the ```{ }```

net/curl.cpp
```diff
@@ -272,9 +279,11 @@ int libcurl_init(long flags, long pipelining, long maxconn) {
 void libcurl_fini() {
     delete cctx.g_timer;
     cctx.g_timer = nullptr;
-    cctx.g_loop->stop();
-    delete cctx.g_loop;
-    cctx.g_loop = nullptr;
+    if (cctx.g_loop) {
+        cctx.g_loop->stop();
```
lihuiba: move the invocation of ```stop()``` to the destructor of ```g_loop```, so we don't need a ```if``` here.

jfeng18:
The fstack compilation error is from PR #1286 (already merged). The `do_accept(sockaddr*, socklen_t*)` override in `FstackDpdkSocketServer` hides the base class overloads. Fix: add `using KernelSocketServer::do_accept;` to bring back the hidden overloads. Will submit a fix PR.

net/curl.cpp
jfeng18: Done. DEFER now placed after the timer null check.

net/curl.cpp
jfeng18: Done. Added `loop->stop()` to `~cURLLoop()` and removed the explicit `stop()` + `if` guard from `libcurl_fini()`.

lihuiba (approved):

PR #1286, opened 2026-06-08
Fix fstack server missing destructor and m_opts shadow

jfeng18:
## Summary
- Add destructor override using fstack_shutdown/fstack_close instead of inherited ::close
- Remove private FstackSockOptBuffer m_opts that shadowed base class m_opts
- Add accept() override to apply socket options via fstack_setsockopt

## Verification
- Compilation: verified
- E2E test: not feasible (requires DPDK environment)

### DIFF
--- a/net/kernel_socket.cpp
+++ b/net/kernel_socket.cpp
@@ -712,6 +712,14 @@ class FstackDpdkSocketServer : public KernelSocketServer {
 public:
     using KernelSocketServer::KernelSocketServer;
 
+    ~FstackDpdkSocketServer() override {
+        terminate();
+        if (m_listen_fd < 0) return;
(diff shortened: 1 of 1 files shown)


jfeng18:
The fstack Build failure needs investigation. This PR modifies code inside `#ifdef ENABLE_FSTACK_DPDK`, so it could be related. Unfortunately I cannot reproduce locally (no f-stack SDK). Could you share the build error or re-run? The code compiles cleanly with `ENABLE_FSTACK_DPDK=ON` on a standard GCC 12 environment with fstack headers stubbed, and `photon_obj` builds successfully.

lihuiba (approved):

PR #1287, opened 2026-06-08
Fix socket server thread leak and static buffer race

jfeng18:
## Summary
- Fix terminate() skipping thread_interrupt/join when waiting=false, causing worker UAF
- Fix skip_read using static buffer shared across OS threads (remove static keyword)

## Verification
- Compilation: verified
- E2E test: not feasible (requires specific terminate timing)

### DIFF
--- a/net/kernel_socket.cpp
+++ b/net/kernel_socket.cpp
@@ -300,11 +300,9 @@ public:
         if (!workth) return;
         auto th = workth;
         workth = nullptr;
-        if (waiting) {
-            thread_interrupt(th);
-            if (!m_block_serv)
-                thread_join((join_handle*)th);
-        }
+        thread_interrupt(th);
+        if (!m_block_serv)
+            thread_join((join_handle*)th);
     }
 
     ISocketServer* set_handler(Handler handler) override {


net/basic_socket.cpp
```diff
@@ -225,19 +225,19 @@ ssize_t sendfile_n(ISocketStream* out_stream,
...
+        ssize_t n_write = out_stream->write(buf, s);
+        if (n_write != (ssize_t) s)
             LOG_ERRNO_RETURN(0, (ssize_t)-1, "failed to write to stream ", out_stream);
         return n_write;
     };
     return doio_loop(func, BufStep(count));
 }
 
 bool ISocketStream::skip_read(size_t count) {
-    static char buf[1024];
```
lihuiba, line 240: static is ok, as the data is discarded

net/kernel_socket.cpp
```diff
@@ -300,11 +300,9 @@ class KernelSocketServer : public SocketServerBase {
         if (!workth) return;
         auto th = workth;
         workth = nullptr;
-        if (waiting) {
-            thread_interrupt(th);
-            if (!m_block_serv)
-                thread_join((join_handle*)th);
-        }
+        thread_interrupt(th);
```
lihuiba, line 303: if not in ```waiting``` state, we don't need to interrrupt it.

net/basic_socket.cpp
jfeng18: Fair point — the data is indeed discarded and the race is benign in practice. Will revert this change.

net/kernel_socket.cpp
jfeng18: Agreed we may not need to interrupt. But should we still join the thread? Without join, the worker thread continues running after terminate() returns. If the caller then destroys the server object, the worker accesses freed members (m_listen_fd, m_handler, etc.) when it resumes from thread_usleep.

Would it be acceptable to skip interrupt but keep the join?

lihuiba: yes, join is till needed

lihuiba (approved):

PR #1288, opened 2026-06-08
Fix HTTP range parsing, cookie jar, and body writev cap

jfeng18:
## Summary
- Fix Range header off-by-two in substring extraction (+1 should be -1)
- Fix cookie header corruption when first cookie in iteration is filtered (use bool flag instead of begin() check)
- Fix writev sending excess data beyond Content-Length (truncate iov to capped count)

## Verification
- Compilation: verified
- E2E test: client_function_test, server_function_test, cookie_jar_test all PASS on ECS

### DIFF
--- a/net/http/body.cpp
+++ b/net/http/body.cpp
@@ -251,9 +251,16 @@ public:
 
     virtual ssize_t writev(const struct iovec *iov, int iovcnt) override {
         auto count = iovector_view((struct iovec*)iov, iovcnt).sum();
-        if (m_cnt + count > m_size) LOG_WARN("data size overflow");
         ssize_t wc = std::min(count, m_size - m_cnt);
-        if (m_stream->writev(iov, iovcnt) != wc) return -1;
+        if (m_cnt + count > m_size) {
+            LOG_WARN("data size overflow");
+            SmartCloneIOV<32> ciovec(iov, iovcnt);
+            iovector_view view(ciovec.ptr, iovcnt);
+            view.shrink_to(wc);
+            if (m_stream->writev(view.iov, view.iovcnt) != wc) return -1;
+        } else {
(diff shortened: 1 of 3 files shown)


net/http/body.cpp
```diff
@@ -253,7 +253,10 @@ class BodyWriteStream: public WOStream {
         auto count = iovector_view((struct iovec*)iov, iovcnt).sum();
         if (m_cnt + count > m_size) LOG_WARN("data size overflow");
         ssize_t wc = std::min(count, m_size - m_cnt);
-        if (m_stream->writev(iov, iovcnt) != wc) return -1;
+        SmartCloneIOV<32> ciovec(iov, iovcnt);
```
lihuiba: do it only in case that m_cnt + count > m_size

jfeng18: Good point — no need to clone/shrink on the normal path. Will wrap the truncation in the overflow check.

net/http/body.cpp
```diff
@@ -259,9 +251,16 @@ class BodyWriteStream: public WOStream {
 
     virtual ssize_t writev(const struct iovec *iov, int iovcnt) override {
         auto count = iovector_view((struct iovec*)iov, iovcnt).sum();
-        if (m_cnt + count > m_size) LOG_WARN("data size overflow");
         ssize_t wc = std::min(count, m_size - m_cnt);
```
lihuiba: caculate ```wc``` in following individual paths, to avoid an extra branching

jfeng18: Done. Now `wc` is calculated in each branch separately — overflow path gets `m_size - m_cnt`, normal path gets `count` (cast to ssize_t).

lihuiba (approved):

PR #1289, opened 2026-06-08
Fix rsocket partial I/O retry and serialize null deref

jfeng18:
## Summary
- Fix rsocket read/write/readv/writev not advancing buffer on partial I/O (replace do_io_fully_action with explicit loops)
- Fix rsocket timeout returning 0 (ambiguous with EOF), now returns -1 with ETIMEDOUT
- Fix serialize null pointer dereference on malformed data (add null check)

## Verification
- Compilation: verified
- E2E test: not feasible (requires rsocket/RDMA environment)

### DIFF
--- a/net/rsocket/rsocket.cpp
+++ b/net/rsocket/rsocket.cpp
@@ -142,20 +142,8 @@ struct RSockFD {
             }
             return x;
         } while (tmo.timeout());
-        return 0;
-    }
-
-    template <int EVENT, typename FUNC, typename... ARGS>
-    ssize_t do_io_fully_action(uint64_t t, FUNC f, size_t n, ARGS... args) {
-        Timeout tmo(t);
-        ssize_t x = 0;
-        while (x < (ssize_t)n) {
-            auto ret = do_io_action<EVENT>(tmo.timeout(), f, args...);
-            if (ret < 0) return ret;
-            if (ret == 0) return x;
-            x += ret;
-        }
-        return x;
+        errno = ETIMEDOUT;
+        return -1;
     }
 
     ssize_t recv(uint64_t timeout, void* buf, size_t count, int flags = 0) {
@@ -167,13 +155,31 @@ struct RSockFD {
         return do_io_action<POLLIN>(timeout, rrecvmsg, msg, flags);
     }
     ssize_t read(uint64_t timeout, void* buf, size_t count) {
-        return do_io_fully_action<POLLIN>(timeout, rrecv, count, buf, count, 0);
+        Timeout tmo(timeout);
+        ssize_t total = 0;
+        while ((size_t)total < count) {
+            auto ret = do_io_action<POLLIN>(tmo.timeout(), rrecv, buf, count - total, 0);
+            if (ret < 0) return ret;
+            if (ret == 0) break;
+            total += ret;
+            buf = (char*)buf + ret;
+        }
+        return total;
     }
     ssize_t readv(uint64_t timeout, const struct iovec* iov, int iovcnt) {
         iovector_view viov{(iovec*)iov, iovcnt};
-        tmp_msg_hdr msg(viov);
-        return do_io_fully_action<POLLIN>(timeout, rrecvmsg, viov.sum(), msg,
-                                          0);
+        size_t count = viov.sum();
+        Timeout tmo(timeout);
+        ssize_t total = 0;
+        while ((size_t)total < count) {
+            tmp_msg_hdr msg(viov);
+            auto ret = do_io_action<POLLIN>(tmo.timeout(), rrecvmsg, msg, 0);
+            if (ret < 0) return ret;
+            if (ret == 0) break;
+            total += ret;
+            viov.extract_front(ret);
+        }
+        return total;
     }
     ssize_t send(uint64_t timeout, const void* buf, size_t count,
                  int flags = 0) {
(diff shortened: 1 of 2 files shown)


jfeng18:
The gcc1121 `Test epoll` failure is unrelated to this change. This PR only modifies `net/rsocket/rsocket.cpp` and `rpc/serialize.h` — no epoll code is touched. Other PRs in the same batch pass gcc1121. Could you please re-run?

Coldwings (changes requested):
The summary says "fix not advancing buffer on partial I/O", but read/write/readv/writev aren't supposed to advance the buffer in the first place — that's the caller's responsibility under standard POSIX semantics. So this isn't really a bug fix.What concerns me more is that we lost the do_io_fully_action template and now have 4 copies of essentially the same loop.

The summary also says "Fix rsocket timeout returning 0 (ambiguous with EOF), now returns -1 with ETIMEDOUT", but it is not a fact. do_io_fully_action returns error when io action timed out.

jfeng18:
Thank you for the detailed review.

You're right on both points:

1. **read/write semantics**: The `_fully` variants (read_fully, write_fully etc.) are meant to retry until the full count is satisfied, but the parameter-pack capture-by-value in `do_io_fully_action` means each retry passes the same original args — the buffer pointer never advances, so partial I/O causes the beginning of the buffer to be overwritten/resent. However, I understand your point that the non-`_fully` variants follow standard POSIX semantics and don't need this.

2. **Code duplication**: Agreed, replacing the template with 4 inline loops was not a good trade-off.

3. **Timeout return value**: You're right that `do_io_fully_action` does return error on timeout — I misread the control flow.

I'll rework this PR to:
- Keep `do_io_fully_action` but fix the capture-by-value issue (pass args by reference or use explicit state)
- Keep the serialize null-check fix (separate concern)
- Drop the timeout return value change

Or if you prefer, I can split into two PRs: one for the serialize null-check (uncontroversial) and one for the rsocket retry fix (needs more thought).

Coldwings:
> Thank you for the detailed review.
> 
> You're right on both points:
> 
> 1. **read/write semantics**: The `_fully` variants (read_fully, write_fully etc.) are meant to retry until the full count is satisfied, but the parameter-pack capture-by-value in `do_io_fully_action` means each retry passes the same original args — the buffer pointer never advances, so partial I/O causes the beginning of the buffer to be overwritten/resent. However, I understand your point that the non-`_fully` variants follow standard POSIX semantics and don't need this.
> 2. **Code duplication**: Agreed, replacing the template with 4 inline loops was not a good trade-off.
> 3. **Timeout return value**: You're right that `do_io_fully_action` does return error on timeout — I misread the control flow.
> 
> I'll rework this PR to:
> 
> * Keep `do_io_fully_action` but fix the capture-by-value issue (pass args by reference or use explicit state)
> * Keep the serialize null-check fix (separate concern)
> * Drop the timeout return value change
> 
> Or if you prefer, I can split into two PRs: one for the serialize null-check (uncontroversial) and one for the rsocket retry fix (needs more thought).

Even these fully variants are telling caller only by return value about the partial result. Still, you are right, there is a bug, makes fully variants are not working as it should be. Thanks to help us discovered this problem.

It‘s OK combined in a single PR, you can simply just work on this PR branch, and you might want to rebase to/merge the latest changes of main branch.

jfeng18:
Thanks for the careful review. You are right on two of the three points, and I want to address the third with evidence.

**CLAIM 2 (lost template, 4 copies of the loop):** Agreed, that is a real regression. I will go further than restoring `do_io_fully_action` — I will delete it and re-implement read/readv/write/writev with the shared `doio_loop` + `BufStep`/`BufStepV` that `KernelSocketStream` already uses (net/kernel_socket.cpp:98-116). rsocket.cpp already includes `<photon/net/basic_socket.h>`, so it is in scope. That removes all four loops, not just dedups them. While doing this I also found a bug in my own readv/writev: they call `viov.extract_front(ret)` on a view over the caller iov, and `extract_front` mutates that array in place (common/iovector.cpp). Since readv/writev take a `const iovec*`, that mutates a const-promised buffer. kernel_socket avoids this with `SmartCloneIOV<8>`; I will do the same.

**CLAIM 3 (timeout wording):** Agreed, "Fix rsocket timeout returning 0" is inaccurate — the dominant timeout path already returns -1/ETIMEDOUT via do_poll. The literal `return 0` is only reachable in a narrow corner. I will drop that hunk.

**CLAIM 1 (read/write should not advance the buffer):** Here I would push back, because I think it conflates recv/send with read/write. In Photon ISocketStream, recv/send are the single-shot POSIX ones (socket.h: "may be LESS than count; may block once at most"), while read/readv/write/writev are the full-transfer family (stream.h: "photon-safe atomic operation"; readv takes `const iovec*`). KernelSocketStream implements read/write with DOIO_LOOP + BufStep (loops until full, advancing the buffer) and recv/send single-shot. rsocket original author followed the same split — read/write went through `do_io_FULLY_action`. The bug: `do_io_fully_action` forwarded the same `args...` (buf, count, 0) every iteration, so on a partial transfer it never advanced buf nor shrank count — partial read overwrites buf head and over-counts (falsely reports full read); partial write re-sends head and drops tail. The fd is O_NONBLOCK and rsend/rrecv can return short, so it is reachable, just masked when the first call transfers everything.

**Plan:** I will split into two PRs — (1) the rsocket fix reworked as above (delete `do_io_fully_action`, use `doio_loop` + `BufStep`/`BufStepV` + `SmartCloneIOV`, drop the timeout hunk, add a partial-I/O test), and (2) the serialize.h null-check on its own (guards a real null deref on truncated/corrupt RPC input, unrelated to rsocket). Does that work for you?

jfeng18:
Closing in favor of two scope-clean PRs that address your review:
- #1432: rsocket buffer-advancement fix, reworked (deletes `do_io_fully_action`, uses `DOIO_LOOP` + `BufStep`/`BufStepV` + `SmartCloneIOV`, no timeout change)
- #1433: the serialize.h null-check, standalone

Thanks again for the review.

PR #1290, opened 2026-06-08
Fix Timer thread leak, ~Timer hang, and thread-key constant

jfeng18:
## What this fixes

Three independent bugs in `thread/`:

1. **Timer thread leak** — `Timer::stub()` set `_th = nullptr` before exiting, so `~Timer()` short-circuited at `if (!_th) return;` and never `thread_join`-ed the joinable stub thread, leaking its stack. Fixed by removing the trailing `_th = nullptr;`.

2. **`~Timer()` hang (Timer.OneShot)** — once `_th` is no longer nulled, `~Timer()` runs its body, whose first statement was `stop()`. For an already-fired one-shot the stub has exited, so `stop()`'s `while (cancel()) _wait_ready.wait_no_lock();` waits on a notification that never comes (lost wakeup) → the `test-thread` / Timer.OneShot 3600s timeout. Fixed by dropping the `stop()` call from `~Timer()`; the destructor's existing `_repeating = false; if (_waiting) thread_interrupt(_th, ECANCELED); thread_join(_th);` already terminates and reaps the stub. (Full root-cause analysis in the thread.)

3. **thread-key destructor skip** — `deallocate_tls` used `THREAD_KEY_1STLEVEL_SIZE` instead of `THREAD_KEY_2NDLEVEL_SIZE` when iterating keys, so some per-key TLS destructors were skipped. Fixed by using the correct constant.

## Net change
3 files, 4 lines: `thread/timer.h` (−1), `thread/thread.cpp` (−1), `thread/thread-key.cpp` (±1).

## Verification
- `test-thread` Timer.OneShot / Reuse / Reapting / Reapting2 — pass.
- Full `test-thread` suite — **56/56 pass** locally (no hang, no regression).

### DIFF
--- a/thread/thread.cpp
+++ b/thread/thread.cpp
@@ -1618,7 +1618,6 @@ insert_list:
             if (!timeout)
                 timeout = _default_timeout;
         } while(_repeating);
-        _th = nullptr;
     }
     void* Timer::_stub(void* _this)
     {
--- a/thread/thread-key.cpp
+++ b/thread/thread-key.cpp
@@ -177,7 +177,7 @@ void deallocate_tls(void** tls_) {
                         thread_keys[idx].dtor(data);
                 }
             } else {
-                idx += THREAD_KEY_1STLEVEL_SIZE;
+                idx += THREAD_KEY_2NDLEVEL_SIZE;
             }
         }
         if (!tls->specific_used) {
--- a/thread/timer.h
+++ b/thread/timer.h
@@ -74,7 +74,6 @@ namespace photon
         ~Timer()
         {
             if (!_th) return;
-            stop();
             _repeating = false;
             if (_waiting)
                 thread_interrupt(_th, ECANCELED);


common/alog.cpp
```diff
@@ -405,7 +405,7 @@ class AsyncLogOutput final : public BaseLogOutput {
     std::vector<std::unique_ptr<spsc>> buf;
     std::vector<std::unique_ptr<photon::spinlock>> lock;
     int num_of_queues;
-    bool stopped = false;
+    std::atomic<bool> stopped{false};
```
lihuiba: not necessary

thread/thread.cpp
```diff
@@ -1201,7 +1199,7 @@ R"(
         return (val << 10) / freq;
     #endif
     }
-    static uint32_t last_tsc = 0;
+    static thread_local uint32_t last_tsc = 0;
```
lihuiba: not necessary

thread/thread.cpp
```diff
@@ -1618,7 +1614,8 @@ R"(
             if (!timeout)
                 timeout = _default_timeout;
         } while(_repeating);
-        _th = nullptr;
+        _waiting = true;
```
lihuiba: why made this change?

common/alog.cpp
jfeng18: OK, will revert. The write/read is across OS threads (std::thread worker vs destruct caller), technically UB per C++11, but the semaphore signal provides happens-before in practice.

thread/thread.cpp
jfeng18: OK, will revert. Aligned 32-bit writes are atomic on x86/ARM in practice.

thread/thread.cpp
jfeng18: Two bugs in the one-shot timer exit path:

1. **Memory leak**: The stub sets `_th = nullptr` before exiting. The destructor sees `_th==nullptr` and skips `thread_join`. Since the thread was made joinable (timer.h:44), the stack is never freed.

2. **stop() hang**: The stub exits the do-while loop without calling `_wait_ready.notify_all()`. If `stop()` is called while the callback is executing, it loops calling `cancel()` (returns -1 since `_waiting==false`), then waits on `_wait_ready` forever.

The fix: replace `_th = nullptr` with `_waiting = true` + `_wait_ready.notify_all()` so the destructor can join and stop() gets notified.

common/alog.cpp
jfeng18: Those changes were removed in an earlier iteration. Current PR only contains the timer fix and thread-key constant fix.

thread/thread.cpp
jfeng18: Those changes were removed in an earlier iteration. Current PR only contains the timer fix and thread-key constant fix.

thread/thread.cpp
lihuiba: ```stop()``` will interrupt ```_th``` with ```EAGAIN```, and ```_th``` will notify the caller via ```_wait_ready```.

Note that, stop() only makes the timer stop to fire. It doesn't destroy the timer.

The destructor interrupts ```_th``` with ```ECANCELED```, followed by joining ```_th```.

jfeng18: Thanks for the explanation. So the intended lifecycle is:
- `stop()` → interrupt with EAGAIN → stub loop exits naturally → notifies `_wait_ready`
- destructor → interrupt with ECANCELED → join

That means the original `_th = nullptr` is indeed wrong — it skips the join in the destructor. But `stop()` already handles the notification via EAGAIN. So our fix (removing `_th = nullptr` and adding `_waiting = true` + `notify_all`) is partially redundant — the `notify_all` duplicates what `stop()` already does.

Should we just remove `_th = nullptr` and leave everything else as-is? The destructor will then join the thread properly.

lihuiba: just remove _th = nullptr and leave everything else as-is.

jfeng18: Done. Just removed `_th = nullptr`, left everything else as-is.

lihuiba:
45: [ RUN      ] Timer.OneShot
45/62 Test #45: test-thread ........................***Timeout 3600.07 sec

jfeng18:
**Root cause of the Timer.OneShot 3600s hang.**

This builds on the leak fix already in this PR (removing `_th = nullptr` from the end of `Timer::stub()`). With `_th` no longer nulled, `~Timer()` stops short-circuiting at `if (!_th) return;` and runs its body — but its first statement is `stop()`, and that is what hangs.

For an already-fired one-shot timer the stub has exited the `do {} while(_repeating)` loop with `_waiting == false`, and never calls `_wait_ready.notify_all()` again. So `stop()`'s `while (cancel()) _wait_ready.wait_no_lock();` loops forever on a notify that never comes (lost wakeup) — `cancel()`→`reset()` returns -1 whenever `!_waiting`. In Timer.OneShot the one-shot `timer2` is destroyed first (reverse declaration order), already in its fired / loop-exited state — exactly this case (CI shows it as the 3600s test-thread timeout).

**The fix: drop the `stop()` call from `~Timer()`.** The destructor's existing `_repeating = false; if (_waiting) thread_interrupt(_th, ECANCELED); thread_join((join_handle*)_th);` already terminates and reaps the stub — exactly the lifecycle you described ("the destructor interrupts _th with ECANCELED, followed by joining _th"). Dropping `stop()` doesn't fix the leak itself (removing `_th = nullptr` did); it makes that fix *reachable* — `~Timer()` now reaches `thread_join` instead of hanging in `stop()`. For a dead one-shot, `_waiting == false`, so the interrupt is skipped and `thread_join` returns immediately on the DONE thread. `thread_join` blocks until the stub reaches `DONE`, a superset of `stop()`'s "no longer firing" guarantee, so no in-flight-callback wait is lost.

Verified locally: `test-thread` Timer.OneShot / Reuse / Reapting / Reapting2 all pass.

(Separately, `stop()` itself is still hang-prone if called on an already-finished timer or from inside its own callback — both the same lost-wakeup. Pre-existing and out of scope here; happy to file a follow-up.)

lihuiba (approved):

PR #1306, opened 2026-06-10
Fix macOS sendfile source/dest argument order

jfeng18:
## Summary
- macOS `sendfile(2)` signature: `sendfile(int fd_source, int s_dest, ...)`
- Linux `sendfile(2)` signature: `sendfile(int out_fd, int in_fd, ...)`
- Code had `::sendfile(out_fd, in_fd, ...)` matching Linux convention, completely broken on macOS
- Fix: swap to `::sendfile(in_fd, out_fd, ...)`

## Verification
- E2E test: not feasible (no macOS environment available)

### DIFF
--- a/net/basic_socket.cpp
+++ b/net/basic_socket.cpp
@@ -146,7 +146,7 @@ ssize_t sendfile(int out_fd, int in_fd, off_t *offset, size_t count,
                  Timeout timeout) {
 #ifdef __APPLE__
     off_t len = count;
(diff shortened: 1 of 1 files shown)


jfeng18:
The gcc14 `Test epoll` failure is unrelated to this change. This PR only swaps two arguments in the `__APPLE__` sendfile block in `net/basic_socket.cpp` — no epoll code is touched, and the change is not even compiled on Linux. Could you please re-run?

lihuiba (approved):

PR #1307, opened 2026-06-10
Add maximum RPC request size to prevent OOM from wire

jfeng18:
## Summary
- RPC skeleton reads `header.size` (uint32_t from wire) and allocates without upper bound
- Attacker can send crafted header causing up to ~2GB allocation → OOM
- Add `MAX_RPC_MESSAGE_SIZE` constant (64MB) and check before allocation

## Verification
- E2E test: not feasible (requires RPC server fixture with crafted input)

### DIFF
--- a/rpc/rpc.cpp
+++ b/rpc/rpc.cpp
@@ -188,6 +188,8 @@ namespace rpc {
         return new StubImpl(stream, ownership);
     }
 
+    static constexpr uint32_t MAX_RPC_MESSAGE_SIZE = 64 * 1024 * 1024; // 64MB
+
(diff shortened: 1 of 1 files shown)


jfeng18:
The `arm-linux-build-and-test` failure is in the Test step — all 6 builds (gcc9-14, C++14-23) passed successfully. This PR only adds 4 lines to `rpc/rpc.cpp` (a size check before allocation). Other PRs in the same batch have their ARM jobs cancelled or not-found, so there is no baseline to compare against. Could you please re-run or share the test failure details? Thank you.

lihuiba (approved):

PR #1314, opened 2026-06-10
feat(thread): make WorkPool ring size configurable at construction

taoting1234:
Previously the internal lock-free ring buffer capacity was hardcoded at
65536 (RING_SIZE constant). This commit makes it a constructor parameter
with 65536 as the default, preserving full backward compatibility.

The ring uses FlexLockfreeMPMCRingQueue (dynamically-sized) allocated via
FlexQueue::create(). The photon-style send/recv waiting logic (semaphore +
yield spinning) is implemented directly in WorkPool::impl rather than using
RingChannel, because RingChannel inherits from the flex queue and its
derived members would overlap with the flexible slots[] array.

Changes:
- thread/workerpool.h: add ring_size parameter (default 65536)
- thread/workerpool.cpp: switch to FlexLockfreeMPMCRingQueue with inline
  semaphore-based waiting, exception-safe construction, ring_size > 0
  validation, and runtime abort on allocation failure
- thread/test/test-pool.cpp: add tests for custom ring sizes

### DIFF
--- a/thread/workerpool.cpp
+++ b/thread/workerpool.cpp
@@ -22,6 +22,7 @@ limitations under the License.
 #include <photon/thread/thread.h>
 
 #include <algorithm>
+#include <cassert>
 #include <future>
 #include <random>
 #include <thread>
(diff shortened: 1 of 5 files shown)


thread/workerpool.cpp
```diff
@@ -31,20 +32,25 @@ namespace photon {
...
+    using FlexRing = FlexLockfreeMPMCRingQueue<Delegate<void>>;
+
     photon::spinlock worker_lock;
     std::vector<std::thread> owned_std_threads;
     std::vector<photon::vcpu_base *> vcpus;
     std::atomic<uint64_t> vcpu_index{0};
-    photon::common::RingChannel<
-        LockfreeMPMCRingQueue<Delegate<void>, RING_SIZE>>
-        ring;
+    FlexRing* ring;
```
Coldwings: You might would like to use `RingChannel<FlexRing>`, so no needs to re-implement the whole receive-wait-notify stuff in `WorkPool::impl`.

taoting1234: Thanks for the suggestion. Unfortunately, `RingChannel<FlexRing>` doesn't work due to a memory layout conflict:

`RingChannel` inherits from `QueueType` and adds `queue_sem`, `idler`, and yield fields **after** the queue. When `QueueType` uses a flexible array (N=0), the `slots[0]` zero-length array sits at the same offset as RingChannel's members.

Memory layout with `RingChannel<FlexLockfreeMPMCRingQueue<T>>`:
offset 0:    LockfreeRingQueueBase members (capacity, mask, tail, head)
offset ~192: slots[0] (zero-length)          ← in LockfreeMPMCRingQueue
offset ~192: queue_sem                       ← in RingChannel (SAME ADDRESS!)
offset ~232: idler
offset ~240: default_yield_turn

When `slots[i]` is written (via `push`/`pop`), it overwrites `queue_sem` and corrupts the channel. The extra memory allocated by `FlexQueue::create()` sits at the **end** of the object, but `slots[i]` accesses the **declared position** in the base class, not the trailing allocation.

This is a fundamental conflict between C++ inheritance and flexible array members. To use RingChannel here, we'd need to either:
1. Change RingChannel to use composition instead of inheritance, or
2. Move `slots[]` to the end of `RingChannel` (not in the base class)

Both are beyond the scope of this PR (making ring_size configurable in WorkPool).

The current approach uses `FlexRing` directly and implements the wait/notify logic in `WorkPool::impl`, which avoids the layout conflict while keeping the PR focused on configurability.

Coldwings (approved):
LGTM

PR #1378, opened 2026-06-14
make it possible to cross-compile to Windows target from MacOS and mingw

lihuiba:
Cross-compilation to Windows target with CMake and a new toolchain file (in MacOS):
```bash
cmake -B build-win \
  -D CMAKE_TOOLCHAIN_FILE=CMake/toolchain-windows-mingw64.cmake \
  -D CMAKE_BUILD_TYPE=MinSizeRel \
  -D PHOTON_BUILD_TESTING=ON \
  -D PHOTON_ENABLE_ECOSYSTEM=ON \
  -D PHOTON_BUILD_DEPENDENCIES=ON
```
Note that **dependencies** need to be built as well.

```
% ls libphoton*
libphoton_sole.a	libphoton.a		libphoton.dll		libphoton.dll.a
```

```
% ls *.exe
check_inet.exe				test-curl.exe				test-rpc.exe
check_inet2.exe				test-delegates.exe			test-scalepool.exe
check_inet3.exe				test-dg-socket.exe			test-server.exe
check_inet4.exe				test-digest.exe				test-singleflight.exe
check_ipaddr.exe			test-executor-async.exe			test-socket.exe
check_ipaddr2.exe			test-executor-export_as_executor.exe	test-sockpool.exe
client_function_test.exe		test-executor-std.exe			test-specific-key.exe
client_tls_test.exe			test-exportfs.exe			test-st-utest.exe
cookie_jar_test.exe			test-filecopy.exe			test-std-compat.exe
headers_test.exe			test-fs.exe				test-thread-local.exe
perf_usleepdefer_semaphore.exe		test-go-channel.exe			test-thread.exe
perf_workpool.exe			test-ipv6.exe				test-throttle-file.exe
perf-alog.exe				test-lib-data.exe			test-throttle.exe
perf-objcache.exe			test-lockfree.exe			test-tls-order-native.exe
perf-rcuptr.exe				test-memory-stream.exe			test-tls-order-photon.exe
server_function_test.exe		test-multi-vcpu-locking.exe		test-tls.exe
test-alog.exe				test-objcache.exe			test-vdma.exe
test-checksum.exe			test-ooo.exe				test-xxhash.exe
test-client.exe				test-pool.exe				websocket_test.exe
test-common.exe				test-pooled-stack-allocator.exe
test-constexprstr.exe			test-rpc-message.exe
```

* add support for Windows IOCP (I/O Completion Port).
* change all ```long``` types (as well as ```UL```/```L``` suffixes) to deterministic 64-bit integer types, and prohibits all future usage of ```long```s.
* wine + test-alog.exe (+ Rosetta2) completes successfully
image
(gtest outputs abnormally in this case)

* wine + test-thread.exe (+ Rosetta2) completes successfully
image

==========
Vibe coded with Qoder + qwen3.7max.


jfeng18 (commented):
Nice work on the Windows cross-compilation support! The overall design and coding style look good — IOCP engine aligns well with the existing epoll/kqueue/iouring pattern, and the `long` → `uint64_t` unification is thorough.

Found a few implementation issues that need fixing before merge:

**Critical (blocks compile/run):**

1. **`io/iocp.cpp` ~L944**: `new_iocp_engine()` returns `NewObj<EventEngineIOCP>()->init()`, but `init()` returns `int` (not `this`). On success this gives `nullptr`; on failure a wild pointer. Suggest: make `init()` return `EventEngineIOCP*` (like `iouringEngine::init()`).

2. **`io/iocp.cpp` ~L676-711**: `pread`/`pwrite` never set `OVERLAPPED.Offset/OffsetHigh`, so all reads/writes happen at offset 0 regardless of the `offset` parameter. (Note: `sendfile` at L789 does set them correctly — same pattern needed here.)

3. **`CMakeLists.txt` L115 + `include/unistd.h` L44-48**: CMake forces `-include common/windows_compat.h` and headers reference `common/win32_compat.cpp`, but neither file exists in this branch. Compilation will fail at the first TU.

**Medium (edge cases):**

4. `cancel_afd_poll()` drains with timeout=0 — if the cancellation completion arrives late, a ghost completion may fire after `_ov_pending=false`.
5. `include/sys/socket.h` L236: `(int*)addrlen` cast from `socklen_t*` — safe on MinGW (socklen_t=int) but worth a static_assert or comment.

Items 1-3 are small fixes (return type, two offset lines, add the compat files). Happy to help if needed. Design LGTM — looking forward to merging after the fixes. 🎉

lihuiba:
@jfeng18 :
NewObj::init returns T* 
See common/utility.h:33 for details.

Other issues are fixed with a new commit.

Coldwings (approved):
LGTM

net/http/message.cpp
```diff
@@ -270,13 +270,19 @@ size_t Message::body_size() const {
     it = headers.find("Content-Range");
     if (it != headers.end()) {
         size_t start, end;
+<<<<<<< HEAD
```
Coldwings: git conflict needs resolving.

zhenjiaseu (approved):
LGTM

jfeng18:
Thanks for the quick turnaround on the previous review points! 🎉

Did a follow-up pass and found a few latent issues in `windows_compat` that only matter if `PHOTON_ENABLE_ECOSYSTEM=ON` on Windows (currently OFF by default, so this PR is unaffected):

1. **`gmtime_r` macro inverts return-value semantics** — `#define gmtime_r(t, r) gmtime_s(r, t)` doesn't account for the fact that `gmtime_s` returns `errno_t` (0 on success), while POSIX `gmtime_r` returns the `tm*` (non-NULL on success). Code like `oss.cpp:update_gmt_date()` using `if (gmtime_r(...))` would skip the success branch on Windows.

2. **`strptime` is a no-op stub** returning `nullptr` — `oss.cpp:get_lastmodified()` would always fail.

3. **`timegm` not provided** — would cause link errors if reached.

4. Minor POSIX gaps in `windows_compat.cpp`: `truncate()` returns `errno_t` instead of `-1`, `readv()` returns `-1` on EOF instead of `0`, `pread/pwrite` don't set `errno` on `ReadFile/WriteFile` failure.

All of these are out-of-scope for this PR if ecosystem isn't targeted for Windows. Happy to file a follow-up issue if you'd prefer. The core Windows cross-compile work LGTM. 👍

lihuiba:
@jfeng18 fixed compilation with PHOTON_ENABLE_ECOSYSTEM=ON in a new commit

jfeng18:
Verified — all latent issues addressed in fdb5077. LGTM.

PR #1431, opened 2026-06-16
Fix FIEMAP extent flags check and probeFiemap reliability

xiaoyang-hhh:
## Summary

- **Fix bitmask check in `queryRefillRange`**: `fe_flags` is a bitmask where multiple `FIEMAP_EXTENT_*` flags can be set simultaneously (e.g. `FIEMAP_EXTENT_UNKNOWN | FIEMAP_EXTENT_LAST`). Using `==` only matches exact values, missing extents with additional flags. Changed to bitwise `&` in both `cache_store.cpp` and `persistent_cache.cpp`.

- **Fix `probeFiemap` false positives**: The probe used an empty file, which falsely succeeds on tmpfs (ioctl returns 0 with no extents instead of failing). On overlayfs, fiemap returns extents with `FIEMAP_EXTENT_UNKNOWN` that `queryRefillRange` skips, making the fiemap path useless. Now writes data before probing and verifies the returned extents are usable (non-zero count, no UNKNOWN flag). When detection fails, falls back to in-memory range tracking via `RangeModule`.

- **CI: add ext4 loopback for fiemap tests**: The CI container uses overlayfs which does not support reliable fiemap. Mounts a 256MB ext4 loopback at `/root/tmp` so `cache_test` exercises both the fiemap path (ext4) and the `RangeModule` fallback path (tmpfs).

## Test plan
- [x] All 16 `cache_test` tests pass locally (including `refill_range_*` on both tmpfs and ext4)
- [x] CI passes with ext4 loopback mount

### DIFF
--- a/fs/cache/full_file_cache/cache_pool.cpp
+++ b/fs/cache/full_file_cache/cache_pool.cpp
@@ -90,11 +90,27 @@ void FileCachePool::probeFiemap() {
     if (!preExists) mediaFs_->unlink(probePath);
   });
 
-  fiemap_t<1> fie(0, 1);
-  fie.fm_mapped_extents = 0;
-  if (probe->fiemap(&fie) != 0) {
+  // Write data before probing; an empty-file probe falsely succeeds on tmpfs.
+  char buf[4096] = {};
+  if (probe->pwrite(buf, sizeof(buf), 0) != (ssize_t)sizeof(buf)) {
     fiemapSupported_ = false;
-    LOG_INFO("Call fiemap failed, errno `", ERRNO());
+    LOG_INFO("fiemap probe write failed, errno `", ERRNO());
+    return;
+  }
+
+  fiemap_t<1> fie(0, sizeof(buf));
+  fie.fm_mapped_extents = 0;
+  if (probe->fiemap(&fie) != 0 || fie.fm_mapped_extents == 0) {
+    fiemapSupported_ = false;
+    LOG_INFO("fiemap not supported on this filesystem, errno `", ERRNO());
+    return;
+  }
+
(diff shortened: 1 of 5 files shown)


fs/cache/full_file_cache/cache_store.cpp
```diff
@@ -152,7 +151,7 @@ std::pair<off_t, size_t> FileCacheStore::queryRefillRange(off_t offset, size_t s
 
   for (ssize_t i = (ssize_t)(fie.fm_mapped_extents) - 1; i >= 0; i--) {
     auto& extent = fie.fm_extents[i];
-    if ((extent.fe_flags == FIEMAP_EXTENT_UNKNOWN) || (extent.fe_flags == FIEMAP_EXTENT_UNWRITTEN)) continue;
+    if ((extent.fe_flags & FIEMAP_EXTENT_UNKNOWN) || (extent.fe_flags & FIEMAP_EXTENT_UNWRITTEN)) continue;
```
lihuiba: extent.fe_flags & (FIEMAP_EXTENT_UNKNOWN | FIEMAP_EXTENT_UNWRITTEN)

xiaoyang-hhh: done

lihuiba (approved):

fs/cache/full_file_cache/cache_pool.cpp
lihuiba: let it be false by default

lihuiba: turn it true at last if it passes all checks

xiaoyang-hhh: done

PR #1466, opened 2026-06-17
HTTP client: CONNECT tunnel support

WaberZhuang:
### CONNECT tunnel for HTTPS proxy (new)

HTTPS targets through an HTTP/HTTPS proxy are now handled via CONNECT tunnels (RFC 7231 §4.3.6). Previously only HTTP/1.1 forward proxy was supported; HTTPS proxy connections would fail or produce incorrect requests.

The tunnel logic is implemented as private methods of `PooledDialer` (`make_tunnel_stream`, `do_connect_handshake`), reusing the same `ISocketPool` instance for connection grouping — no separate pool class introduced.

### `proxy_header` (new interface)

A new `CommonHeaders<> proxy_header` field is introduced on `Operation`, with a companion `CommonHeaders<>* proxy_headers()` accessor on `Client`, dedicated to proxy-handshake headers (e.g. `Proxy-Authorization`). This separates proxy headers from `req.headers` (target-server headers).

Priority chain in `call()`:

    op->proxy_header  ←  merge client->proxy_headers()  ←  insert URL Basic auth
                              (op wins, allow_dup=0)        (allow_dup=0, manual wins)

- HTTP proxy (target is HTTP): `proxy_header` is merged into `req.headers` before sending. Existing code that sets `Proxy-Authorization` in `req.headers` still works — `merge` with `allow_dup=0` preserves the destination's existing value.
- HTTPS proxy (target is HTTPS): `proxy_header` is consumed exclusively by the CONNECT handshake; `req.headers` is sent to the target server after the tunnel is established.

### `set_proxy()` behavior change

- Operation `set_proxy(url)`: now also resets `op->proxy_header` — old proxy's auth headers are no longer valid for the new proxy.
- Client `set_proxy(url)`: now also resets `client->proxy_headers()`.

Note: the old `m_proxy_auth` field is removed. Auth is now derived from the proxy URL at `call()` time and lives in `proxy_header`.

### Key-based connection pooling (internal)

All connection types — direct TCP/TLS, HTTP proxy, and CONNECT tunnel — share a single `ISocketPool` (`new_tcp_socket_pool`) via `connect(key, connector)`. Connection grouping is purely by key:

| Connection type | Key format |
|---|---|
| Direct / HTTP proxy | `<host>:<port>:<0\|1>` (0=TCP, 1=TLS) |
| CONNECT tunnel | `<proxy_host>:<proxy_port>:<s\|p>:<target_host>:<target_port>[:<auth>]` |

The `connector` lambda (passed as `TempDelegate`, zero `std::function` overhead) is the factory that establishes new connections on pool miss. Different auth credentials or target endpoints produce different keys, preventing cross-contamination.

## Verification

    client_function_test  (-server_no_resp):   18/18 PASSED
    client_tls_test       (client_tls.basic):   1/1  PASSED
    client_tls_test       (client_proxy.*):     8/8  PASSED

### DIFF
--- a/net/http/client.cpp
+++ b/net/http/client.cpp
@@ -19,6 +19,7 @@ limitations under the License.
 #include <algorithm>
 #include <random>
 #include <photon/common/alog-stdstring.h>
+#include <photon/common/estring.h>
 #include <photon/common/iovector.h>
 #include <photon/common/string_view.h>
 #include <photon/net/socket.h>
@@ -37,9 +38,9 @@ class PooledDialer {
 public:
     net::TLSContext* tls_ctx = nullptr;
     std::unique_ptr<ISocketClient> tcpsock;
-    std::unique_ptr<ISocketClient> tlssock;
     std::unique_ptr<ISocketClient> udssock;
     std::unique_ptr<Resolver> resolver;
+    std::unique_ptr<ISocketPool> tcp_pool;
     photon::mutex init_mtx;
     bool initialized = false;
     bool tls_ctx_ownership = false;
@@ -61,20 +62,18 @@ public:
             tls_ctx = new_tls_context(nullptr, nullptr, nullptr);
             tls_ctx->set_verify_mode(VerifyMode::PEER);  // act like curl
         }
-        auto tcp_cli = new_tcp_socket_client(src_ips.data(), src_ips.size());
-        auto tls_cli = new_tls_client(tls_ctx, new_tcp_socket_client(src_ips.data(), src_ips.size()), true);
-        tcpsock.reset(new_tcp_socket_pool(tcp_cli, -1, true));
-        tlssock.reset(new_tcp_socket_pool(tls_cli, -1, true));
+        tcpsock.reset(new_tcp_socket_client(src_ips.data(), src_ips.size()));
         udssock.reset(new_uds_client());
+        tcp_pool.reset(new_tcp_socket_pool(nullptr, -1ULL, false));
         resolver.reset(new_default_resolver(kDNSCacheLife));
         initialized = true;
         return 0;
     }
 
     void at_photon_fini() {
+        tcp_pool.reset();
         resolver.reset();
         udssock.reset();
-        tlssock.reset();
         tcpsock.reset();
         if (tls_ctx_ownership)
             delete tls_ctx;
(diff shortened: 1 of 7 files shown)


net/http/client.cpp
```diff
@@ -61,73 +62,172 @@ class PooledDialer {
...
+        auto stream = tcpsock->connect(proxy_ep);
+        if (!stream) {
+            resolver->discard_cache(proxy_url.host_no_port(), proxy_ip);
+            LOG_ERRNO_RETURN(0, nullptr, "CONNECT tunnel: failed to connect to proxy");
+        }
+        stream->timeout(timeout);
+        if (proxy_url.secure()) {
+            auto tls = new_tls_stream(tls_ctx, stream, SecurityRole::Client, true);
+            if (!tls) {
+                delete stream;
```
lihuiba, line 147: make use of ```DEFER``` macro to delete it in case of error

net/http/client.cpp
```diff
@@ -61,73 +62,172 @@ class PooledDialer {
...
+    // Send CONNECT request and verify 2xx response.
+    int do_connect_handshake(ISocketStream* stream, std::string_view target_host,
+                             uint16_t target_port, const Headers* headers, uint64_t timeout) {
+        char buf[4096];
+        estring authority;
+        authority.appends(target_host, ":", target_port);
+        Request req(buf, (uint16_t)sizeof(buf), Verb::CONNECT, authority);
+        if (headers) {
+            for (auto it = headers->begin(); it != headers->end(); ++it)
+                req.headers.insert(it.first(), it.second());
```
lihuiba, line 176: meger_headers?

net/http/client.cpp
```diff
@@ -61,73 +62,172 @@ class PooledDialer {
...
+    //   tunnel:       <proxy_host>:<proxy_port>:<s|p>:<target_host>:<target_port>[:<auth>]
+    static std::string make_key(std::string_view host, uint16_t port, bool secure) {
+        return estring().appends(host, ":", port, ":", secure ? "1" : "0");
     }
 
-    ISocketStream* dial(std::string_view uds_path, uint64_t timeout = -1ULL);
-};
+    static std::string make_tunnel_key(const StoredURL& proxy, std::string_view target_host,
+                                       uint16_t target_port, const Headers* headers) {
+        auto auth = headers ? headers->get_value("Proxy-Authorization") : std::string_view{};
```
lihuiba, line 99: auto auth = estring::make_conditional_cat_list(headers, ":", headers->get_value("Proxy-Authorization"));

net/http/client.cpp
```diff
@@ -61,73 +62,172 @@ class PooledDialer {
...
+        auto phost = proxy.host_no_port();
```
lihuiba, line 100: no need ```phost``` variable

PR #1484, opened 2026-06-20
refactor socket pool

lihuiba:
refactor socket pool

    1. define a new interface ```ISocketPool``` that supports connection grouping by user-defined keys, in addition to Endpoints

    2. allow for user-defined callback for connection heartbeat, as construction argument

    3. support TCP-Keepalive

    4. re-design the implementation to support these new features

    5. optimization of the data structure to eliminate redundant stores of keys (Endpoints), as well as fds

    6. merge the timer and the collector thread

    The keys (Endpoints) were stored 3 times in: unordered_map (fdmap), StreamListNode, and PooledTCPSocketStream. Now they are stored once in StreamListHead, located at the end of StreamListHead to support variable length. And the unordered_map is changed to unordered_set.

    The maintainance routines, e.g. evict(), release(), collect(), etc., are changed accordingly.

### DIFF
--- a/common/checksum/test/test_checksum.cpp
+++ b/common/checksum/test/test_checksum.cpp
@@ -169,7 +169,7 @@ TEST(Perf, crc32c_sw) {
 
 void do_test64(const char* name, CRC64ECMA crc64ecma) {
     auto start = std::chrono::system_clock::now();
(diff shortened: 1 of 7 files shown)


Coldwings (approved):
LGTM

PR #1486, opened 2026-06-20
Fix UB: Remove flexible array member from Operation class

taoting1234:
Replace char _buf[0] (flexible array member, UB in C++) with explicit buffer pointer management to avoid undefined behavior.

## Problem with Clang

When compiled with Clang under -O2 optimization, the flexible array member causes the compiler to optimize away the default constructor of Operation. This leads to uninitialized members (e.g., StoredURL proxy_url), causing SEGV crashes when the destructor attempts to free uninitialized pointers.

Example crash:
  ==ERROR: AddressSanitizer: SEGV on unknown address
  #0 URL::~URL() at url.h:58:9
  #1 StoredURL::~StoredURL() at url.h:113:5
  #2 Operation::~Operation() at client.h:142:30

The issue occurs because:
1. Flexible array member (char _buf[0]) is not standard C++
2. Clang treats the class as having trivial initialization
3. Under -O2, the constructor call is optimized away
4. Stack-allocated OperationOnStack objects have uninitialized members

## Changes

- Replace char _buf[0] with char* _buf pointing to external buffer
- Add internal constructors accepting buffer pointer parameter
- Heap-allocated Operation (via create()) uses buffer after object (this + 1)
- Stack-allocated OperationOnStack passes its _buf array to base constructor

This eliminates two sources of UB:
1. Flexible array member (char _buf[0]) - not standard C++
2. Derived class overriding base class member (char _buf[BufferSize])

The fix ensures correct behavior regardless of compiler optimizations and eliminates potential issues with -O2 or different compiler versions.

Verified: DadiCacheTest passes with ASAN enabled, no SEGV crashes.

### DIFF
--- a/net/http/client.h
+++ b/net/http/client.h
@@ -130,16 +130,33 @@ public:
         const void *body_buffer = nullptr;
         size_t body_buffer_size = 0;
 
-        char _buf[0];
+        // Avoid UB: use pointer instead of flexible array member
+        char* _buf = nullptr;
+
+        // Internal constructors accepting external buffer pointer
+        Operation(Client* c, Verb v, std::string_view url, uint16_t buf_size, char* buf)
+            : req(buf, buf_size, v, url, c->has_proxy()),
+              enable_proxy(c->has_proxy()),
+              _client(c),
+              _buf(buf) {}
+        Operation(Client* c, uint16_t buf_size, char* buf)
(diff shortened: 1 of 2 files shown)


lihuiba:
As flexible array is a standard feature of C, it is **highly unlikely** that a decent compiler breaks it with optimization on.

PR #1493, opened 2026-06-21
Fix kqueue/fstack event interest: split into per-filter kevent

### LINKED ISSUE #1269: Bug: kqueue/fstack-dpdk event interest translation produces wrong filter values
## Summary

`EventsMap::translate_bitwisely()` (events_map.h:44-50) combines kqueue filter IDs with bitwise OR, but kqueue filters are negative integers (EVFILT_READ=-1, EVFILT_WRITE=-2, EVFILT_EXCEPT=-15), not bitmasks.

Since EVFILT_READ is -1 (all bits set), any OR that includes it collapses to -1. Even EVFILT_WRITE|EVFILT_EXCEPT = 0xFFFFFFFE|0xFFFFFFF1 = 0xFFFFFFFF = -1 = EVFILT_READ — producing a completely wrong filter.

Additionally, the single-interest ternary at kqueue.cpp:108 and fstack-dpdk.cpp:108:
```cpp
(interests == EVENT_READ) ? EVFILT_READ : EVFILT_WRITE
```
silently maps EVENT_ERROR (=4) to EVFILT_WRITE instead of EVFILT_EXCEPT, so `wait_for_fd_error()` is also broken on these backends.

Linux/epoll is unaffected — epoll's EventsMap (epoll.cpp:37) uses `EVUnderlay<EPOLLIN|EPOLLRDHUP, EPOLLOUT, EPOLLERR>` where all values are positive power-of-two flags, so bitwise OR works correctly.

## Root cause

kqueue's `struct kevent` has a single `short filter` field — it is NOT a bitmask. Each kevent can only carry one filter. The EventsMap abstraction was designed for epoll (bitmask semantics) and does not translate to kqueue's per-filter model.

## Trigger paths

- `curl.cpp:134-137` passes `EVENT_READ|EVENT_WRITE` to `rm_interest` → only EVFILT_READ gets deleted, EVFILT_WRITE leaks
- `st.cpp:395-396` passes combined `POLLIN|POLLOUT` to `add_interest` → only EVFILT_READ is registered, write readiness silently dropped
- `fstack-dpdk.cpp:175,190` has the same issue

## Possible fix direction

For `add_interest`/`rm_interest`: iterate over individual interest bits and issue a separate `enqueue()` per filter, rather than combining into one value.

For the ternary: add an `EVENT_ERROR → EVFILT_EXCEPT` branch.

Would like to hear your thoughts on the preferred approach before submitting a PR.

jfeng18:
kqueue filter values are negative integers (EVFILT_READ=-1, EVFILT_WRITE=-2). The previous implementation used bitwise OR to combine them via `translate_bitwisely()`, which produces incorrect results (e.g. `-1 | -2 = -1`).

Fix: split multi-interest registrations into per-filter `enqueue()` calls, each submitting an independent kevent. Error propagation is preserved.

Closes #1269

### DIFF
--- a/io/fstack-dpdk.cpp
+++ b/io/fstack-dpdk.cpp
@@ -172,8 +172,19 @@ public:
         if (e.interests & EVENT_READ) entry.reader_data = e.data;
         if (e.interests & EVENT_WRITE) entry.writer_data = e.data;
         if (e.interests & EVENT_ERROR) entry.error_data = e.data;
-        auto events = evmap.translate_bitwisely(e.interests);
-        return enqueue(e.fd, events, EV_ADD, 0, e.data, true);
+        if (e.interests & EVENT_READ) {
+            int ret = enqueue(e.fd, evmap.UNDERLAY_EVENT_READ, EV_ADD, 0, e.data, true);
+            if (ret < 0) return ret;
+        }
+        if (e.interests & EVENT_WRITE) {
+            int ret = enqueue(e.fd, evmap.UNDERLAY_EVENT_WRITE, EV_ADD, 0, e.data, true);
+            if (ret < 0) return ret;
+        }
+        if (e.interests & EVENT_ERROR) {
+            int ret = enqueue(e.fd, evmap.UNDERLAY_EVENT_ERROR, EV_ADD, 0, e.data, true);
+            if (ret < 0) return ret;
+        }
+        return 0;
     }
 
     int rm_interest(Event e) override {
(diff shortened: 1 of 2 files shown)


lihuiba:
This seems a serious bug, but why it has been running well on macOS?

jfeng18:
All current callers pass single-bit interests, so the OR never hits multiple negative filters — that's why macOS works fine. This fix aligns the kqueue/fstack implementation with the `CascadingEventEngine` interface contract that `interests` is bitwisely OR-able.

lihuiba (approved):

PR #1494, opened 2026-06-21
Fix thread_migrate TOCTOU race and work-stealing sleepq dangling ref

### LINKED ISSUE #1294: Bug: work-stealing leaves stolen threads as dangling references in victim's sleepq
## Problem

`ws_scan_standbyq()` (thread.cpp ~line 2016) steals threads from a victim vcpu's standby queue and reassigns `th->vcpu`, but does NOT pop the thread from the victim's `sleepq`. The stolen thread remains as a stale entry in the victim's sleepq heap.

When the victim's `resume_threads_inlined()` later processes this sleepq entry, it operates on a thread that has migrated to another vcpu — corrupting `idx` values and potentially double-scheduling the thread.

## How it happens

1. vcpu-A's thread T sleeps with a timer → T enters both `sleepq` and later `standbyq` (via cross-vcpu `thread_interrupt`)
2. vcpu-B's work-stealer calls `ws_scan_standbyq()` → pops T from standbyq, sets `th->vcpu = CURRENT->vcpu`
3. vcpu-A's `resume_threads_inlined()` fires for T's timer → finds T in sleepq, but T is now on vcpu-B → corrupted wakeup

## Proposed fix

In `ws_scan_standbyq()`, after popping a thread from standbyq, pop it from the victim's sleepq as well:
```cpp
victim->sleepq.pop(th);  // before reassigning th->vcpu
```
Needs careful lock ordering — the victim's sleepq may require the victim's lock.

Alternatively, have `resume_threads_inlined()` check `th->vcpu == CURRENT->vcpu` before operating on the entry.

### LINKED ISSUE #1295: Bug: TOCTOU race in thread_migrate allows work-stealing to corrupt runq
## Problem

`thread_migrate()` (thread.cpp ~line 2137) checks `th->state` and `th->vcpu` without holding any lock, then `do_thread_migrate()` acquires the runq lock and calls `remove_from_list(th)`. A concurrent work-stealer on another OS thread can steal the thread between the unlocked checks and the lock acquisition.

## Race sequence

1. Thread A calls `thread_migrate(th)`, checks `th->vcpu == CURRENT->vcpu` (true) and `th->state == READY` (true) — no lock held
2. Thread B (work-stealer) acquires victim runq lock, removes `th`, changes `th->vcpu`, inserts into stealer's runq, releases lock
3. Thread A acquires victim runq lock, calls `remove_from_list(th)` — but `th` is now in Thread B's runq → intrusive list corruption across two runqs

## Proposed fix

Move the `th->state` and `th->vcpu` validation inside `do_thread_migrate()`, after acquiring the runq lock. Re-check under lock:
```cpp
if (th->state != states::READY || th->vcpu != CURRENT->vcpu) {
    // thread was stolen, bail out
    return -1;
}
remove_from_list(th);
```

jfeng18:
Two fixes for thread scheduling races:

1. **thread_migrate TOCTOU** (#1295): Move `SCOPED_LOCK(th->lock)` + state re-check into `do_thread_migrate()` after acquiring `runq_lock` (via `AtomicRunQ`), ensuring lock order consistency with work-stealing path (runq_lock → th->lock).

2. **work-stealing sleepq dangling ref** (#1294): Skip threads with `th->idx != -1` in `ws_scan_q()` to avoid stealing threads still referenced in the source vCPU sleepq.

Closes #1295, closes #1294

### DIFF
--- a/thread/thread.cpp
+++ b/thread/thread.cpp
@@ -2003,6 +2003,10 @@ insert_list:
             if ((possibly_running && th->state == states::RUNNING) ||
                                     !th->allow_work_stealing()) {
                 th = th->next();
+            } else if (th->idx != -1) {
+                // thread still referenced in source vCPU's sleepq; skip to
+                // avoid corrupting the heap structure after stealing.
+                th = th->next();
             } else {
                 auto next = th->remove_from_list();
                 stolen.push_back(th); count++;
@@ -2154,7 +2158,7 @@ insert_list:
         assert(vb != th->vcpu);
         AtomicRunQ arq;
         SCOPED_LOCK(th->lock);
-        if (th->state != READY) {
+        if (th->state != READY || th->vcpu != CURRENT->vcpu) {
             LOG_ERROR_RETURN(EINVAL, -1,
                 "thread ` state changed during migrate", th)
         }
--- a/thread/thread.cpp
+++ b/thread/thread.cpp
@@ -2152,7 +2152,13 @@ insert_list:
     }
     static int do_thread_migrate(thread* th, vcpu_base* vb) {
         assert(vb != th->vcpu);
-        AtomicRunQ().remove_from_list(th);
+        AtomicRunQ arq;
+        SCOPED_LOCK(th->lock);
+        if (th->state != READY) {
+            LOG_ERROR_RETURN(EINVAL, -1,
+                "thread ` state changed during migrate", th)
+        }
+        arq.remove_from_list(th);
         th->get_vcpu()->nthreads--;
         th->state = STANDBY;
         auto vcpu = (vcpu_t*)vb;
--- a/thread/thread.cpp
+++ b/thread/thread.cpp
@@ -2004,9 +2004,11 @@ insert_list:
                                     !th->allow_work_stealing()) {
                 th = th->next();
             } else if (th->idx != -1) {
-                // thread still referenced in source vCPU's sleepq; skip to
-                // avoid corrupting the heap structure after stealing.
+                // sleeping threads interrupted by another vCPU are inserted to  
+                // standby q without poping from sleeping q, they can not be stolen.
                 th = th->next();
+                // note that migrated threads are also inserted to standby q, but
+                // they are not in a sleeping q, so they are able to be stolen.
             } else {
                 auto next = th->remove_from_list();
                 stolen.push_back(th); count++;


thread/thread.cpp
```diff
@@ -2003,6 +2003,8 @@ R"(
             if ((possibly_running && th->state == states::RUNNING) ||
                                     !th->allow_work_stealing()) {
                 th = th->next();
+            } else if (th->idx != -1) {
+                th = th->next();
```
lihuiba, line 2009: so this prohibits stealing from standby queues?

jfeng18: No — `idx` is the sleepq heap index; standby threads should always have `idx == -1` (cleared on sleepq pop). This skips only threads with a stale sleepq reference, preventing heap corruption if stolen. Normal standby stealing is unaffected.

lihuiba: When a sleeping thread is interrupted from another vCPU, it is inserted into standby queue without being poped from sleep queue, until it is "officially" resumed by its own vCPU in ```resume_threads_inlined()```.

So its ```idx``` remains valid (!= -1) during in standby queue.

So the new condition added effectively prohibits stealing from standby queues. Right?

lihuiba: If standby queues can not be stolen, we'd better not scan them.

jfeng18: You're right — I missed the cross-vCPU interrupt path where `prelocked_thread_interrupt` (L1460) inserts into standbyq without `sleepq.pop`. So `idx != -1` does effectively block stealing those threads. Agree that skipping standby scan entirely would be cleaner. I'll update the PR.

jfeng18: Correction — standbyq actually contains two kinds of threads:

1. Cross-vCPU interrupted sleeping threads: `idx != -1` (still in sleepq, not stealable)
2. Migrated ready threads (from `do_thread_migrate`): `idx == -1` (already popped from sleepq, stealable)

The `idx != -1` check correctly filters case 1 while allowing case 2. If we skip standbyq scanning entirely, migrated threads become invisible to work-stealing.

So the current code is correct as-is — apologies for the premature "I'll update" earlier.

lihuiba: Oh you are right. I missed the migration case.

lihuiba (approved):

thread/thread.cpp
```diff
@@ -2152,7 +2154,13 @@ R"(
     }
     static int do_thread_migrate(thread* th, vcpu_base* vb) {
         assert(vb != th->vcpu);
-        AtomicRunQ().remove_from_list(th);
+        AtomicRunQ arq;
+        SCOPED_LOCK(th->lock);
+        if (th->state != READY) {
```
lihuiba: Considering work-stealing, we should also check whether ```th->vcpu == CURRENT->vcpu```, in addition to ```th->state```.

thread/thread.cpp
lihuiba: Perhaps we should add some comments on this condition

lihuiba (approved):

thread/thread.cpp
lihuiba: I wrote some comments to explain standby queue and migration.

PR #1503, opened 2026-06-25
fix(tls): leak BIO_METHOD singleton to avoid use-after-free at exit

### LINKED ISSUE #1502: TLS: BIO_s_sockstream() use-after-free in SSL_shutdown() during process exit
# TLS: BIO_s_sockstream() use-after-free in SSL_shutdown() during process exit

## Summary

`BIO_s_sockstream()` manages the `BIO_METHOD` singleton via a `static std::unique_ptr`. During process exit, glibc's `__cxa_finalize` frees it before worker threads that still hold TLS connections have a chance to call `SSL_shutdown()`, causing a deterministic SIGSEGV in any application that joins worker threads during static destruction.

## Crash Sequence

Given:

- A `static` object (e.g., a library context) constructed early during initialization.
- Worker threads spawned by the application, each holding a `thread_local` connection pool (`PooledDialer`) containing idle TLS connections.
- `BIO_METHOD` constructed lazily on the first TLS connection — **after** the static context.

`__cxa_finalize` destroys static-local variables in **reverse construction order**:

```
__cxa_finalize (main thread, reverse construction order)
  │
  ├─ 1. ~unique_ptr<BIO_METHOD>   ← BIO_METHOD freed (constructed late → destroyed first)
  │
  └─ 2. ~StaticContext             ← constructed early → destroyed later
       └─ join(worker_thread)      ← waits for worker thread to finish

           worker thread exits
             └─ __call_tls_dtors
                  └─ ~thread_local PooledDialer
                       └─ at_photon_fini() hook
                            └─ ~TCPSocketPool
                                 └─ delete_all()
                                      └─ ~TLSSocketStream
                                           └─ SSL_shutdown(ssl)
                                                └─ BIO_write(bio, ...)
                                                     └─ meth->bwrite  ← SIGSEGV
```

The static context's destructor acts as a **synchronization point**: it joins the worker thread, guaranteeing that the worker's `thread_local` destructors run while `BIO_METHOD` memory is already freed. This makes the crash **deterministic**, not a race condition.

## Current Code (upstream)

```cpp
static BIO_METHOD* BIO_s_sockstream() {
    static std::unique_ptr<BIO_METHOD, BIOMethodDeleter> meth([] {
        auto m = BIO_meth_new(BIO_TYPE_SOURCE_SINK, "BIO_PHOTON_SOCKSTREAM");
        BIO_meth_set_write(m, &TLSSocketStream::ssbio_bwrite);
        // ... set other callbacks ...
        return m;
    }());
    return meth.get();
}
```

The `unique_ptr` destructor calls `BIO_meth_free()` during `__cxa_finalize`, freeing the `BIO_METHOD` and its function pointer table.

## Suggested Fix

Replace the RAII-managed `unique_ptr` with a raw pointer (intentional leak):

```cpp
static BIO_METHOD* BIO_s_sockstream() {
    static BIO_METHOD* const meth = [] {
        auto m = BIO_meth_new(BIO_TYPE_SOURCE_SINK, "BIO_PHOTON_SOCKSTREAM");
        BIO_meth_set_write(m, &TLSSocketStream::ssbio_bwrite);
        // ... set other callbacks ...
        return m;
    }();
    return meth;
}
```

The leaked `BIO_METHOD` (~64 bytes) is reclaimed by the OS at process exit. Intentionally leaking heap-allocated singletons that must outlive all users is a common C/C++ practice — it avoids the static-destruction-ordering problem entirely without requiring callers to coordinate shutdown.

The `static BIO_METHOD* const` with IIFE lambda preserves:
- Thread-safe one-time initialization (C++11 static-local guarantee)
- Same code structure as the current upstream implementation
- No `BIOMethodDeleter` needed

## Environment

- Photon: current master (post commit `#1285`)
- OpenSSL: 1.1.x / 3.x
- glibc: any version with `__cxa_finalize` + `__call_tls_dtors`

## Reproduction

1. Create a static object that spawns worker threads and joins them in its destructor.
2. Worker threads use `thread_local` `PooledDialer` with TLS connections.
3. Let the process exit normally.
4. Observe deterministic SIGSEGV in `SSL_shutdown()` → `BIO_write()`.

yuchen0cc:
### Summary

Replace `unique_ptr<BIO_METHOD>` with an intentionally-leaked raw pointer in `BIO_s_sockstream()`, preventing SIGSEGV when worker threads call `SSL_shutdown()` after `__cxa_finalize` has already freed the singleton.

### Changes

- Remove `BIOMethodDeleter` struct
- Change `static std::unique_ptr<BIO_METHOD, BIOMethodDeleter>` to `static BIO_METHOD* const` (intentional leak, ~64 bytes, reclaimed by OS at exit)

Closes #1502

### DIFF
--- a/net/security-context/tls-stream.cpp
+++ b/net/security-context/tls-stream.cpp
@@ -392,12 +392,8 @@ public:
 
 #endif
 
-    struct BIOMethodDeleter {
-        void operator()(BIO_METHOD* ptr) { BIO_meth_free(ptr); }
-    };
-
     static BIO_METHOD* BIO_s_sockstream() {
-        static std::unique_ptr<BIO_METHOD, BIOMethodDeleter> meth([] {
+        static BIO_METHOD* const meth = [] {
             auto m = BIO_meth_new(BIO_TYPE_SOURCE_SINK, "BIO_PHOTON_SOCKSTREAM");
             BIO_meth_set_write(m, &TLSSocketStream::ssbio_bwrite);
             BIO_meth_set_read(m, &TLSSocketStream::ssbio_bread);
@@ -406,8 +402,8 @@ public:
             BIO_meth_set_create(m, &TLSSocketStream::ssbio_create);
             BIO_meth_set_destroy(m, &TLSSocketStream::ssbio_destroy);
             return m;
-        }());
-        return meth.get();
+        }();
+        return meth;
     }
 
     static ISocketStream* get_bio_sockstream(BIO* b) {


lihuiba:
why worker thread can be running after __cxa_finalize has already freed the singleton?

net/security-context/tls-stream.cpp
```diff
@@ -392,12 +392,8 @@ class TLSSocketStream : public ForwardSocketStream {
 
 #endif
 
-    struct BIOMethodDeleter {
-        void operator()(BIO_METHOD* ptr) { BIO_meth_free(ptr); }
-    };
-
     static BIO_METHOD* BIO_s_sockstream() {
-        static std::unique_ptr<BIO_METHOD, BIOMethodDeleter> meth([] {
+        static BIO_METHOD* const meth = [] {
```
lihuiba, line 396: ```({ ... })``` is more readable than a lambda followed by invocation.

lihuiba (approved):

net/security-context/tls-stream.cpp
yuchen0cc: > `({ ... })` is more readable than a lambda followed by invocation.

({...}) is a GNU statement expression (GCC/Clang extension, not standard C++). The current form [] { ... }() is a standard C++11 IIFE, portable across all conforming compilers. Prefer keeping it as-is.

PR #1516, opened 2026-06-26
oss mini sdk: add interfaces to upload/download data via customized body writer/reader

fangqianan:
### Changes
Add 3 OSS mini SDK interfaces, to save temporary memory allocation:  
1. put_object/upload_part (customozed body writer).  
2. get_object_range (customozed body reader).
Besides, keep the old interfaces to make projects like ossfs2 unchanged.

### Background
Currently OSS mini SDK provides interfaces like upload_part which takes preallocated buffer as the uploaded data storager. This means we have to allcoate a buffer sized of one part (for ossfs2, it's 8 MB) for the caller, which is too much.  
Here I provide some new interfaces to take a customized body_writer as the parameter, so the caller can allocate smaller buffers in the stack several times to avoid allocating buffer at once.

### DIFF
--- a/ecosystem/oss.cpp
+++ b/ecosystem/oss.cpp
@@ -344,14 +344,14 @@ class OssClientImpl : public Client {
   int fill_meta(HTTP_STACK_OP& op, ObjectHeaderMeta& meta);
   int fill_upload_response(HTTP_STACK_OP& op, ObjectUploadOptions& opts);
 
-  ssize_t get_object_range(std::string_view object, const struct iovec* iov,
-                           int iovcnt, off_t offset,
+  ssize_t get_object_range(std::string_view object, off_t offset,
+                           size_t cnt, BodyReader reader,
                            ObjectHeaderMeta* meta = nullptr);
 
   int batch_get_objects(std::vector<GetObjectParameters>& params);
 
-  ssize_t put_object(std::string_view object, const struct iovec* iov,
-                     int iovcnt, ObjectUploadOptions& opts);
+  ssize_t put_object(std::string_view object, size_t cnt,
+                     BodyWriter writer, ObjectUploadOptions& opts);
 
   ssize_t append_object(std::string_view object, const struct iovec* iov,
                         int iovcnt, off_t position,
@@ -362,8 +362,9 @@ class OssClientImpl : public Client {
 
   int init_multipart_upload(std::string_view object, void** context);
 
-  ssize_t upload_part(void* context, const struct iovec* iov, int iovcnt,
-                      int part_number, ObjectUploadOptions& opts);
+  ssize_t upload_part(void* context, size_t cnt,
+                      int part_number, BodyWriter writer,
+                      ObjectUploadOptions& opts);
 
   int upload_part_copy(void* context, off_t offset, size_t count,
                        int part_number, std::string_view from = {});
@@ -886,14 +887,12 @@ int OssClient::fill_upload_response(HTTP_STACK_OP& op,
   return 0;
 }
 
-ssize_t OssClient::get_object_range(std::string_view obj_path,
-                                    const struct iovec* iov, int iovcnt,
-                                    off_t offset, ObjectHeaderMeta* meta) {
+ssize_t OssClient::get_object_range(std::string_view obj_path, off_t offset,
+                                    size_t cnt, BodyReader reader,
+                                    ObjectHeaderMeta* meta) {
   int retry_times = m_oss_options.retry_times;
   auto retry_interval = m_oss_options.retry_base_interval_us;
 
-  iovector_view view((struct iovec*)iov, iovcnt);
-  auto cnt = view.sum();
   if (cnt == 0) return 0;
 
   bool invalidate_cache = false;
@@ -913,8 +912,7 @@ retry:
   }
 
   uint64_t content_length = op.resp.headers.content_length();
-  auto ret = op.resp.readv(iov, iovcnt);
-
+  auto ret = reader(&op.resp);
   // we have encountered partial read issue because of the socket was
   // unexpectedly closed
   if (ret != static_cast<ssize_t>(content_length)) {
@@ -1236,11 +1234,8 @@ int OssClient::batch_get_objects(std::vector<GetObjectParameters>& params) {
   return 0;
 }
 
-ssize_t OssClient::put_object(std::string_view object, const struct iovec* iov,
-                              int iovcnt, ObjectUploadOptions& opts) {
-  iovector_view view((struct iovec*)iov, iovcnt);
-  auto cnt = view.sum();
-
+ssize_t OssClient::put_object(std::string_view object, size_t cnt,
+                              BodyWriter writer, ObjectUploadOptions& opts) {
   OssUrl oss_url(m_endpoint, m_bucket, object, m_is_http);
   auto content_type = lookup_mime_type(object);
 
@@ -1249,7 +1244,14 @@ ssize_t OssClient::put_object(std::string_view object, const struct iovec* iov,
     op.req.headers.insert(OSS_HEADER_KEY_CONTENT_TYPE, content_type);
   }
   op.req.headers.content_length(cnt);
-  op.body_writer = {&view, &body_writer_cb};
+  auto body_writer_wrapper = [&writer, cnt](photon::net::http::Request* req) -> ssize_t {
+    ssize_t n = writer(req);
+    if (n != (ssize_t)cnt) {
+      LOG_ERROR_RETURN(0, -1, "body writer wrote `, expected `", n, cnt);
+    }
+    return 0;
+  };
+  op.body_writer = body_writer_wrapper;
   int r = sign_and_call(op, Verb::PUT, oss_url);
   if (r < 0) return r;
   r = verify_crc64_if_needed(op, oss_url.object(), opts.expected_crc64);
@@ -1334,13 +1336,9 @@ int OssClient::init_multipart_upload(std::string_view object, void** context) {
   return 0;
 }
 
-ssize_t OssClient::upload_part(void* context, const struct iovec* iov,
-                               int iovcnt, int part_number,
+ssize_t OssClient::upload_part(void* context, size_t cnt,
+                               int part_number, BodyWriter writer,
                                ObjectUploadOptions& opts) {
-  iovector_view view((struct iovec*)iov, iovcnt);
-  auto cnt = view.sum();
-  assert(cnt > 0);
-
   assert(context);
 
   oss_multipart_context* ctx = (oss_multipart_context*)context;
@@ -1358,7 +1356,14 @@ ssize_t OssClient::upload_part(void* context, const struct iovec* iov,
   DEFINE_ONSTACK_OP(m_client, Verb::PUT, oss_url.append_params(query_params));
 
   op.req.headers.content_length(cnt);
-  op.body_writer = {&view, &body_writer_cb};
+  auto body_writer_wrapper = [&writer, cnt](photon::net::http::Request* req) -> ssize_t {
+    ssize_t n = writer(req);
+    if (n != (ssize_t)cnt) {
+      LOG_ERROR_RETURN(0, -1, "body writer wrote `, expected `", n, cnt);
+    }
+    return 0;
+  };
+  op.body_writer = body_writer_wrapper;
   int r = sign_and_call(op, Verb::PUT, oss_url, query_params);
   if (r < 0) return r;
 
--- a/ecosystem/oss.h
+++ b/ecosystem/oss.h
@@ -174,13 +174,20 @@ class Client : public Object {
 
   // return value is the real size if the operation succeeds, otherwise
   // return -1
+  ssize_t get_object_range(std::string_view object, char* buf, size_t size,
+                           off_t offset, ObjectHeaderMeta* meta = nullptr) {
+    return get_object_range(object, offset, size,
+        [buf, size](IStream* input) -> ssize_t {
+          return input->read(buf, size);
+        }, meta);
+  }
   ssize_t get_object_range(std::string_view object, const struct iovec* iov,
                            int iovcnt, off_t offset,
                            ObjectHeaderMeta* meta = nullptr) {
     iovector_view view((struct iovec*)iov, iovcnt);
     return get_object_range(object, offset, view.sum(),
-        [&view](IStream* input) -> ssize_t {
-          return input->readv(view.iov, view.iovcnt);
+        [iov, iovcnt](IStream* input) -> ssize_t {
+          return input->readv(iov, iovcnt);
         }, meta);
   }
 
@@ -199,6 +206,13 @@ class Client : public Object {
   // return -1.
   // if expected_crc64 is specified, we will compare the value with the
   // returned object crc64 to validate the object integrity.
+  ssize_t put_object(std::string_view object, const char* buf, size_t size,
+                     ObjectUploadOptions& opts) {
+    return put_object(object, size,
+        [buf, size](IStream* output) -> ssize_t {
+          return output->write(buf, size);
+        }, opts);
+  }
   ssize_t put_object(std::string_view object, const struct iovec* iov,
                      int iovcnt, uint64_t* expected_crc64 = nullptr) {
     ObjectUploadOptions opts{.expected_crc64 = expected_crc64};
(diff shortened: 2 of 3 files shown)


zhenjiaseu:
I think we can simplify things.
1. seems read callback is not that necessary? we can just call current get_object_range interface multi-times.
2. is it possbile to make write callback one option of ObjectUploadOptions? so we don't add one new interface.

ecosystem/oss.h
```diff
@@ -150,6 +151,23 @@ struct ObjectUploadOptions {
   std::string *etag = nullptr;
 };
 
+// Callback for writing data into the HTTP request body (upload direction).
+// The caller writes data to the IStream* (which wraps the HTTP request).
+// Return 0 on success, -1 on error.
+// NOTE: The callback may be invoked more than once if the HTTP layer retries
+// on transient failures. Implementations must be idempotent — e.g., seek the
+// source back to the beginning before each write.
+using BodyWriter = Delegate<ssize_t, IStream*>;
```
lihuiba: ```TempDelegate``` is preferred, to allow for temp lambdas.

ecosystem/oss.h
lihuiba: the ```iov``` overloading, as well as a flat buffer overloading, can be inline implemented on top of the ```BodyReader``` overloading.

ecosystem/oss.h
```diff
@@ -184,6 +206,14 @@ class Client : public Object {
   virtual ssize_t put_object(std::string_view object, const struct iovec* iov,
```
lihuiba: the ```iov``` overloading, as well as a flat buffer overloading, can be inline implemented on top of the ```BodyReader``` overloading.

ecosystem/oss.h
lihuiba: the ```iov``` overloading, as well as a flat buffer overloading, can be inline implemented on top of the ```BodyReader``` overloading.

lihuiba (approved):

fangqianan:
> I think we can simplify things.
> 
> 1. seems read callback is not that necessary? we can just call current get_object_range interface multi-times.
> 2. is it possbile to make write callback one option of ObjectUploadOptions? so we don't add one new interface.

1. Calling current get_object_range interface multi-times means the user need send http requests multi times? Adding read callback can make sure the user obtain the whole buffer via only one request?
2. This makes sense.

zhenjiaseu:
> > I think we can simplify things.
> > 
> > 1. seems read callback is not that necessary? we can just call current get_object_range interface multi-times.
> > 2. is it possbile to make write callback one option of ObjectUploadOptions? so we don't add one new interface.
> 
> 1. Calling current get_object_range interface multi-times means the user need send http requests multi times? Adding read callback can make sure the user obtain the whole buffer via only one request?
> 2. This makes sense.

seems it‘s OK to add the read callback. It can be useful to download a large range or entire obj.

ecosystem/oss.h
```diff
@@ -160,9 +178,20 @@ class Client : public Object {
 
   // return value is the real size if the operation succeeds, otherwise
   // return -1
-  virtual ssize_t get_object_range(std::string_view object,
-                                   const struct iovec* iov, int iovcnt,
-                                   off_t offset,
+  ssize_t get_object_range(std::string_view object, const struct iovec* iov,
+                           int iovcnt, off_t offset,
+                           ObjectHeaderMeta* meta = nullptr) {
+    size_t cnt = 0;
```
lihuiba: auto cnt = iovector_view(iov, iovcnt).sum();

ecosystem/oss.h
```diff
@@ -160,9 +178,20 @@ class Client : public Object {
...
+    for (int i = 0; i < iovcnt; i++) cnt += iov[i].iov_len;
+    if (cnt == 0) return 0;
+    auto reader = [&](IStream* s, uint64_t) -> ssize_t {
+      return s->readv(iov, iovcnt);
+    };
+    return get_object_range(object, offset, cnt, reader, meta);
```
lihuiba: as we we ```TempDelegate```, we can:
```cpp
return get_object_range(object, offset, cnt, [&](IStream* s, uint64_t) {
  return s->readv(iov, iovcnt);
}, meta);
```
if you like.

ecosystem/oss.h
```diff
@@ -181,8 +210,25 @@ class Client : public Object {
     ObjectUploadOptions opts{.expected_crc64 = expected_crc64};
     return put_object(object, iov, iovcnt, opts);
   };
-  virtual ssize_t put_object(std::string_view object, const struct iovec* iov,
-                             int iovcnt, ObjectUploadOptions& opts) = 0;
+  ssize_t put_object(std::string_view object, const struct iovec* iov,
+                     int iovcnt, ObjectUploadOptions& opts) {
+    size_t cnt = 0;
```
lihuiba: auto cnt = iovector_view(iov, iovcnt).sum();

ecosystem/oss.h
```diff
@@ -181,8 +210,25 @@ class Client : public Object {
...
+    for (int i = 0; i < iovcnt; i++) cnt += iov[i].iov_len;
+    auto writer = [&](IStream* s) -> ssize_t {
+      auto ret = s->writev(iov, iovcnt);
+      if (ret != (ssize_t)cnt) return -1;
+      return 0;
+    };
+    return put_object(object, cnt, writer, opts);
```
lihuiba: writer can be also a temp lambda

lihuiba:
> 2\. is it possbile to make write callback one option of ObjectUploadOptions? so we don't add one new interface

👍

lihuiba:
@fangqianan you need to sign the CLA before you code is merged.

fangqianan:
> @fangqianan you need to sign the CLA before you code is merged.

OK，done

ecosystem/oss.h
```diff
@@ -150,6 +151,17 @@ struct ObjectUploadOptions {
...
+struct BodyWriteCtx {
+  void* buf;       // destination buffer to fill
+  size_t count;    // max bytes to fill
+  off_t offset;    // byte offset in the logical data source
+};
+
+// Data source callback for upload operations (pread-style).
+// Fill ctx->buf with up to ctx->count bytes from ctx->offset.
+// Retry-safe: the adapter re-reads from offset 0 on retry.
+using BodyWriter = TempDelegate<ssize_t, BodyWriteCtx *>;
```
lihuiba: why not ```TempDelegate<ssize_t, void*, size_t, off_t>``` ? 
What's the benefit of ```BodyWriteCtx```?

fangqianan: I just wanted to make the declaration shorter. `TempDelegate<ssize_t, void*, size_t, off_t>` removing `struct BodyWriteCtx` also seems to make sense. I'll change it that way.

ecosystem/oss.h
```diff
@@ -184,6 +190,9 @@ class Client : public Object {
   virtual ssize_t put_object(std::string_view object, const struct iovec* iov,
                              int iovcnt, ObjectUploadOptions& opts) = 0;
 
+  virtual ssize_t put_object(std::string_view object, size_t content_length,
+                             BodyWriter writer, ObjectUploadOptions& opts) = 0;
```
lihuiba, line 231: are you going to make BodyWriter part of ObjectUploadOptions?

ecosystem/oss.h
```diff
@@ -150,6 +150,12 @@ struct ObjectUploadOptions {
   std::string *etag = nullptr;
 };
 
+// Data source callback for upload operations (pread-style).
+// Signature: ssize_t(void* buf, size_t count, off_t offset)
+// Fill buf with up to count bytes from offset. Return bytes filled.
+// Retry-safe: the adapter re-reads from fixed offset on retry.
+using BodyWriter = TempDelegate<ssize_t, void*, size_t, off_t>;
+
```
lihuiba, line 166: will there be a reader for getting object?

fangqianan: In fact I don't need to allocate large buffer for get_object_range, so I decide to not add a reader for getting object in this PR for minimal changes.

ecosystem/oss.h
fangqianan: We later decided to not add the bodywriter into ObjectUploadOptions. Since the current interface accpets iov, which is not needed for the fd-like case. Adding bodywriter into ObjectUploadOptions may make the interface look strange.

ecosystem/oss.h
lihuiba: No, you are not only developing ossfs, but also a relatively general purpose mini SDK. So your'd better make read and write symmetric.

If later you or another one needs that for read, you or he will have to do it again with higher cost.

ecosystem/oss.cpp
```diff
@@ -318,6 +318,31 @@ static ssize_t body_writer_cb(void* iov_view, photon::net::http::Request* req) {
...
+  constexpr size_t BUF_SIZE = 64 * 1024;
+  char buf[BUF_SIZE];
+  size_t remaining = c->content_length;
+  off_t offset = 0;
+  while (remaining > 0) {
+    size_t count = std::min(remaining, BUF_SIZE);
+    ssize_t n = (*(c->writer))(buf, count, offset);
+    if (n <= 0)
+      LOG_ERROR_RETURN(0, -1, "body writer read failed at offset `", offset);
+    ssize_t w = req->write(buf, n);
```
lihuiba: http::Request is inherited from IStream, so why don't you pass req as IStream* to the user and save a temp buffer, as well as memcpy?

lihuiba: And you can make use ```http::Client::Operation::BodyWriter``` too.

ecosystem/oss.h
fangqianan: OK

ecosystem/oss.h
```diff
@@ -160,9 +172,19 @@ class Client : public Object {
...
+                           ObjectHeaderMeta* meta = nullptr) {
+    iovector_view view((struct iovec*)iov, iovcnt);
+    return get_object_range(object, view.sum(), offset,
+        [&view](IStream* input, size_t max_bytes, off_t) -> ssize_t {
+          auto ret = input->readv(view.iov, view.iovcnt);
+          return ret == (ssize_t)max_bytes ? ret : -1;
+        }, meta);
+  }
+
+  virtual ssize_t get_object_range(std::string_view object, size_t count,
```
lihuiba: offset followed by count is more friendly

ecosystem/oss.cpp
```diff
@@ -1249,7 +1268,8 @@ ssize_t OssClient::put_object(std::string_view object, const struct iovec* iov,
     op.req.headers.insert(OSS_HEADER_KEY_CONTENT_TYPE, content_type);
   }
   op.req.headers.content_length(cnt);
```
lihuiba, line 1246: is content length absolutely needed?

fangqianan: Yes, otherwise OSS will return error (or we need to specify Transfer Coding to Chunk, which may not be necessary).

ecosystem/oss.h
```diff
@@ -160,9 +172,21 @@ class Client : public Object {
...
-                                   off_t offset,
+  ssize_t get_object_range(std::string_view object, const struct iovec* iov,
+                           int iovcnt, off_t offset,
+                           ObjectHeaderMeta* meta = nullptr) {
+    iovector_view view((struct iovec*)iov, iovcnt);
+    return get_object_range(object, offset, view.sum(), 
+        [&view](IStream* input, off_t offset, size_t max_bytes) -> ssize_t {
+          struct iovec tmp[view.iovcnt];
+          iovector_view sliced(tmp, view.iovcnt);
+          view.slice(max_bytes, offset, &sliced);
```
lihuiba: why we can not assume that max_bytes == view.sum() and we need slicing here?
and we don't need them in put_object()?

ecosystem/oss.cpp
```diff
@@ -913,10 +932,23 @@ ssize_t OssClient::get_object_range(std::string_view obj_path,
   }
 
   uint64_t content_length = op.resp.headers.content_length();
-  auto ret = op.resp.readv(iov, iovcnt);
-
+  size_t remaining = content_length;
+  off_t reader_offset = 0;
   // we have encountered partial read issue because of the socket was
   // unexpectedly closed
+  while (remaining > 0) {
```
lihuiba: we can define that the readers have to do the loop it self, if they need it.

fangqianan: We designed this way on purpose, to simplify the writer/reader the user have to implement. Because the writer/reader needs to be idempotent for retrying in sign_and_call, the caller have to be more careful while implementing. By doing the loop inside the SDK, the caller just needs to read and write, which is more user-friendly.

ecosystem/oss.h
fangqianan: We encountered partial read (OSS GetObject returning buffer with smaller size), but we don't know that actual size and do a while loop outside. In case of dead loop, we need to handle this to return 0 in next round to break the loop.
And such issue doesn't go with put_object.

zhenjiaseu (approved):
LGTM

ecosystem/oss.h
```diff
@@ -160,9 +174,18 @@ class Client : public Object {
...
   // return -1
-  virtual ssize_t get_object_range(std::string_view object,
-                                   const struct iovec* iov, int iovcnt,
-                                   off_t offset,
+  ssize_t get_object_range(std::string_view object, const struct iovec* iov,
+                           int iovcnt, off_t offset,
+                           ObjectHeaderMeta* meta = nullptr) {
+    iovector_view view((struct iovec*)iov, iovcnt);
+    return get_object_range(object, offset, view.sum(),
+        [&view](IStream* input) -> ssize_t {
```
lihuiba: it better to pass iov, iovcnt to the lambda by value to avoid an extra indirection: ```[iov, iovcnt](IStream* input) -> ssize_t { ... }```

ecosystem/oss.h
```diff
@@ -181,8 +204,17 @@ class Client : public Object {
     ObjectUploadOptions opts{.expected_crc64 = expected_crc64};
     return put_object(object, iov, iovcnt, opts);
   };
-  virtual ssize_t put_object(std::string_view object, const struct iovec* iov,
-                             int iovcnt, ObjectUploadOptions& opts) = 0;
+  ssize_t put_object(std::string_view object, const struct iovec* iov,
+                     int iovcnt, ObjectUploadOptions& opts) {
+    iovector_view view((struct iovec*)iov, iovcnt);
+    return put_object(object, view.sum(),
+        [&view](IStream* output) -> ssize_t {
```
lihuiba: it better to pass iov, iovcnt to the lambda by value to avoid an extra indirection

ecosystem/oss.h
```diff
@@ -215,8 +247,17 @@ class Client : public Object {
     ObjectUploadOptions opts{.expected_crc64 = expected_crc64};
     return upload_part(context, iov, iovcnt, part_number, opts);
   };
-  virtual ssize_t upload_part(void* context, const struct iovec* iov,
-                              int iovcnt, int part_number,
+  ssize_t upload_part(void* context, const struct iovec* iov, int iovcnt,
+                      int part_number, ObjectUploadOptions& opts) {
+    iovector_view view((struct iovec*)iov, iovcnt);
+    return upload_part(context, view.sum(), part_number,
+        [&view](IStream* output) -> ssize_t {
```
lihuiba: it better to pass iov, iovcnt to the lambda by value to avoid an extra indirection

ecosystem/oss.h
```diff
@@ -160,9 +174,18 @@ class Client : public Object {
 
   // return value is the real size if the operation succeeds, otherwise
   // return -1
-  virtual ssize_t get_object_range(std::string_view object,
-                                   const struct iovec* iov, int iovcnt,
-                                   off_t offset,
+  ssize_t get_object_range(std::string_view object, const struct iovec* iov,
+                           int iovcnt, off_t offset,
```
lihuiba, line 185: consider adding an overloading for flat buffer const char* buf, size_t size

ecosystem/oss.h
```diff
@@ -181,8 +204,17 @@ class Client : public Object {
     ObjectUploadOptions opts{.expected_crc64 = expected_crc64};
     return put_object(object, iov, iovcnt, opts);
   };
-  virtual ssize_t put_object(std::string_view object, const struct iovec* iov,
-                             int iovcnt, ObjectUploadOptions& opts) = 0;
+  ssize_t put_object(std::string_view object, const struct iovec* iov,
```
lihuiba, line 221: consider adding an overloading for flat buffer const char* buf, size_t size

ecosystem/oss.h
```diff
@@ -215,8 +247,17 @@ class Client : public Object {
     ObjectUploadOptions opts{.expected_crc64 = expected_crc64};
     return upload_part(context, iov, iovcnt, part_number, opts);
   };
-  virtual ssize_t upload_part(void* context, const struct iovec* iov,
-                              int iovcnt, int part_number,
+  ssize_t upload_part(void* context, const struct iovec* iov, int iovcnt,
```
lihuiba, line 271: consider adding an overloading for flat buffer const char* buf, size_t size

ecosystem/oss.h
fangqianan: OK, added

lihuiba (approved):

PR #1526, opened 2026-07-01
set x86-64 standard to v2; disable avx512 code for macOS;

lihuiba:


CMakeLists.txt
```diff
@@ -95,13 +95,10 @@ if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
...
 
 # CMake didn't provide an abstraction of optimization level for now.
-set(CMAKE_CXX_FLAGS_DEBUG "-O0 -g")
-set(CMAKE_CXX_FLAGS_RELEASE "-O2 -DNDEBUG")
-set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O2 -DNDEBUG -g")
-set(CMAKE_CXX_FLAGS_MINSIZEREL "-O2")    # Only for CI test
-if(CMAKE_BUILD_TYPE MATCHES "MinSizeRel" AND NOT WIN32)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=native")
-endif()
+set(CMAKE_CXX_FLAGS_DEBUG "-O0 -g -march=x86-64-v2")
```
Coldwings: Arch related flags should not be add here.

Those flags should add these on line:120
These will fail compiling on aarch64.

CMakeLists.txt
Coldwings, line 112: Those flags should set here.

Coldwings (approved):

PR #1529, opened 2026-07-02
feat(tls): add custom CA certificate support for TLSContext

Moonpuck:
Add set_ca_cert() and set_ca_file() to TLSContext, allowing each HTTP client to specify custom CA certificates for peer verification, equivalent to curl's --cacert / --capath options.

Changes:
- tls-stream.h: add set_ca_cert(const char*) and set_ca_file(const char*, const char*) pure virtual methods to TLSContext interface
- tls-stream.cpp: implement both methods using OpenSSL X509_STORE and SSL_CTX_load_verify_locations; add missing #include <openssl/pem.h>; add null check for set_ca_cert(nullptr)
- client.h: document that HTTP clients within the same std::thread share TLS config and connection pool

### DIFF
--- a/net/http/client.h
+++ b/net/http/client.h
@@ -204,7 +204,9 @@ protected:
     std::vector<IPAddr> m_bind_ips;
 };
 
-//A Client without cookie_jar would ignore all response-header "Set-Cookies"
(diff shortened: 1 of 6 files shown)


Coldwings (approved):
LGTM

PR #1535, opened 2026-07-06
fix(fs): support SEEK_DATA/SEEK_HOLE for AioFileAdaptor

xiaoyang-hhh:
## What

Add an `lseek` override in `BaseFileAdaptor` ([fs/localfs.cpp](fs/localfs.cpp)) that handles `SEEK_DATA` / `SEEK_HOLE` by delegating to the kernel (`::lseek` on the real fd) and syncing the self-managed `m_offset` with the result. All other `whence` values keep the `VirtualFile` (`m_offset`-based) route.

## Why

`AioFileAdaptor` inherits `VirtualFile::lseek`, which only supports `SEEK_SET/CUR/END` and returns `-1` for `SEEK_DATA/SEEK_HOLE`. Placing the hybrid `lseek` in the shared base fixes this for both `AioFileAdaptor` and the plain `BaseFileAdaptor` path.

- `SEEK_SET/CUR/END` → `VirtualFile::lseek` (updates `m_offset`, consistent with aio's positional I/O).
- `SEEK_DATA/SEEK_HOLE` → kernel `::lseek`, then sync `m_offset` so subsequent reads start at the right place.
- `PsyncFileAdaptor` keeps its own kernel-offset `lseek` (it uses the kernel fd offset for `read`/`write`) and is unaffected.

### DIFF
--- a/fs/localfs.cpp
+++ b/fs/localfs.cpp
@@ -287,6 +287,17 @@ namespace fs
         {
             return AIOEngine::fdatasync(fd);
         }
+#ifdef __linux__
+        virtual off_t lseek(off_t offset, int whence) override
+        {
+            if (whence == SEEK_DATA || whence == SEEK_HOLE) {
+                off_t ret = UISysCall(::lseek(fd, offset, whence));
+                if (ret >= 0) m_offset = ret;
+                return ret;
+            }
+            return VirtualFile::lseek(offset, whence);
+        }
+#endif
         virtual int close() override
         {
             if (fd < 0) return 0;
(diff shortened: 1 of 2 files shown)


lihuiba:
Why not override the function in AioFileAdaptor?

lihuiba:
Add some test to verify the new feature.

xiaoyang-hhh:
> Why not override the function in AioFileAdaptor?

> Add some test to verify the new feature.

Done. Moved the `lseek` override into `AioFileAdaptor` and added test.

lihuiba (approved):

PR #1542, opened 2026-07-09
fix race-condition in new sockpool

lihuiba:
the iterator of heartbeat loop may become invalid during yield, solved with a new singly linked list.


WaberZhuang (approved):
LGTM

PR #1553, opened 2026-07-15
fix(ringchannel): block on semaphore instead of busy-yield when ring is full

taoting1234:
## Problem

`RingChannel::send<PhotonPause>` busy-yields (`while (!push) thread_yield()`) when the ring is full. This wastes CPU and, when sender vcpu count is much larger than physical CPU count, the high CPU usage of sender std-threads prevents the kernel scheduler from timely scheduling the receiver std-thread, causing recv to go a long time without running.

## Fix

Replace busy-yield with blocking wait, mirroring the existing recv-side `idler` / `queue_sem` / `pending` pattern:

- **PhotonPause**: register as `send_waiter` → yield loop → `send_sem.wait` (recv signals `send_sem` after pop via `notify_senders`, no fixed latency)
- **ThreadPause**: yield loop → `std::this_thread::sleep_for` (std threads can't use photon semaphore, uses `steady_clock` since `photon::now` is not updated outside vcpu)
- **Other Pause types**: unchanged (`while (!push) Pause::pause()`)

Shared logic extracted into `SendBackoff<T>` static methods to avoid duplication between `RingChannel` and `FlexRingChannel`. Members (`send_sem`, `send_waiters`, `send_pending`) declared in each class.

Fast path (`push` succeeds first try) has zero overhead.

## Implementation

```
send: push fail → send_waiters.fetch_add(seq_cst) → DEFER fetch_sub → yield/wait
recv: pop success → notify_senders() → send_pending CAS cap → send_sem.signal(1)
```

Uses SFINAE `enable_if` for compile-time Pause type dispatch (C++14 compatible). `notify_senders` and `push_backoff` are static methods on `SendBackoff<T>`, taking the 3 send-side members as references.

### DIFF
--- a/common/lockfree_queue.h
+++ b/common/lockfree_queue.h
@@ -623,12 +623,105 @@ namespace common {
  * LockfreeBatchMPMCRingQueue, or LockfreeSPSCRingQueue, with their own template
  * parameters.
  */
+// Shared send-side backoff logic. Members (send_sem, send_waiters,
+// send_pending) are declared in each channel class; only the logic is shared.
+template <typename T>
+struct SendBackoff {
+    static void notify_senders(photon::semaphore& send_sem,
+                               std::atomic<uint64_t>& send_waiters,
+                               std::atomic<uint64_t>& send_pending) {
+        std::atomic_thread_fence(std::memory_order_seq_cst);
+        auto cur_waiters = send_waiters.load(std::memory_order_seq_cst);
+        if (cur_waiters == 0) return;
+        auto sp = send_pending.load(std::memory_order_acquire);
+        for (;;) {
+            if (sp >= cur_waiters) {
+                auto fresh = send_waiters.load(std::memory_order_relaxed);
+                if (fresh <= cur_waiters) return;
+                cur_waiters = fresh;
+                continue;
+            }
+            if (send_pending.compare_exchange_weak(sp, sp + 1,
+                    std::memory_order_acq_rel,
+                    std::memory_order_acquire)) {
+                send_sem.signal(1);
+                return;
+            }
+        }
+    }
+
+    template <typename Pause = ThreadPause, typename PushFn,
+              typename std::enable_if<std::is_same<Pause, PhotonPause>::value, int>::type = 0>
+    static void push_backoff(const T& x, PushFn push_fn, uint64_t yield_turn, uint64_t yield_usec,
+                             photon::semaphore& send_sem,
+                             std::atomic<uint64_t>& send_waiters,
+                             std::atomic<uint64_t>& send_pending) {
+        if (!push_fn(x)) {
+            send_waiters.fetch_add(1, std::memory_order_seq_cst);
+            DEFER(send_waiters.fetch_sub(1, std::memory_order_seq_cst));
+            Timeout yield_timeout(yield_usec);
+            uint64_t yt = yield_turn;
+            while (!push_fn(x)) {
+                if (yt > 0 && !yield_timeout.expired()) {
+                    yt--;
+                    photon::thread_yield();
+                } else {
+                    // wait for 100ms
+                    int r = send_sem.wait(1, 100UL * 1000);
+                    if (r == 0)
+                        send_pending.fetch_sub(1, std::memory_order_acq_rel);
+                    yt = yield_turn;
+                    yield_timeout.timeout(yield_usec);
+                }
+            }
+        }
+    }
+
+    template <typename Pause = ThreadPause, typename PushFn,
+              typename std::enable_if<std::is_same<Pause, ThreadPause>::value, int>::type = 0>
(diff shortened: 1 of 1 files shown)


common/lockfree_queue.h
```diff
@@ -649,8 +673,50 @@ class RingChannel : public QueueType {
 
     template <typename Pause = ThreadPause>
     void send(const T& x) {
-        while (!push(x)) {
-            Pause::pause();
+        if (std::is_same<Pause, PhotonPause>::value) {
```
Coldwings: Since there is no consteval condition in C++14, using two separated implementation of send with different template parameter instead of a run-time condition might be better.

Coldwings (approved):
LGTM

common/lockfree_queue.h
```diff
@@ -747,6 +830,81 @@ class FlexRingChannel {
         : queue(q), default_yield_turn(yield_turn),
           default_yield_usec(yield_usec) {}
 
+    void notify_senders() {
```
lihuiba: The 2 groups of notify_senders、push_backoff are highly similar. Try merge them.

beef9999:
How much performance boost will this optimization bring?

taoting1234:
> @beef9999 asked: How much performance boost will this optimization bring?

This is not a performance optimization — it's a fix for a **vcpu starvation deadlock**.

**Before**: When the ring is full, `send<PhotonPause>` does `while(!push) thread_yield()` with no sleep/notification. Under load (observed: 264 senders vs 8 io-high vcpus), this creates a positive feedback loop:

1. Senders busy-yield forever, monopolizing all vcpus
2. `recv` (main_loop) starved — can't pop from ring
3. IO completion handlers starved
4. Ring stays full → senders yield more → full deadlock

Observed in production: container CPU 2013%, 892 coroutines frozen for 19+ minutes (verified by two identical pstacks), system completely unresponsive. **Throughput = 0.**

**After**: Blocked senders sleep on `send_sem` (woken by recv via `notify_senders` after pop). Vcpus are released to recv/IO handlers. No starvation, no deadlock.

The "boost" is going from **deadlock (0 throughput)** to **normal operation**. On the normal (non-saturated) path, fast path `push` succeeds on first try with zero overhead — the new code is never reached.

lihuiba (approved):

lihuiba:
> How much performance boost will this optimization bring?

I think this only reduces cpu consumption, not increasing performance.

lihuiba:
> starvation deadlock

I don't believe it.

If there are starvation or deadlock, they must be due to other reasons.

A busy yielding loop would consume all free cycles of a vCPU, but it does not block other unrelated threads.

Coldwings:
@taoting1234 
Noticed that the whole thing is only reducing CPU consuming.

It is not relevant to **vcpu starvation** since thread_yield always yields out run chance to other photon threads, so that **unresponsive** will never happen.

I can hardly get that why busy-yield sending will block receiver.
The case you said about 

> 1. Senders busy-yield forever, monopolizing all vcpus

It does makes sender vcpu never get sleep in kernel, but it is not a block to photon scheduler, since yield means "yield out current photon thread and makes next one in run-queue in running state".

> 2. `recv` (main_loop) starved — can't pop from ring

Receiver should be another vcpu, and no reason can not pop from ring since the ring is not full.

> 3. IO completion handlers starved

IO completion and work-pool task completion is not relevant to ring channel, IO completion goes with thread-local io event engine and task goes with simple semaphore, not RingChannel.

> 4. Ring stays full → senders yield more → full deadlock

The most possible cause to such situation is "Workers not able to consume more tasks".

For those senders vcpus num less than actually cpu threads num, it is not a problem.The only situation cause such scene is:
** A Work-pool Task generating a batch of tasks to the same work-pool, and needs to be done when all sub-tasks are done **
but unfortunately, this code will not help to resolve such scene.

In current state, the real problem caused such fiasco is, since sender vcpus is more than physical cpu threads, highly cpu usage makes no chance actual linux kernel scheduling receiver, and makes running extremely slow.

taoting1234:
> @Coldwings said: Noticed that the whole thing is only reducing CPU consuming. It is not relevant to **vcpu starvation** since thread_yield always yields out run chance to other photon threads, so that **unresponsive** will never happen. [...]

Thanks for the correction, you're right — my analysis was wrong.

`thread_yield` does yield to other photon threads on the same vcpu, so it doesn't block the photon scheduler. recv runs on a separate worker vcpu and is not affected by sender yield-spinning on other vcpus. IO completion goes through thread-local io event engine, not RingChannel.

The workpool nesting deadlock you described (a task submitting sub-tasks to the same workpool and synchronously waiting) is a different deadlock scenario that this PR does not fix.

What this PR actually helps with: when sender vcpu count is much larger than physical CPU count, busy-yield causes those vcpus' std threads to continuously consume high CPU, making the kernel scheduler unable to timely schedule the receiver's std thread, so recv goes a long time without running. Switching to `send_sem.wait` makes senders block and sleep, releasing CPU so the kernel can properly schedule the receiver.

PR #1554, opened 2026-07-15
build on CLANG

cecilia-nallico:
Various things have to be updated for a modern clang build to pass.

We're using this command:

```
cmake -B build -D PHOTON_BUILD_TESTING=ON -DPHOTON_CXX_STANDARD=20 -DCMAKE_C_COMPILER=/usr/bin/clang -DCMAKE_CXX_COMPILER=/usr/bin/clang++
```

- some overriding methods were missing the override.
- we had one too many attribute pop in crc.cpp
- in some examples and tests we're initializing a `char[]` from a flag which is forbidden in clang unless the VLA extension is enabled.


lihuiba:
Thanks for your contribution. Please sign the CLA so that we can accept it.

BTW, which version of clang are you using? In Linux?

We are using clang shipped with macOS these years, and it compiles fine. My current version is Apple clang version 21.0.0 (clang-2100.1.1.101).

PR #1556, opened 2026-07-16
add spill_additional_fp_simd_regs()

lihuiba:
Some additional registers are defined as callee-saved ones, such as d8 ~ d15 fp/simd registers in AARCH64, or xmm6 ~ xmm15 simd registers in Windows x64.

But in fact, it is likely in reality that they do not actually need spilling, so we leave them here for the users to manually do spilling if needed.

Some TLS callbacks in aarch64 need this extra spilling, as TLS extensively uses SIMD processing.


Coldwings (approved):
LGTM

Issue #967, opened 2025-08-27
thread stuck if destroy workpool in current thread


ivanallen:
### issue

I invoked `join_current_vcpu_into_workpool` to join the current thread, and destroy the workpool in the current thread when the program is ready to exit, which leads to the thread stuck on `photon::WorkPool::impl::~impl`.

Starting a separate thread to delete the workpool can avoid this problem.

There is no problem in Photon 0.9.0. I guess that #866 breaks something.

```
(gdb) bt
#0  0x00007ffff624759b in sched_yield () from /lib64/libc.so.6
#1  0x00000000007fb424 in __gthread_yield () at /opt/rh/gcc-toolset-12/root/usr/include/c++/12/x86_64-redhat-linux/bits/gthr-default.h:693
#2  0x00000000007fc47f in std::this_thread::yield () at /opt/rh/gcc-toolset-12/root/usr/include/c++/12/bits/std_thread.h:328
#3  0x00000000025d990f in photon::WorkPool::impl::~impl (this=0x7fffe5d47800, __in_chrg=<optimized out>) at /home/liufeng/ws/xbase/xmake_globaldir/.xmake/cache/packages/2508/p/photon/0.9.2-rc1/source/photon/thread/workerpool.cpp:59
#4  0x00000000025dc730 in std::default_delete<photon::WorkPool::impl>::operator() (this=0x50200000eff0, __ptr=0x7fffe5d47800) at /opt/rh/gcc-toolset-12/root/usr/include/c++/12/bits/unique_ptr.h:95
#5  0x00000000025dafce in std::unique_ptr<photon::WorkPool::impl, std::default_delete<photon::WorkPool::impl> >::~unique_ptr (this=0x50200000eff0, __in_chrg=<optimized out>)
    at /opt/rh/gcc-toolset-12/root/usr/include/c++/12/bits/unique_ptr.h:396
#6  0x00000000025d8bc4 in photon::WorkPool::~WorkPool (this=0x50200000eff0, __in_chrg=<optimized out>) at /home/liufeng/ws/xbase/xmake_globaldir/.xmake/cache/packages/2508/p/photon/0.9.2-rc1/source/photon/thread/workerpool.cpp:195
#7  0x0000000001e0081d in std::default_delete<photon::WorkPool>::operator() (this=0x503000074320, __ptr=0x50200000eff0) at /opt/rh/gcc-toolset-12/root/usr/include/c++/12/bits/unique_ptr.h:95
[...]

PR #1562, opened 2026-07-20
fix(executor): prevent self-join in ~ExecutorImpl when destroyed from worker thread

### LINKED ISSUE #1561: ExecutorImpl::~ExecutorImpl() self-join when destroyed from worker thread
## Problem

When a language binding with GC (e.g. pybind11, Python refcount) holds the last reference to an object owning an `Executor`, the destructor may be triggered **on the executor's own worker thread** during GC/finalization.

`~ExecutorImpl()` calls `th->join()` unconditionally → `pthread_join(self)` → `EDEADLK` → `std::system_error` escapes noexcept dtor → `std::terminate`.

Same pattern as WorkPool (#967, fixed in #969).

## Reproduction

```cpp
auto exec = std::make_shared<Executor>(...);
exec->async_perform(new auto([&exec] {
    exec.reset();  // ~ExecutorImpl on worker thread → crash
}));
```

## Fix

Detect self-destruction via `std::this_thread::get_id() == worker_id`, detach instead of join.

han-ian:
Fixes #1561

Same pattern as #967 (WorkPool), fixed in #969.

## Background

In pure C++ usage, `Executor` is typically destroyed from the creating thread, so `th->join()` is safe. The problem only surfaces when object lifetime is managed by a GC/refcount runtime (e.g. **pybind11**, Python, Java JNI):

- A task running on the executor's worker thread holds the last reference (via `py::object` / `shared_ptr`)
- When the task completes and releases the reference, cascade destruction triggers `~ExecutorImpl` **on the worker thread itself**
- `th->join()` → `pthread_join(self)` → `EDEADLK` → `std::system_error` escapes noexcept dtor → `std::terminate`

This is not reproducible in pure C++ tests without simulating the GC cascade.

## Changes

- Detect self-destruction via `worker_id` comparison
- Detach + spin-wait `running` instead of join
- Add `test_self_join.cpp` with synchronous (promise/future) verification

### DIFF
--- a/common/executor/executor.cpp
+++ b/common/executor/executor.cpp
@@ -19,6 +19,7 @@ public:
     using CBList =
         common::RingChannel<LockfreeMPMCRingQueue<Delegate<void>, 32UL * 1024>>;
     std::unique_ptr<std::thread> th;
+    std::atomic<bool> running{false};
     photon::thread *pth = nullptr;
     CBList queue;
     photon::ThreadPoolBase *pool;
@@ -34,10 +35,18 @@ public:
 
     ~ExecutorImpl() {
         queue.send({});
-        if (th)
-            th->join();
-        else
+        if (th) {
+            if (std::this_thread::get_id() == th->get_id()) {
+                // Self-destruction: called from within the worker thread.
+                th->detach();
+                while (running) photon::thread_yield();
+            } else {
+                th->join();
+            }
+        } else {
             while (pool) photon::thread_yield();
+        }
+
     }
 
     struct CallArg {
@@ -68,6 +77,8 @@ public:
     }
 
     void do_loop() {
+        running = true;
+        DEFER(running = false);
         pth = photon::CURRENT;
         pool = photon::new_thread_pool(32);
         LOG_INFO("worker start");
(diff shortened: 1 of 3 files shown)


common/executor/executor.cpp
```diff
@@ -28,16 +30,25 @@ class Executor::ExecutorImpl {
...
 
     ExecutorImpl() {}
 
     ~ExecutorImpl() {
         queue.send({});
-        if (th)
-            th->join();
-        else
+        if (th) {
+            if (std::this_thread::get_id() == worker_id) {
```
lihuiba: ```th->get_id()``` instead of ```worker_id```

common/executor/executor.cpp
```diff
@@ -19,6 +19,8 @@ class Executor::ExecutorImpl {
     using CBList =
         common::RingChannel<LockfreeMPMCRingQueue<Delegate<void>, 32UL * 1024>>;
     std::unique_ptr<std::thread> th;
+    std::atomic<bool> running{false};
+    std::thread::id worker_id;
```
Coldwings: `worker_id` might have no meaning since the thread handler able to be accessed by `th`

common/executor/executor.cpp
han-ian: Fixed.

lihuiba (approved):

PR #1569, opened 2026-07-25
fix(iouring): plug UAF on _async_io early returns

### LINKED ISSUE #1270: Bug: io_uring _async_io early-return paths produce use-after-free
## Summary

In `iouring-wrapper.cpp`, `_async_io()` uses stack-allocated `ioCtx` structs whose addresses are stored as io_uring SQE `user_data`. Multiple early-return paths destroy the stack frame before pending CQEs referencing these addresses are consumed, producing use-after-free in `reap_events()`.

Found via code review. Not yet observed as a production crash, but UAF is a deterministic source-code property — any execution that takes an early-return path will produce a dangling pointer that `reap_events` will dereference.

### 1. Lines 238-240 (timer SQE allocation failure)

The first SQE (line 220) already has `user_data=&io_ctx` (line 230) and `IOSQE_IO_LINK` (line 236). `io_uring_get_sqe` advances `sqe_tail` unconditionally, so this SQE is irrevocably queued. If `_get_sqe()` for the timer SQE returns nullptr, `_async_io` returns -1 immediately. The queued SQE will be submitted by the next `wait_and_fire_events` call, producing a CQE that dereferences the destroyed stack frame.

### 2. Lines 262-264 (cancel SQE allocation failure)

The original IO was already submitted. If `_get_sqe()` for the cancel SQE fails, `_async_io` returns -1 without waiting for the original IO CQE.

### 3. Line 268 (shutdown wakeup)

The cancel SQE is prepared (lines 265-267) but NOT submitted (no `try_submit()` call before `thread_sleep`). If `do_shutdown_usleep` caps the sleep at 10ms, both `io_ctx` AND `cancel_ctx` (also stack-allocated, line 265, `user_data` set at line 267) are destroyed — producing TWO dangling `user_data` pointers.

## Impact

`reap_events` dereferences the dangling pointers: writing `ctx->res` (line 377), branching on `ctx->is_canceller`/`ctx->is_event` (lines 385-386), calling `thread_interrupt(ctx->th_id)` with garbage (line 399).

## Possible fix directions

**Option A:** Loop in cancel path — after waking from `thread_sleep(-1)`, check `errno`; if woken by CQE processing (EOK), break; otherwise re-sleep. Risk: could hang during shutdown if IO is stuck (e.g., NFS).

**Option B:** Heap-allocate `ioCtx`/`cancel_ctx` so dangling pointers are harmless; free in `reap_events` after CQE processing.

**Option C:** For `_get_sqe()` failure paths (1 and 2), wait for outstanding CQE before returning instead of returning immediately.

Would appreciate maintainer input on which direction fits the project's design philosophy before submitting a fix.

jfeng18:
## Problem

Stack-allocated `ioCtx`/`timer_ctx` addresses are stored in SQE `user_data`, but `_async_io` could return early on three paths: link-timeout SQE allocation failure, cancel SQE allocation failure, and the cancel branch waking up before all CQEs are reaped. `reap_events` then dereferences dangling stack addresses.

## Fix

Add an `ioCtx::done` flag set by `reap_events` on the final CQE of a request; neutralize the already-queued first SQE with a NOP on allocation failure; wait on the done flags before returning from the cancel paths. The normal fast path is untouched.

## Tests

New `test-iouring-uaf`: sanity / interrupt storm / multi-thread storm / shutdown-with-inflight-io. Note: developed on a 4.9 kernel where io_uring is unavailable, so the new tests `GTEST_SKIP` locally; relying on CI (kernel >= 5.1) for runtime/ASan validation.

Fixes #1270

### DIFF
--- a/io/iouring-wrapper.cpp
+++ b/io/iouring-wrapper.cpp
@@ -54,6 +54,10 @@ public:
     int reset() override {
         fini();
         m_event_contexts.clear();
+        // The old ring died along with all its in-flight requests, so their
+        // CQEs can never arrive. Bump the generation to release the waiters
+        // in _async_io.
+        m_generation++;
         return init();
     }
 
@@ -224,6 +228,8 @@ public:
     }
 
     int32_t _async_io(io_uring_sqe* sqe, Timeout timeout, uint32_t ring_flags) {
+        auto* first_sqe = sqe;
+        auto gen = m_generation;
         sqe->flags |= (uint8_t) (ring_flags & 0xff);
         ioCtx io_ctx(false, false);
         io_uring_sqe_set_data(sqe, &io_ctx);
@@ -231,12 +237,20 @@ public:
         ioCtx timer_ctx(true, false);
         __kernel_timespec ts;
         auto usec = timeout.timeout_us();
-        if (usec < (uint64_t) std::numeric_limits<int64_t>::max()) {
+        bool has_timer = usec < (uint64_t) std::numeric_limits<int64_t>::max();
+        if (has_timer) {
             sqe->flags |= IOSQE_IO_LINK;
             ts = usec_to_timespec(usec);
             sqe = _get_sqe();
-            if (sqe == nullptr)
-                return -1;
+            if (sqe == nullptr) {
+                // The first SQE is already in the SQ ring and will be submitted
+                // sooner or later. Turn it into a harmless NOP without user data
+                // (prep_rw clears sqe->flags, including IOSQE_IO_LINK), so that
+                // its CQE won't reference the stack contexts after we return.
+                io_uring_prep_nop(first_sqe);
+                io_uring_sqe_set_data(first_sqe, nullptr);
+                return -1;      // errno was set to EBUSY by _get_sqe
+            }
             io_uring_prep_link_timeout(sqe, &ts, 0);
             io_uring_sqe_set_data(sqe, &timer_ctx);
         }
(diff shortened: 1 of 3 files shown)


lihuiba:
```cpp
    int32_t _async_io(io_uring_sqe* sqe, Timeout timeout, uint32_t ring_flags) {
        ...
        SCOPED_PAUSE_WORK_STEALING;
        photon::thread_sleep(-1);

        if (likely(errno == EOK)) {
            ...
        } else {
            // Interrupted by external user thread. Try to cancel the previous I/O
            ERRNO err_backup;
            sqe = _get_sqe();
            if (sqe == nullptr)
                return -1;
            ioCtx cancel_ctx(true, false);
            io_uring_prep_cancel(sqe, &io_ctx, 0);
            io_uring_sqe_set_data(sqe, &cancel_ctx);
            photon::thread_sleep(-1);
            errno = err_backup.no;
            return -1;
        }
    }
```
The first ```return``` above does lead to dangling stack addresses in rare cases. Can we simply go back the 1st ```photon::thread_sleep(-1)``` and sleep again?

If the second ```photon::thread_sleep(-1);``` is also interrupted by user, can we also sleep again and effectively make it non-interruptible?

jfeng18:
Exactly — that's what this patch does. On the interrupted path, both the "cancel SQE unavailable" case and the post-cancel wait loop on `thread_sleep(-1)`, ignoring further user interrupts, until `reap_events` has seen the final CQE of every stack context (the `ioCtx::done` flags), which effectively makes the cancel path non-interruptible. The done flags are used as the loop condition because the io / linked-timer / cancel CQEs may arrive in different reap batches, so a single wakeup is not a reliable signal; the `m_generation` check additionally keeps the loop from hanging when `reset()` (after fork) drops the old ring, whose CQEs can never arrive.

lihuiba:
how about this:
```cpp
    int32_t _async_io(io_uring_sqe* sqe, Timeout timeout, uint32_t ring_flags) {
        ...
        SCOPED_PAUSE_WORK_STEALING;
again:
        photon::thread_sleep(-1);

        if (likely(errno == EOK)) {
            ...
        } else {
            // Interrupted by external user thread. Try to cancel the previous I/O
            ERRNO err_backup;
            sqe = _get_sqe();
            if (sqe == nullptr)
                goto again;
            ioCtx cancel_ctx(true, false);
            io_uring_prep_cancel(sqe, &io_ctx, 0);
            io_uring_sqe_set_data(sqe, &cancel_ctx);
            do { photon::thread_sleep(-1);}
            while(errno != EOK);
            errno = err_backup.no;
            return -1;
        }
    }
```

jfeng18:
That sketch is the same "sleep until quiescent" idea, but `errno == EOK` is not a reliable signal that all related CQEs have been consumed. Two gaps: (1) if the I/O completes right as the cancel is submitted, its CQE fires the first `EOK` and `again2` exits immediately, while the cancel's own CQE (`-ENOENT`) is still in flight and will be reaped later, dereferencing the already-destroyed `cancel_ctx`; (2) in the normal cancel outcome, the original I/O's `-ECANCELED` and the linked timer's `-ECANCELED` are consumed silently in `reap_events` (`continue`, no interrupt), so if either lands in a later reap batch than the interrupting CQE, the loop exits while the ring still references the stack contexts. The per-request `done` flags exist to observe exactly those CQEs that never generate an interrupt. `m_generation` covers the remaining case where the CQEs can never arrive at all — `reset()` after fork drops the old ring; ForkTest.Iouring caught that in CI while this patch was developed.

lihuiba:
how about adding another sleep loop until io_ctx, timer_ctx and cancel_ctx are all done:
```cpp
    int32_t _async_io(io_uring_sqe* sqe, Timeout timeout, uint32_t ring_flags) {
        ...
        SCOPED_PAUSE_WORK_STEALING;
again:
        photon::thread_sleep(-1);

        if (likely(errno == EOK)) {
            ...
        } else {
            // Interrupted by external user thread. Try to cancel the previous I/O
            ERRNO err_backup;
            sqe = _get_sqe();
            if (sqe == nullptr)
                goto again;
            ioCtx cancel_ctx(true, false);
            io_uring_prep_cancel(sqe, &io_ctx, 0);
            io_uring_sqe_set_data(sqe, &cancel_ctx);
            do { photon::thread_usleep(-1);}
            while(errno != EOK);
            do { photon::thread_usleep(100);}
            while(io_ctx.done && timer_ctx.done && cancel_ctx.done);
            errno = err_backup.no;
            return -1;
        }
    }
```
This is a rare case, so a simple loop is acceptable.

Coldwings (approved):
LGTM

It seems like this is already a simple solution covers those cases.

PR #1570, opened 2026-07-25
fix(rpc): plug OOO engine stack context use-after-free

### LINKED ISSUE #1291: Bug: OOO RPC engine use-after-free when client times out during response body read
## Summary

The out-of-order RPC engine has a use-after-free vulnerability when a client times out while the receiver coroutine is processing the response body.

## Root Cause

In the OOO engine, the receiver extracts `targ` — a raw pointer to the caller's stack-local `OooArgs` struct. The receiver then calls `targ->do_collect()` which yields (performs I/O to read the response body). During this yield, the caller's timeout fires and the caller returns, destroying its stack frame. When the receiver resumes from `do_collect()`, it dereferences `targ` which now points to freed stack memory.

Sequence:
1. Caller creates `OooArgs` on stack, submits request, waits for response header
2. Receiver reads response header, extracts `targ` (raw pointer to caller's stack `OooArgs`)
3. Receiver calls `targ->do_collect()` which yields to read body from socket
4. While receiver is yielded, caller times out and returns — stack frame with `OooArgs` is destroyed
5. Receiver resumes, dereferences `targ` — use-after-free

## Possible Fix Directions

**Option A:** Heap-allocate `OooArgs` with `shared_ptr` shared between caller and receiver.

**Option B:** Implement a coordination protocol (flag or condition variable) preventing the caller from exiting while the receiver holds the pointer.

Would appreciate maintainer input on the preferred approach before submitting a fix.

jfeng18:
## Problem

`OooEngine::m_map` stores raw pointers to callers' stack-allocated `OutOfOrderContext`. After the receiver takes a pointer out of the map, it may yield in `do_collect()`; a timed-out caller can meanwhile return and destroy the stack context, leaving the receiver dereferencing freed stack memory.

## Fix

Add an `m_collecting` handshake: the receiver marks the context in flight; timed-out callers, the `do_issue()` failure path and the `wait_completion()` entry window wait for the hand-back instead of returning. The `!th` branch becomes a legal timed-out-caller state.

## Tests

New `test-ooo-timeout` with 4 real-path cases covering all three windows plus a concurrency stress; all pass locally under ASan. Replaces the old white-box test `error_thread_become_NULL` (`#define private public`) whose injected state is now a legal real path.

Fixes #1291

### DIFF
--- a/rpc/out-of-order-execution.cpp
+++ b/rpc/out-of-order-execution.cpp
@@ -33,6 +33,9 @@ namespace rpc {
         uint64_t m_issuing = 0;
         uint64_t m_tag = 0;
         bool m_running = true;
+        // The context currently being collected by the receiver (under m_mutex_r).
+        // A timed-out caller must not destroy its stack context until this is cleared.
+        OutOfOrderContext* m_collecting = nullptr;
 
         // rlock used as both reader lock and wait notifier.
         // add yield in lock will break the assuption that threads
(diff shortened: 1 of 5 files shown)


lihuiba:
> After the receiver takes a pointer out of the map, it may yield in `do_collect()`; a timed-out caller can meanwhile return and destroy the stack context, leaving the receiver dereferencing freed stack memory.

Maybe we can detect such case and let the timed-out caller simply wait for the former one to finish?

jfeng18:
Yes, that's exactly what this patch does: the receiver marks the context it has taken out of the map in `m_collecting`, and a timed-out caller that finds itself being collected waits in `wait_if_collecting()` until the hand-back completes, then returns the collected result instead of destroying its stack frame. The same handshake also covers the `do_issue()` failure path and the `wait_completion()` entry window.

Coldwings (approved):
LGTM

lihuiba:
how about using a sleep loop if such case occurs:
```cpp
int wait_completion(OutOfOrderContext& args) //recieving work
{
      ...
      auto ret = m_wait.wait(args.phaselock, args.timeout);
      // Check if collected
      if (args.phase == OooPhase::COLLECTED &&
          args.th == CURRENT) {
          return args.ret;
      }
      if (ret == -1) {
          // or just timed out
          {
              SCOPED_LOCK(m_mutex_map);
              m_map.erase(args.tag);
              m_cond_collected.notify_one();
          }
          // wait if args is in use
          while (args.in_use) thread_usleep(100);
          LOG_ERROR_RETURN(ETIMEDOUT, -1, "waiting for completion timeout");
      }
```

PR #1572, opened 2026-07-28
feat(http): per-client dialers with injectable dialer/resolver

Coldwings:
## Summary

Rework the HTTP client's dialer on three fronts: replace the process-wide `thread_local PooledDialer` (shared by all clients on a vCPU) with per-`(client, vCPU)` dialers owned by the `Client`; make dialing proxy-aware, so that a TLS origin behind a proxy is reached through a real `CONNECT` tunnel and the headers meant for the proxy are decided per hop rather than stored on the request; and share one DNS cache process-wide instead of one per vCPU. Adds `set_dialer()` / `set_resolver()` / `set_proxy_authenticator()` injection points; the existing API stays source-compatible.

## Motivation

1. **Hidden global sharing**: `dialer.init()` latches on first use, so the first client's TLS context / bind IPs silently win over every later client on that vCPU (the old comment in `client.h` even suggested "use separate std::threads for different TLS configurations" as a workaround).
2. **Destruction ordering**: `thread_local` destructors run at OS-thread exit, after `photon::fini()` has torn down the runtime — the existing `at_photon_fini` hook is a patch over the wrong lifetime ownership, and deleting a `Client` never releases its connection pools.
3. **DNS cache duplication**: with a per-vCPU resolver, every host is cold-resolved once per vCPU, i.e. O(vCPU x host) resolutions in multi-vCPU deployments.
4. **Proxying was outside the dialing abstraction**: a dialer that only learns where the origin is cannot tell whether the connection has to go to a proxy instead. Worse, the built-in dialer reached a TLS origin behind a proxy by handing the proxy an absolute-URI request over a plaintext connection — a combination that can never work, as the answer would have to arrive as TLS. Raised in review by @WaberZhuang.

### Per-client dialers (lifetime fix)

- `PooledDialer` becomes owned by a `(client, vCPU)` pair, created lazily on first use per vCPU, tracked by a per-vCPU `DialerRegistry`.
- A dialer is destroyed by whichever comes first: `~ClientImpl` (claiming it out of the registry; when deleted cross-vCPU, a helper photon thread is migrated to the owning vCPU), or the `photon::fini()` hook of its vCPU (covers leaked clients and the fini+init cycle of `pthread_atfork` handlers). Ownership is decided by who unlinks the dialer from the registry, so the two paths cannot race.

### Proxy-aware dialing, and CONNECT tunneling for TLS origins

- `IDialer` is now a single `dial(const DialTarget&, timeout)`. `DialTarget` carries the whole picture — the origin (host / port / secure), the unix socket path, the proxy and its credentials — and `via_proxy()` / `need_tunnel()` name the two ways a proxy is used, so an injected dialer can serve both.
- A plaintext origin behind a proxy is still served by forwarding an absolute-URI request. A TLS origin is now reached by opening a `CONNECT` tunnel and handshaking with the origin *inside* it (SNI = origin), which is the point of tunneling rather than forwarding: the proxy sees no plaintext.
- Correspondingly, a request line is written in absolute-URI form only for a plaintext origin; a tunneled request is in origin-form, exactly like a direct one.
- Tunnels are pooled under the proxy they were opened through, the origin they lead to, and the credentials that authenticated them. The last of the three matters: a `CONNECT` is authenticated once, for the tunnel, so a tunnel opened with one set of credentials must not be handed to a request carrying another — unlike a forwarded request, whose `Proxy-Authorization` travels per request and therefore permits sharing one connection to the proxy across credentials. Their outer legs are not pooled separately: a tunnel is reused as a whole, never its outer leg alone.
- A non-2xx `CONNECT` response fails with `ECONNREFUSED`, which the round-trip already maps to a fast retry instead of an exponential backoff. Any bytes trailing the response terminator are rejected as `EPROTO`, because the client is the one that speaks first inside a tunnel.
- This is a visible behaviour change for a TLS origin behind a proxy: it used to be forwarded as an absolute-URI request, which handed the proxy the whole request in cleartext and never gave the origin a TLS session of its own. A proxy that does not implement `CONNECT` therefore stops working where it previously appeared to. `ecosystem/test/test_oss_custom_headers.cpp` was relying on exactly that, and is adapted below.
- `Proxy-Authorization` is unified: the userinfo of a per-operation proxy URL is now honored (it used to be ignored — only client-level credentials worked). Where those credentials go is decided per hop, together with the rest of the headers meant for the proxy — see the next section.

### Proxy headers, and the pooling they imply

Raised in review, and in the discussion around #1466: richer proxy headers are needed, the tunnel pool has to account for them, and neither belongs on the `Operation`.

Which headers a proxy gets to read is a property of one hop, not of a request: a redirect can turn a forwarded request into a tunneled one, and only the former is read by the proxy — inside a tunnel it is the origin that reads them. A request object is therefore the wrong home for them, and holding `Proxy-Authorization` there is already wrong in both directions. The tunneling commit above guarded it by the scheme of the *first* URL, which a redirect defeats:

- `http://` origin, forwarded with the credentials (correctly), redirected to an `https://` one: the request keeps them while it travels **inside** the tunnel, so the origin is shown the proxy's credentials.
- `https://` origin, tunneled without them (correctly), redirected to an `http://` one: it is forwarded with no credentials at all, and is answered with 407.

Nothing in a set of headers records who its reader was meant to be, so a redirect cannot separate the proxy's headers from the caller's own; guessing by name would not cover the headers a caller adds itself. Removing them afterwards is not the answer either — headers are stored in insertion order, so erasing one costs about as much as rebuilding the message. So the headers meant for the proxy are never entered into the request at all:

- `Message::send_header()` takes an optional `extra` set of headers, written on the wire after its own without an entry being registered for any of them. A redirect of that message cannot carry them along, because there is nothing there to carry. The isolation is structural rather than a cleanup step that has to be remembered.
- New `ProxyAuthenticator` — `Delegate<int, const DialTarget&, ProxyAuth&>`, set by `Client::set_proxy_authenticator()` — is called once per dial that goes through a proxy, just before connecting. It returns the headers the proxy is to read and the pool key they imply, so they may be refreshed and may differ from one request to the next while belonging to no request. `DialTarget` carries both to the dialer, injected ones included.
- A caller-supplied component now takes part in a tunnel's pool key. A `CONNECT` authenticates a tunnel once, so only the identity that opened it may reuse it — but only the caller can tell which of its headers mean identity (a tenant, a route) and which are per-request noise (a trace id). Keying on all of them would open a tunnel per request; keying on none would let one identity ride another's tunnel. `Proxy-Authorization` stays keyed automatically, as before.
- The `CONNECT` is now built as a `Request` of its own instead of being concatenated as a string, so the authenticator's headers are deduplicated against `Host` and `Proxy-Authorization` rather than appended blindly. `make_request_line()` learns the authority-form target that `CONNECT` asks for. Its response is still parsed by hand, since it has to be read without consuming the bytes that follow it.

### One DNS cache per process

- The default resolver becomes a single process-wide cache, so each host is cold-resolved once rather than once per vCPU.
- The cache owns a `photon::Timer` and thus lives on the vCPU that created it. A borrow spans a single resolver call; the owning vCPU unpublishes the cache in its fini hook and drains the borrowers before freeing it — and, should draining ever time out, leaks it loudly rather than freeing it under a borrower. The next dial on a live vCPU publishes a fresh cache in its place.
- `set_resolver()` overrides this entirely, as before.

### Dependency injection (API addition, source-compatible)

- New public interface `IDialer` plus three non-virtual setters:
  - `Client::set_dialer(IDialer*)` — take over connection establishment / pooling entirely;
  - `Client::set_resolver(Resolver*)` — supply a DNS cache of one's own;
  - `Client::set_proxy_authenticator(ProxyAuthenticator)` — take over the headers sent to the proxy, and the pooling of the connections they authenticate.
- `new_http_client()` signature and default behavior are unchanged; existing code compiles and behaves as before, except that clients no longer accidentally share TLS contexts, and https-over-proxy now actually works — both bug fixes.

### Tests

- `http_client.per_client_dialer_lifecycle`: per-client dialer isolation; deleting one client tears down only its own pools.
- `http_client.dialer_injection` / `http_client.resolver_injection`: the new setters.
- `http_client.cross_vcpu_client_destruction`: client used on a worker vCPU, deleted from the main vCPU while the worker is alive (migrate path).
- `http_client.proxy_request_line`: absolute-URI for a plaintext origin, origin-form for a TLS one.
- `http_client.connect_tunnel`: a fake `CONNECT` proxy in front of a TLS origin — asserts the origin is reached, that the proxy was asked for the right authority, and that the tunnel is pooled (two requests, one `CONNECT`).
- `http_client.connect_tunnel_auth`: the credentials reach the proxy and do not leak to the origin.
- `http_client.connect_tunnel_refused`: a 407 from the proxy fails the call.
- `http_client.connect_tunnel_distinct_credentials`: two users of the same proxy get a tunnel each, rather than the second borrowing the first's authenticated one.
- `http_client.proxy_authenticator_headers_and_pool_key`: the authenticator's headers reach the proxy and not the origin; a header it declares to be noise does not multiply tunnels, while one it declares to be an identity gets a tunnel of its own.
- `http_client.proxy_headers_do_not_follow_a_redirect_into_a_tunnel`: a forwarded request, answered by the proxy with a redirect to a TLS origin, continues as a tunneled one — and the origin receives none of the headers the proxy read on the first hop.
- `http_client.proxy_headers_come_back_when_a_redirect_leaves_the_tunnel`: the mirror case — a tunneled request redirected to a plaintext origin is forwarded, so the headers the proxy reads must be there again.
- `OssCustomHeaderTest` (all six) now points the OSS client at a plaintext endpoint. The suite uses `opts.proxy` to route requests to a local capture handler and asserts on the headers and target the SDK built; with the default https endpoint the handler only ever saw the `CONNECT`. It models an origin behind a forwarding proxy — it answers 200 with a body — and tests header and URL-style construction, not proxy or TLS semantics, which the client's own tests above cover.
- `ReqHeaders.connect_is_in_authority_form`: a `CONNECT` names a host and a port, with neither scheme nor path.
- client_function_test 34/34, headers_test 7/7, client_tls_test 3/3, plus server_function_test and websocket_test. The three new `http_client` tests above were each checked by temporarily reverting the fix they cover, so that they are known to fail without it.

One existing assertion changed: `ReqHeaders.redirect` required a `https://` URL with `enable_proxy=true` to produce an absolute-URI request line. That is the unusable combination described in Motivation 4 (plaintext to the proxy, TLS expected back), so the test now covers the new rule instead — absolute-URI for a plaintext origin, origin-form for a TLS one.

## References

- #1466 (@WaberZhuang) implements `CONNECT` tunneling as well, and found the underlying bug first; which PR ends up carrying it is still to be aligned. Its design informed this one concretely — most usefully the pooling key that scopes a tunnel by credentials, which caught a real bug in the first version of this commit.
- Follow-up (separate PR planned): make `TCPSocketPool` fork-safe via `ResetHandle`, complementing the fini+init `pthread_atfork` pattern; split out of this PR to keep it focused.
- Follow-up (separate PR planned): RFC 7230-compliant `Transfer-Encoding` parsing (case-insensitive, tolerant of OWS, `chunked` required to be last) and response body decompression.

### DIFF
--- a/net/http/client.h
+++ b/net/http/client.h
@@ -55,11 +55,34 @@ struct DialTarget {
     uint16_t proxy_port = 0;
     bool proxy_secure = false;    // the proxy itself speaks TLS
     std::string_view proxy_auth;  // Proxy-Authorization value, may be empty
+    // Headers for the proxy itself to read, and the part of the connection
+    // pool key that they imply. Both are produced by the client's
+    // ProxyAuthenticator, once per dial. Never shown to the origin.
+    const HeadersBase* proxy_headers = nullptr;
+    std::string_view proxy_pool_key;
 
     bool via_proxy() const { return !proxy_host.empty(); }
     // a TLS origin behind a proxy is reached by tunneling with CONNECT
     bool need_tunnel() const { return via_proxy() && secure; }
 };
+
+// What a ProxyAuthenticator produces for one dial through a proxy.
+struct ProxyAuth {
+    CommonHeaders<4 * 1024 - 1> headers;   // put into the CONNECT, or into a forwarded request
+    estring pool_key;                      // connections differing here are never shared
+};
+
+// Called once per dial that goes through a proxy, just before connecting, so
+// that the headers may be refreshed and may differ from one request to the next.
+// Returns 0 on success, or a negative number to fail the request.
+//
+// Whatever identity the headers carry must be reflected in `pool_key`: a tunnel
+// is authenticated once, by the CONNECT that opened it, so it may only be reused
+// by the identity that opened it. Only the caller can tell which of its headers
+// mean identity (a tenant, a route) and which are mere per-request noise (a
+// trace id) -- keying on all of them would open a tunnel per request, and on
+// none of them would let one identity ride another's tunnel.
+using ProxyAuthenticator = Delegate<int, const DialTarget&, ProxyAuth&>;
 
 // Establishes socket streams for HTTP clients. The built-in implementation
 // maintains a connection pool per (client, vCPU), and one DNS cache shared by
@@ -191,6 +214,11 @@ public:
     virtual Headers* common_headers() = 0;
 
     void set_proxy(std::string_view proxy);
+    // Take over the headers sent to the proxy, and the pooling of the connections
+    // they authenticate. Not owned. See ProxyAuthenticator above.
+    void set_proxy_authenticator(ProxyAuthenticator authenticator) {
+        m_proxy_authenticator = authenticator;
+    }
     void set_user_agent(std::string_view user_agent) {
         m_user_agent = std::string(user_agent);
     }
@@ -239,6 +267,7 @@ public:
 protected:
     StoredURL m_proxy_url;
     std::string m_proxy_auth;
+    ProxyAuthenticator m_proxy_authenticator;
     std::string m_user_agent;
     uint64_t m_timeout = -1ULL;
     bool m_proxy = false;
--- a/net/http/client.h
+++ b/net/http/client.h
@@ -30,6 +30,7 @@ limitations under the License.
 namespace photon {
 namespace net {
 class TLSContext;
+class Resolver;
 namespace http {
 
 class IWebSocketStream;  // Forward declaration for websocket_connect
@@ -38,6 +39,18 @@ class ICookieJar : public Object {
 public:
     virtual int get_cookies_from_headers(std::string_view host, Message* message) = 0;
     virtual int set_cookies_to_headers(Request* request) = 0;
+};
+
+// Establishes socket streams for HTTP clients. The built-in implementation
+// maintains a per-vCPU connection pool and a DNS cache for each client.
+// A user-provided dialer is shared by all vCPUs the client runs on, so it
+// must be safe for concurrent use across vCPUs.
+class IDialer : public Object {
+public:
+    virtual ISocketStream* dial(std::string_view host, uint16_t port, bool secure,
+                                uint64_t timeout = -1ULL) = 0;
+    // dial to a Unix Domain Socket
+    virtual ISocketStream* dial(std::string_view uds_path, uint64_t timeout = -1ULL) = 0;
 };
 
 class Client : public Object {
@@ -183,6 +196,16 @@ public:
     void timeout_ms(uint64_t tmo) { timeout(tmo * 1000ULL); }
     void timeout_s(uint64_t tmo) { timeout(tmo * 1000ULL * 1000ULL); }
 
+    // Inject a dialer to take over connection establishment (and pooling, if
+    // any), replacing the built-in per-vCPU pooled dialer. Not owned; must
+    // outlive the client, and must be safe for concurrent use across vCPUs.
+    void set_dialer(IDialer* dialer) { m_dialer = dialer; }
+
+    // Inject a DNS resolver shared by the built-in dialers of this client
+    // (typically to share one DNS cache among clients / vCPUs). Not owned;
+    // must outlive the client. No effect on a dialer set by set_dialer().
+    void set_resolver(Resolver* resolver) { m_resolver = resolver; }
+
     virtual ISocketStream* native_connect(std::string_view host, uint16_t port,
                                           bool secure = false, uint64_t timeout = -1ULL) = 0;
 
@@ -202,11 +225,13 @@ protected:
     uint64_t m_timeout = -1ULL;
     bool m_proxy = false;
     std::vector<IPAddr> m_bind_ips;
+    IDialer* m_dialer = nullptr;
+    Resolver* m_resolver = nullptr;
 };
 
 // Create an HTTP client. Without cookie_jar, "Set-Cookies" headers are ignored.
-// Note: HTTP clients within the same std::thread share TLS config and connection pool.
-// Use separate std::threads for different TLS configurations.
+// Each client owns its connection pools (created lazily, one per vCPU used),
+// destroyed with the client, or at photon::fini() of the respective vCPU.
 Client* new_http_client(ICookieJar *cookie_jar = nullptr, TLSContext *tls_ctx = nullptr);
 
 ICookieJar* new_simple_cookie_jar();
--- a/net/http/client.h
+++ b/net/http/client.h
@@ -41,16 +41,33 @@ public:
     virtual int set_cookies_to_headers(Request* request) = 0;
 };
 
+// Where the connection of one HTTP request has to go. `host`/`port`/`secure`
+// always describe the origin server; when a proxy is in effect the connection
+// is made to the proxy instead, and the request is either forwarded in
+// absolute-URI form (a plaintext origin) or tunneled with CONNECT (a TLS
+// origin, which must be handshaked with the origin inside the tunnel).
+struct DialTarget {
+    std::string_view host;        // origin host, without port
+    uint16_t port = 0;            // origin port
+    bool secure = false;          // the origin speaks TLS
+    std::string_view uds_path;    // if set, connect here instead of TCP
+    std::string_view proxy_host;  // empty: connect to the origin directly
+    uint16_t proxy_port = 0;
+    bool proxy_secure = false;    // the proxy itself speaks TLS
+    std::string_view proxy_auth;  // Proxy-Authorization value, may be empty
+
+    bool via_proxy() const { return !proxy_host.empty(); }
+    // a TLS origin behind a proxy is reached by tunneling with CONNECT
+    bool need_tunnel() const { return via_proxy() && secure; }
+};
+
 // Establishes socket streams for HTTP clients. The built-in implementation
-// maintains a per-vCPU connection pool and a DNS cache for each client.
-// A user-provided dialer is shared by all vCPUs the client runs on, so it
-// must be safe for concurrent use across vCPUs.
+// maintains a connection pool per (client, vCPU), and one DNS cache shared by
+// the whole process. A user-provided dialer is shared by all vCPUs the client
+// runs on, so it must be safe for concurrent use across vCPUs.
 class IDialer : public Object {
 public:
-    virtual ISocketStream* dial(std::string_view host, uint16_t port, bool secure,
-                                uint64_t timeout = -1ULL) = 0;
-    // dial to a Unix Domain Socket
-    virtual ISocketStream* dial(std::string_view uds_path, uint64_t timeout = -1ULL) = 0;
+    virtual ISocketStream* dial(const DialTarget& target, uint64_t timeout = -1ULL) = 0;
 };
 
 class Client : public Object {
@@ -201,9 +218,10 @@ public:
     // outlive the client, and must be safe for concurrent use across vCPUs.
     void set_dialer(IDialer* dialer) { m_dialer = dialer; }
 
-    // Inject a DNS resolver shared by the built-in dialers of this client
-    // (typically to share one DNS cache among clients / vCPUs). Not owned;
-    // must outlive the client. No effect on a dialer set by set_dialer().
+    // Inject a DNS resolver, replacing the process-wide default one that the
+    // built-in dialers of this client would otherwise share. Not owned; must
+    // outlive the client, and must be safe for concurrent use across vCPUs.
+    // No effect on a dialer set by set_dialer().
     void set_resolver(Resolver* resolver) { m_resolver = resolver; }
 
     virtual ISocketStream* native_connect(std::string_view host, uint16_t port,
@@ -231,7 +249,8 @@ protected:
 
 // Create an HTTP client. Without cookie_jar, "Set-Cookies" headers are ignored.
 // Each client owns its connection pools (created lazily, one per vCPU used),
-// destroyed with the client, or at photon::fini() of the respective vCPU.
+// destroyed with the client, or at photon::fini() of the respective vCPU. The
+// DNS cache behind the pools is shared by the whole process.
 Client* new_http_client(ICookieJar *cookie_jar = nullptr, TLSContext *tls_ctx = nullptr);
 
 ICookieJar* new_simple_cookie_jar();
--- a/net/http/client.cpp
+++ b/net/http/client.cpp
@@ -39,6 +39,8 @@ static constexpr char USERAGENT[] = "PhotonLibOS_HTTP";
 // a CONNECT response is a status line and a few headers; anything longer than
 // this is not something we would be able to make sense of anyway
 static constexpr size_t kTunnelRespSize = 4 * 1024;
+// a CONNECT request is a request line, a Host, and whatever the authenticator adds
+static constexpr uint16_t kTunnelReqSize = 8 * 1024 - 1;
 
 class ClientImpl;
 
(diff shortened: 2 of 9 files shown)


net/http/client.h
```diff
@@ -40,6 +41,18 @@ class ICookieJar : public Object {
...
+// Establishes socket streams for HTTP clients. The built-in implementation
+// maintains a per-vCPU connection pool and a DNS cache for each client.
+// A user-provided dialer is shared by all vCPUs the client runs on, so it
+// must be safe for concurrent use across vCPUs.
+class IDialer : public Object {
+public:
+    virtual ISocketStream* dial(std::string_view host, uint16_t port, bool secure,
+                                uint64_t timeout = -1ULL) = 0;
+    // dial to a Unix Domain Socket
+    virtual ISocketStream* dial(std::string_view uds_path, uint64_t timeout = -1ULL) = 0;
```
WaberZhuang: `IDialer` should take proxy access into account.

WaberZhuang (changes requested):

net/http/client.h
Coldwings: Good catch — addressed in f5d5bc5, and the proxy headers / pooling follow-up in ab140a6 and 0457d89.

`IDialer` is now a single `dial(const DialTarget&, timeout)`, and `DialTarget` describes the whole connection rather than just the origin:

```cpp
struct DialTarget {
    std::string_view host;        // origin host, without port
    uint16_t port = 0;            // origin port
    bool secure = false;          // the origin speaks TLS
    std::string_view uds_path;    // if set, connect here instead of TCP
    std::string_view proxy_host;  // empty: connect to the origin directly
    uint16_t proxy_port = 0;
    bool proxy_secure = false;    // the proxy itself speaks TLS
    std::string_view proxy_auth;  // Proxy-Authorization value, may be empty
    // Headers for the proxy itself to read, and the part of the connection
    // pool key that they imply. Both are produced by the client's
    // ProxyAuthenticator, once per dial. Never shown to the origin.
    const HeadersBase* proxy_headers = nullptr;
    std::string_view proxy_pool_key;

    bool via_proxy() const { return !proxy_host.empty(); }
    // a TLS origin behind a proxy is reached by tunneling with CONNECT
    bool need_tunnel() const { return via_proxy() && secure; }
};
```

so an injected dialer can tell the two ways a proxy is used apart, instead of having proxying decided behind its back.

This overlaps #1466, which had already found and fixed the underlying bug: a TLS origin behind a proxy was reached by handing the proxy an absolute-URI `https://...` request over a plaintext connection, which can never work, as the answer would have to arrive as TLS. Happy to align on which PR ends up carrying it.

Either way, #1466's design fed into this one, and one detail earned its keep immediately — scoping a tunnel's pool key by the credentials that authenticated it. My first version keyed only on `(proxy, origin)`, and a test written after reading #1466 showed a second set of credentials silently reusing the first's authenticated tunnel. It is now keyed on proxy, origin and auth. The distinction is that a `CONNECT` is authenticated once, for the tunnel, whereas a forwarded request carries `Proxy-Authorization` per request and may therefore share one connection to the proxy across credentials. @lihuiba's `make_conditional_cat_list` suggestion on that same key in #1466 is what the key builder here uses.

## Proxy headers, and the pooling they imply

I first left these two open. They are needed now, so ab140a6 implements them — and doing so turned up something worse than a missing feature. @lihuiba's objection to a `proxy_header` on `Operation` points at the real defect: **which headers a proxy gets to read is a property of one hop, not of a request.** A redirect can turn a forwarded request into a tunneled one, and only the former is read by the proxy — inside a tunnel it is the origin that reads them. So a request object cannot hold them correctly, and `Proxy-Authorization` sitting there was already wrong in both directions. f5d5bc5 guarded it by the scheme of the *first* URL, which a redirect defeats:

- `http://` origin, forwarded with the credentials (correctly), redirected to an `https://` one: the request keeps them while it travels **inside** the tunnel, so the origin is shown the proxy's credentials.
- `https://` origin, tunneled without them (correctly), redirected to an `http://` one: it is forwarded with no credentials at all, and is answered with 407.

Both are pinned by tests, the leak end to end: the TLS origin echoes back that it received `Proxy-Authorization`.

Nothing in a set of headers records who its reader was meant to be, so a redirect has nothing to separate the proxy's headers from the caller's own by. Filtering by name would not do it either — it cannot cover the headers a caller adds itself. And removing them after the fact is expensive: headers are appended in insertion order, so erasing one costs about as much as rebuilding the message. Hence:

- `Message::send_header(stream, extra)` writes `extra` on the wire after its own headers **without registering an entry for any of them**. A redirect of that message cannot carry them along because there is nothing there to carry — the isolation is structural, not a cleanup step someone has to remember. The `CONNECT` gets them this way too, as a `Request` of its own.
- `ProxyAuthenticator` — `Delegate<int, const DialTarget&, ProxyAuth&>`, set by `Client::set_proxy_authenticator()` — is called once per dial through a proxy, just before connecting, so the headers may be refreshed and may differ from one request to the next while belonging to no request:

```cpp
struct ProxyAuth {
    CommonHeaders<4 * 1024 - 1> headers;   // put into the CONNECT, or into a forwarded request
    estring pool_key;                      // connections differing here are never shared
};
```

- The `pool_key` is the caller-supplied component of the tunnel's key that I suspected would be needed. A `CONNECT` authenticates a tunnel once, so only the identity that opened it may reuse it — but only the caller can tell which of its headers mean identity (a tenant, a route) and which are per-request noise (a trace id). Keying on all of them would open a tunnel per request; keying on none would let one identity ride another's tunnel. `Proxy-Authorization` stays keyed automatically, and the caller's contribution goes last in the key, where the colons it may contain cannot be mistaken for the separators of the fields before it.
- The outer leg of a tunnel is still not pooled on its own: a tunnel is reused as a whole.

Covered by `http_client.proxy_request_line`, `connect_tunnel`, `connect_tunnel_auth`, `connect_tunnel_refused`, `connect_tunnel_distinct_credentials`, `proxy_authenticator_headers_and_pool_key`, `proxy_headers_do_not_follow_a_redirect_into_a_tunnel`, `proxy_headers_come_back_when_a_redirect_leaves_the_tunnel` and `ReqHeaders.connect_is_in_authority_form`. The three new `http_client` ones were each checked by temporarily reverting the fix they cover, so they are known to fail without it. PTAL — @lihuiba too, since the shape of the authenticator and of the key is what you flagged.

Coldwings: Update: the two points I had left open above are implemented in ab140a6, and my reply is edited accordingly — headers for the proxy now come from a `ProxyAuthenticator` called once per dial, and the tunnel's pool key admits a caller-supplied component.

Implementing it turned up a defect worth calling out on its own: because `Proxy-Authorization` was stored on the request, a plaintext request forwarded with credentials and then **redirected to a TLS origin** kept them while travelling inside the tunnel — the origin was shown the proxy's credentials. f5d5bc5 guarded that by the scheme of the first URL, which a redirect defeats; the mirror case failed the other way, with a forwarded hop after a tunnel carrying no credentials at all and being answered 407. Both are now pinned by tests, the leak end to end.

The fix is structural rather than a filter: `Message::send_header(stream, extra)` writes the proxy's headers on the wire without registering an entry for any of them, so a redirect has nothing to carry over. Details and reasoning in the edited reply and in the PR description.

liulanzheng:
Two comments from GPT, for reference：
1，The registry unlink resolves ownership of the dialer, but I don't think it protects the lifetime of the owning vCPU.
After ~ClientImpl() unlinks the dialer, the owning vCPU's fini hook no longer sees it and may return. photon::fini() can then proceed to vcpu_fini(), which eventually destroys and frees the vcpu_t.
Meanwhile the client destructor still keeps the raw vcpu_base* and passes it to thread_migrate(). thread_migrate() doesn't appear to acquire any lifetime reference or verify that the target vCPU is still alive; do_thread_migrate() directly accesses vcpu->nthreads and vcpu->standbyq.
So there seems to be a window where the client wins ownership of the dialer, but the target vCPU is destroyed before the helper is migrated, resulting in a possible UAF rather than merely a failed migration.
I think claiming the dialer also needs to prevent the owning vCPU from completing fini until the destruction handoff has finished.

2，_users is shared across resolver generations. After at_photon_fini() unpublishes R1, another vCPU can publish R2 and increment the same _users. The owner of R1 will then wait for R2's borrowers as well, and may hit the drain timeout/leak path even though R1 itself has no borrowers left.
The borrow count should probably belong to the published resolver generation rather than SharedResolver globally.

Coldwings:
Thanks — both are real, and both are mine to fix. Confirmed against the code.

**1. Cross-vCPU dialer destruction can migrate onto a freed vCPU.** Exactly right: unlinking the dialer from the registry resolves who *deletes the dialer*, not the lifetime of its *vCPU*. Once `~ClientImpl` erases the dialer, that vCPU's `at_photon_fini` finds nothing to wait for and returns; `photon::fini()` then runs the fini hooks and only afterwards calls `vcpu_fini()`, which frees the `vcpu_t`. Meanwhile `destroy_dialer()` still holds the raw `vcpu_base*` and calls `thread_migrate(th, e.vcpu)`, and `do_thread_migrate()` writes `vcpu->nthreads++` / `move_to_standbyq_atomic()` with no liveness check on the target. So a claim that wins the race but migrates late lands on freed memory — a UAF, not a graceful failure. (The non-photon `!CURRENT` branch is fine: it disowns and lets the hook delete, no migrate.)

Fix: the claim must also pin the owning vCPU until the handoff finishes. Adding a `destroying` counter to the registry, incremented under the registry lock in the same critical section that erases the dialer and decremented after `destroy_dialer()` returns; `at_photon_fini` waits for it to reach zero after draining its list, before it lets `fini()` proceed to `vcpu_fini()`. No deadlock: the hook waits by yielding, so the migrated helper still gets scheduled on that vCPU, deletes the dialer, and releases the claim.

**2. `_users` is shared across resolver generations.** Also right. After `at_photon_fini` unpublishes R1, a `borrow()` on another vCPU publishes R2 and increments the *same* `_users`, so R1's drain loop waits for R2's borrowers too and can hit the timeout/leak path with no borrowers of its own left. The count has to belong to the generation, not to `SharedResolver`.

Fix: move the resolver and its borrow count into a per-generation node; `borrow()` counts on the current generation, `put()` decrements the generation it borrowed from, and the drain in `at_photon_fini` waits only on the generation it unpublished. A new generation started after the unpublish carries its own count and is unaffected.

Pushing both to this branch now; will follow up with the commits.

Coldwings:
Done, both on this branch:

- **1 (cross-vCPU UAF):** aa039d4 — a `destroying` counter on the registry, held from the erase that claims the dialer until `destroy_dialer()` returns, and waited on in `at_photon_fini` before `fini()` may proceed to `vcpu_fini()`.
- **2 (shared borrow count):** cfcac34 — the resolver and its borrow count now live in a per-generation node; the drain in `at_photon_fini` waits only on the generation it unpublished.

`client_function_test` and `client_tls_test` pass locally.

Coldwings:
CI was red on every job before b746c6e, with exactly one failing test out of 68 — `test-ecosystem` — so the breadth was one suite repeated across platforms, not many causes. Recording why it broke and why the fix is a test change rather than a client change.

`OssCustomHeaderTest` uses `opts.proxy` as an interception point: it points the OSS client at a local capture handler, then asserts on the headers and the request target the SDK built. Its endpoint carried no scheme, and a scheme-less endpoint means **https** (`m_is_http` defaults to false, and `http_or_s(false)` yields `https://`). A TLS origin behind a proxy is exactly the case this branch changes, so both halves of the test's premise fell away at once:

- the handler now only ever sees the `CONNECT`, whose headers are the proxy's hop, so none of the SDK headers are on it — which is the separation this PR is for;
- and it answers `200` with a 1-byte body, as an origin would, so the tunnel handshake rejects the body as bytes trailing the `CONNECT` response and the call fails with `EPROTO`.

The old behaviour it was relying on is the one being fixed: forwarding an https request in absolute-URI form handed the proxy the entire request in cleartext, `Authorization` included, and never gave the origin a TLS session of its own. Keeping the test green by not tunnelling would mean keeping that.

So the fix is to give the suite the hop it actually models. `capture_handler` plays an origin behind a *forwarding* proxy — it replies with a body — and the suite tests header and URL-style construction, not proxy or TLS semantics. With a plaintext endpoint the request arrives as an absolute-URI forward, which is what the handler can observe, and every existing assertion holds unchanged: the target is now `http://test-bucket.oss-test.example.com/test-obj` instead of the `https://` one it used to see, so `find("/test-obj")` still matches and `find("/test-bucket/")` still misses for virtual-hosted style. No assertion was weakened or removed. Tunnelling stays covered by the http client's own tests.

Worth flagging for anyone using a proxy with an OSS endpoint: since a scheme-less endpoint is https, that configuration now opens a `CONNECT` tunnel. Real proxies implement it; a plaintext-only interception point does not, and will now fail instead of quietly seeing everything. The PR body notes this under the tunnelling section.

`OssCustomHeaderTest` 6/6 and `test-ecosystem` 16/16 pass locally. The one test I excluded is `redis.cmd`, which connects to `127.0.0.1:6379` with no guard and segfaults when no Redis is running — pre-existing and unrelated, untouched here.

PR #1573, opened 2026-07-28
fix(alog): prevent child deadlock on rotation lock across fork()

Coldwings:
## Problem

The log rotation path in `LogOutputFile::write()` guards rotation with a function-local `static std::mutex`. POSIX `fork()` clones only the calling thread, so if a fork happens while another thread is holding this mutex during rotation, the child inherits the mutex in a permanently locked state — the owner thread does not exist in the child. The next time the child's log size exceeds the rotate limit, it deadlocks forever.

The existing `pthread_atfork(nullptr, nullptr, &reset_all_handle)` in `photon.cpp` only resets fds / event engines and does not cover this lock.

## Fix

Standard three-phase `pthread_atfork` protection:

- Hoist the mutex from function-local static to file scope so the atfork callbacks can reference it.
- Register via `__attribute__((constructor))`:
  - *prepare*: acquire the lock before fork, waiting out any in-flight rotation, so at the instant of fork the lock is held by the forking thread itself;
  - *parent* / *child*: release it after fork. Since the holder is the forking thread, unlocking on both sides is well-defined.

As a side benefit, the prepare phase also guarantees the child never observes a half-done rotation (rename/reopen in progress).

## Test

Added `ALog.fork_during_rotation` regression test: a background thread keeps writing to force frequent rotations while the main thread forks 8 times; each child writes ~2MB to force its own rotation. If the lock were inherited locked, the child would deadlock, which is caught by a 10s `waitpid` timeout.

All 24 tests in `test-alog` pass.

### DIFF
--- a/common/alog.cpp
+++ b/common/alog.cpp
@@ -25,6 +25,7 @@ limitations under the License.
 #include <mutex>
 #include <condition_variable>
 #include <unistd.h>
+#include <pthread.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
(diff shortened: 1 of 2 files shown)


pakkah:
LGTM

lihuiba (approved):

PR #1598, opened 2026-08-04
fix(thread): more tests on work-stealing

lihuiba:
Fix several races/deadlocks in work-stealing, found by the new tests:

- ABBA deadlock: steal scans held the victim's runq/standbyq lock and blocked on th->lock, while die()/thread_interrupt() take th->lock before those locks. Use try_lock() in ws_scan_q()/ws_scan_standbyq() and skip busy threads; the idler rescans shortly.
- vcpu_list_rwlock could yield in the idler context, where the runq may contain nothing else to run. Remove it and iterate pvcpu under the spinlock vcpu_list_lock only.
- Threads still resident in a sleepq (idx != -1), e.g. interrupted sleepers sitting in both the owner's sleepq and standbyq, must only be resumed by their own vCPU. Add thread::stealable() and use it in both scan paths.
- vcpu was published to pvcpu in the ctor, before idle_worker existed. Publish via go_online() only after vcpu_init() completes, handle thread_create() failure, and extract the common teardown into vcpu_destroy().
- Initialize idle_worker to nullptr.

Add 14 tests covering: steal from runq (single/multiple stealers), passive/active/thread flags gating, thread_pause_work_stealing() and SCOPED_PAUSE_WORK_STEALING, sleeping/interrupted sleepers not stolen, standbyq migration, mixed stealable/unstealable, stolen thread full life-cycle, multi-stealer stress, and vCPU churn under active stealing.

Validated: full test-thread 69/69, --vcpus=4, and 50x repeated runs of the WorkStealing suite with no failure or hang.

### DIFF
--- a/thread/test/test.cpp
+++ b/thread/test/test.cpp
@@ -21,6 +21,10 @@ limitations under the License.
 #include <stdlib.h>
 #include <queue>
 #include <algorithm>
+#include <atomic>
+#include <functional>
+#include <thread>
+#include <unistd.h>
 #include <sys/time.h>
 #include <gflags/gflags.h>
 #include "../../test/gtest.h"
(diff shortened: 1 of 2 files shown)


Coldwings (approved):
LGTM

PR #1599, opened 2026-08-04
feat(thread): add global pooled coroutine stack allocator

Coldwings:
### What

Adds a process-wide pooled coroutine stack allocator (`thread/stack-pool.cpp`), selectable via `PhotonOptions::use_global_pooled_stack_allocator` or `use_global_pooled_stack_allocator()`. It is a three-layer magazine/depot design (per-vcpu magazine → global per-size-class depot → per-stack `mmap`), exposing `global_pooled_stack_trim()` and `global_pooled_stack_stats()`. The existing thread-local `pooled_stack_allocator` is left untouched.

### Why

The existing `pooled_stack_allocator` keeps a separate pool per vcpu (OS thread). When photon threads migrate across vcpus, a stack allocated on one vcpu and freed on another accumulates on the destination, so the pools drift out of balance and process-wide stack memory is hard to bound. The global pool shares one depot across all vcpus, so cross-vcpu free is balanced and the idle cache is bounded process-wide (`max_pooled_bytes` / `max_pending_bytes` / `max_cold_bytes`). In a migrate-heavy benchmark the vcpu-local pool's virtual size grows to tens of GB while the global pool stays bounded and delivers higher throughput.

### How

- Backing store is a per-stack `mmap` with `MAP_NORESERVE`; each block is `[guard page(s)][stack][meta page]` (2 VMAs/block). A per-block metadata page enables O(1) double-free / corruption detection with a per-process magic, and a range gate rejects foreign pointers before dereferencing.
- The free path runs in the die-defer context: it is lock-free and syscall-free (per-vcpu magazine push), never allocates, and never `mmap`/`munmap`s. Overflow spills to a depot pending list; a bounded back-pressure valve `madvise`s excess into a cold list (still mapped, RSS freed) and only `munmap`s beyond the cold bound.
- There is no preemptive total cap: a stack allocation failure is catastrophic for a coroutine, so on `mmap` refusal the allocator returns its entire idle cache to the OS and retries, giving up only when the OS still cannot satisfy the request. `munmap` cost is amortized to the alloc slow path and to explicit `trim()`.
- Concurrency: the depot uses `photon::spinlock` (pure spin, safe in die-defer); size-class registration publishes with release/acquire; per-vcpu hit/miss counters are atomic; magazine allocation is kept out of the spinlock via a pre-allocated per-vcpu spare; first-time init is serialized with `std::call_once`; `pthread_atfork` handlers keep the depot consistent across `fork`.

### Testing

- `test-global-stack-pool` (registered with CTest): exact-size reuse, multi-size-class + passthrough, reclaim-on-`mmap`-failure (under `RLIMIT_AS` in a child), double-free detection, guard-page `SIGSEGV`, cold-pool reuse, and cross-vcpu boundedness on a `WorkPool`. Passes under `MinSizeRel` and `-D PHOTON_BUILD_WITH_ASAN=ON`.
- `tsan-stack-pool` (manual): multi-thread stress (workers + concurrent `stats()`/`trim()`) reports zero data races under ThreadSanitizer.
- `bench-stack-allocator` (manual): malloc / pooled / global three-way comparison across create-join, burst, migrate-heavy and steady-churn patterns.
- Existing `test-thread`, `test-pool`, `test-pooled-stack-allocator` continue to pass.


photon.h
```diff
@@ -62,6 +62,7 @@ struct PhotonOptions {
     uint32_t iouring_sq_thread_cpu;
     uint32_t iouring_sq_thread_idle_ms = 1000;     // by default polls for 1s
     bool use_pooled_stack_allocator = false;
```
lihuiba: how about ```unsigned char use_pooled_stack_allocator = 0/1/2/3/4/...;```

lihuiba (approved):

thread/stack-pool.cpp
```diff
@@ -0,0 +1,788 @@
...
+        // Reset the locks (their prior holders no longer exist in the child),
+        // then absorb every vcpu's magazines so no cached block is lost, and
+        // drop the PerVCPU nodes of threads that no longer exist (only the
+        // forking thread survives). Malloc-free except for delete of dead nodes.
+        uint32_t nc = n_classes.load(std::memory_order_relaxed);
+        new (&reg_lock) spinlock();
+        for (uint32_t i = 0; i < nc; i++) new (&classes[i].lock) spinlock();
+        PerVCPU* self = (PerVCPU*)pthread_getspecific(key);
+        PerVCPU* keep = nullptr;
+        for (PerVCPU* v = vcpu_list; v; ) {
```
lihuiba: multi-vcpu is *NOT* compatible with fork. we should forbid this combination.

thread/stack-allocator.h
```diff
@@ -47,4 +47,73 @@ inline void use_pooled_stack_allocator() {
...
+
+void* global_pooled_stack_alloc(void*, size_t stack_size);
+void global_pooled_stack_dealloc(void*, void* stack_ptr, size_t stack_size);
+
+// Select the global pooled stack allocator and capture its options. Must be
+// called before photon threads are created, or while the pool is otherwise
+// quiescent (no concurrent allocation): the options are read locklessly on the
+// hot path, so reconfiguring under concurrent load races. A later quiescent
+// call updates the options in place and keeps the existing pools.
+int use_global_pooled_stack_allocator(const GlobalStackPoolOptions& options = {});
```
lihuiba, line 168: how about upgrading the interface of stack allocator to proxy?

thread/stack-allocator.h
```diff
@@ -47,4 +47,73 @@ inline void use_pooled_stack_allocator() {
...
+    // between photon threads). 0 disables wiping.
+    uint32_t wipe_bytes           = 0;
+    // Keep pooled blocks PROT_NONE while idle so use-after-free faults; costs
+    // two extra syscalls per reuse. For debugging only.
+    bool     paranoid             = false;
+    // MADV_NOHUGEPAGE on the stack region, matching the other allocators.
+    bool     no_huge_page         = true;
+};
+
+struct GlobalStackPoolStats {
```
lihuiba: consider defining a base ```StackPoolStats``` that includes common fields

photon.h
Coldwings: Done in 074a907 — adopted exactly this: the field is now `uint8_t use_pooled_stack_allocator` with named values (`STACK_ALLOCATOR_DEFAULT` / `_POOLED` / `_GLOBAL_POOLED`), and the separate `use_global_pooled_stack_allocator` bool is gone, so the allocators are mutually exclusive by construction.

Keeping it 1 byte turns out to matter for ABI, which was the real reason to prefer `unsigned char` over `int`. Measured layouts of `PhotonOptions`:

| variant | sizeof | off(use_pooled_stack_allocator) | off(bypass_threadpool) |
|---|---|---|---|
| original `bool` | 16 | 12 | 13 |
| `int` | 20 | 12 | 16 |
| `uint8_t` | 16 | 12 | 13 |

With `int`, `bypass_threadpool` shifts 13 -> 16 and the struct grows, i.e. already-compiled callers would read the wrong field. With `uint8_t` the layout is byte-identical to before this PR, so the change is ABI compatible. `bool true` stores byte value 1 == `STACK_ALLOCATOR_POOLED`, so existing call sites (`thread/st.cpp`, `common/test/perf_rcuptr.cpp`, `examples/perf/io-perf.cpp`, `thread/test/test-pooled-stack-allocator.cpp`) keep their original meaning and are left untouched. A `static_assert` pins the width so this cannot silently regress.

thread/stack-pool.cpp
Coldwings: Agreed, and simplified in 1c05afa. You are right that this cannot be repaired at the allocator level: the child inherits the memory of every vcpu but only the calling thread, so photon threads on the other vcpus are gone while their state remains.

The child handler no longer tries to absorb the vanished threads' magazines or delete their `PerVCPU` nodes — that was working around an unsupported combination while adding real risk (walking and relinking their blocks). It now only releases the locks taken in `prepare`, which is all that is actually needed so the child never inherits a locked spinlock. Blocks cached by threads that do not exist in the child are simply leaked there, and the header now states that fork + multiple vcpus is unsupported.

45 lines down to about 10. If you would also like an explicit hard prohibition (e.g. failing or warning loudly when a fork is detected with more than one vcpu), I think that belongs at the photon level rather than in the allocator, since the allocator cannot reliably tell what the caller is about to do — happy to follow up wherever you prefer.

thread/stack-allocator.h
Coldwings: Done in 22d5bd4, using `Delegates` (both interfaces coexist, the public API is additive only).

```cpp
using StackAllocator = Delegates<
    DFeature::Feature<DStackAlloc,   void*(size_t)>,
    DFeature::Feature<DStackDealloc, void(void*, size_t)>,
    DFeature::Feature<DStackTrim,    size_t(size_t)>,
    DFeature::Feature<DStackStats,   StackPoolStats()>>;
```

Any object exposing `alloc`/`dealloc` (and optionally `trim`/`stats`) can be installed with `set_photon_thread_stack_allocator(obj)`, and `get_photon_thread_stack_allocator()` exposes the installed one so `trim()`/`stats()` work generically on whichever allocator is in use. `GlobalStackPool` already had methods with those names, so it now installs itself as an object.

The main win is the one you implied: binding an object binds every method at once. The old interface took two independent delegates *each with a default argument*, so `set_photon_thread_stack_allocator(my_alloc)` silently paired a custom allocator with the default deallocator — stacks freed by an allocator that did not create them. Installing an allocator without both is now rejected with EINVAL. `thread.cpp` holds a single facade internally instead of two delegates; the hot path cost is unchanged (still one indirect call, and one less hop for the global pool).

One constraint worth recording: `Delegates` binds a single `__objptr`, so all methods must belong to the same object. The legacy two-delegate entry point therefore fails with EINVAL if the two delegates carry different non-null objects (they never do in practice — every allocator shipped with photon uses free functions).

Two side effects of using it here: `delegates.h` had no include guard, and the single-signature `DUtils::bind_func` overloads for a `void*`/`nullptr` object referenced an undeclared `DC` (they only fail once instantiated, which no caller did). Both fixed in ce5b330.

thread/stack-allocator.h
Coldwings: Done in 22d5bd4, together with the facade since the two go hand in hand.

```cpp
struct StackPoolStats {          // common to any pooling stack allocator
    size_t   pooled_bytes = 0;   // idle memory kept for reuse
    size_t   live_bytes = 0;     // handed out and in use
    uint64_t hits = 0;
    uint64_t misses = 0;
};
struct GlobalStackPoolStats : StackPoolStats {
    size_t   mapped_bytes = 0, pending_bytes = 0, cold_bytes = 0;
    uint64_t os_maps = 0, os_unmaps = 0, corruptions = 0;
};
```

The split is by what is inherently generic versus mmap-specific: mapped/pending/cold bytes and the mmap/munmap/corruption counters only mean something for this pool. The facade's `stats()` feature returns the base, so generic code and user-written allocators share one type, while `global_pooled_stack_stats()` still returns the full struct.

This is what makes the base worth having now rather than being speculative: `StackAllocator::stats()` gives it a second consumer immediately, and it pins down the shape users should report from their own allocators.

lihuiba (approved):

PR #1610, opened 2026-08-12
fix(thread): make context-switch asm symbols LTO-safe

NewbieOrange:
The symbols defined in the file-scope asm() blocks had local binding, so
LTO could not resolve the references from inline asm call sites placed in
other partitions. _photon_thread_die was additionally dropped by LTO once
its only C++ caller was inlined.
Emit .globl plus .hidden for the asm-defined symbols, un-static
_asan_start, and mark _photon_thread_die used. No ABI change.

### DIFF
--- a/thread/thread.cpp
+++ b/thread/thread.cpp
@@ -69,13 +69,18 @@ limitations under the License.
 // Define assembly section header for clang and gcc
 #if defined(__APPLE__)
 #define DEF_ASM_FUNC(name) ".text\n" \
+                           ".globl "#name"\n" \
+                           ".private_extern "#name"\n" \
                            #name": "
 #elif defined(_WIN64)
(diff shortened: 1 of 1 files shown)


NewbieOrange:
successfully built with LTO on linux/amd64 and passed stress test (w/o work-stealing).

NewbieOrange:
The failed action seems like a flaky test

lihuiba (approved):

lihuiba:
how much does LTO improves?
do you need to backport it to a release branch?

NewbieOrange:
> how much does LTO improves? 

about ~10% extra qps in our stress test with 4KB requests, but that's with the whole service being LTO'd. we haven't measured the difference of just photon with LTO.

> do you need to backport it to a release branch?

no, we are using internal branch based on release/0.9

PR #1620, opened 2026-08-19
[Backport][0.9 to 0.8] | | rpc: fix OOO engine stack context UAF (#1291) (#1570) (#1614)

photonlibos:
> [Backport][main to 0.9] | rpc: fix OOO engine stack context UAF (#1291) (#1570) (#1614)

* rpc: fix OOO engine stack context UAF (#1291) (#1570)

A timed-out caller could return and destroy its stack
OutOfOrderContext while the receiver was still collecting it
after a yield in do_collect(). Add m_collecting handshake: the
receiver marks the context in flight; timed-out callers, the
do_issue() failure path and the wait_completion() entry window
wait for the hand-back. notify_all() wakes both waiting callers
and shutdown(). The !th branch becomes a legal timed-out state
(continue); error_thread_become_NULL replaced by real-path tests.

* chore: drop files not present in 0.9

* Refactor CMakeLists to use add_executable for tests

* add rpc/test/test-ooo-timeout.cpp

---------

Co-authored-by: Jiangtian Feng <fengjiangtian.fjt@alibaba-inc.com>
Co-authored-by: Coldwings <11592728+Coldwings@users.noreply.github.com>

## Conflicts

- 0aecf4e9eab9ab564d1b2ba2bab909ef7e755ceb: rpc/out-of-order-execution.cpp

## Dropped new files

The following new files introduced by cherry-pick were dropped because they don't exist in `release/0.8`:

- rpc/test/test-ooo-timeout.cpp


Coldwings:
Closing this backport: the conflicts here are not resolvable in isolation, and the fix needs a separate, dedicated effort for 0.8 and earlier.

## Why it cannot be resolved as a conflict resolution

`release/0.8` still has the older OOO engine, without the phase state machine: no `OooPhase`, `phase`, `phaselock`, `timeout`, `m_mutex_map`, `m_wait`, `issue_wait` or `result_collected`. The cherry-picked code depends on all of them.

This is not only about the three conflicted hunks. `wait_if_collecting()` applied **cleanly**, outside any conflict block, and still fails to compile here (`OutOfOrderContext has no member named 'phaselock'`). Deleting every conflict marker would still leave the branch unbuildable, with 84 compile errors.

The missing prerequisite is 44abd5c "fix RPC stop by timeout during calling (#552)", which introduces the phase machine across 12 files, including `rpc.h` / `rpc.cpp` and the installed public header `include/photon/rpc/out-of-order-execution.h`, so it carries ABI implications. Pulling that into a patch release just to enable this fix is not the right trade-off.

## Does 0.8 have the bug?

Partly, so this is not a false alarm — recorded here for whoever picks it up:

- **Affected: the `do_issue()` failure path.** In `do_call`, `OooArgs args` is a stack variable, and `issue_operation()` only does `m_map.erase(args.tag); return -1;` on failure, with no wait. Meanwhile the receiver holds a raw pointer to that context across a yield: `it->second->ret = targ->do_collect(targ)` (`do_recv_body` -> `readv`). `do_issue` is `do_send` -> `writev`, which also yields and can fail. The two sides hold different locks (`m_mutex_w` vs `m_mutex_r`), so they can truly race: once the caller returns, the receiver writes through a dangling stack pointer and a stale iterator. This is exactly the path #1291 addressed with `wait_if_collecting()` in `issue_operation`.
- **Not affected: the timeout path.** 0.8's `wait_completion()` has no timed wait, so there is no "caller times out, deregisters and returns early" path. That one arrived with #552 and is the most easily triggered on 0.9 / main. 0.8 checks the timeout only at the top of `do_send`, before any yield.
- A secondary variant exists via `scoped_lock lock(m_mutex_r, 1)`, which by its own comment allows lock-free entry when interrupted; it depends on an external interrupt and is rarer.

Note this is the result of code review, not of a runtime reproduction.

## Suggested follow-up

A minimal, 0.8-native fix is feasible: apply the `m_collecting` + hand-back-wait idea to the `do_issue()` failure path only, adapted to 0.8's `collected` / `th` design, without introducing the phase or timeout machinery. That keeps the public header and the ABI untouched, unlike backporting #552.

PR #1624, opened 2026-08-21
iouring: reap events immediately after eager submit

lihuiba:


Coldwings (approved):
LGTM

Coldwings:
The failing `debug-build-from-source` workflow failed because `pagure.io` now in archiving state. That should be ignored, and will be fix in a standalone PR.

PR #1632, opened 2026-08-26
[Backport][0.9 to 0.8] | ci: gate PR builds on ready-for-review, split backport control into need-backport (#1630) (#1631)

photonlibos:
> ci: gate PR builds on ready-for-review, split backport control into need-backport (#1630) (#1631)

red run.

## Conflicts

- 3bc0d38f39dbda57c312df811467916d126447e7: .github/workflows/ci.linux.arm.yml,.github/workflows/ci.linux.x86_64.yml,.github/workflows/ci.macos.arm.yml

### DIFF
--- a/.github/workflows/auto-pr-backport.yml
+++ b/.github/workflows/auto-pr-backport.yml
@@ -70,10 +70,11 @@ jobs:
           const isManual = '${{ github.event_name }}' === 'workflow_dispatch';
           const commit = '${{ steps.inputs.outputs.COMMIT }}';
           if (isManual) {
-            core.info('Manual trigger: bypassing source PR / bugfix label checks');
+            core.info('Manual trigger: bypassing source PR / need-backport label checks');
             core.setOutput('SKIP', 'false');
             core.setOutput('SOURCE_HEAD_REF', '');
-            core.setOutput('HAS_BUGFIX', 'true');
+            core.setOutput('SOURCE_LABELS', '[]');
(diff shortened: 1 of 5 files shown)


Coldwings:
Conflicts resolved in 7efbf14.

`release/0.8` still runs the pre-matrix CI layout — one job per toolchain on `ubuntu-latest`
inside the `photon-ut-base` container, plus two separate macOS workflows — so the cherry-pick
conflicted everywhere `main` had reshaped a job definition. Only the semantic change is taken:

- every job now gates on `!github.event.pull_request.draft`;
- the workflows trigger on `ready_for_review` instead of `unlabeled`;
- the `-noop` concurrency group is gone.

Runners, containers and job names are left exactly as they are on this branch, and the jobs
`main` added later (`debug-build-from-source`, `RocksDB`) are **not** imported — a backport of a
gating change should not quietly change what gets built.

Two spots the cherry-pick could not reach were converted by hand, since without them the label
trigger would survive on this branch and keep producing all-skipped runs:

- the `gcc921` / `gcc1031` / `gcc1121` / `gcc1211` jobs in `ci.linux.x86_64.yml`, which have no
  counterpart in `main`'s matrix;
- `ci.macos.x86_64.yml`, which `main` has since folded into a matrix inside `ci.macos.yml`.

The resulting diff against `release/0.8` is 4 files / 58 insertions / 33 deletions, all of it the
gating change plus the `auto-pr-backport.yml` label logic that came over cleanly.

Coldwings (approved):

PR #1637, opened 2026-08-27
[Backport][0.8 to 0.7] | | ci: gate PR builds on ready-for-review, split backport control into need-backport (#1630) (#1631) (#1632)

photonlibos:
> [Backport][0.9 to 0.8] | ci: gate PR builds on ready-for-review, split backport control into need-backport (#1630) (#1631) (#1632)

* ci: gate PR builds on ready-for-review, split backport control into need-backport (#1630) (#1631)

red run.

* ci: resolve backport conflicts, keep 0.8 job topology

release/0.8 still runs the pre-matrix CI layout — one job per toolchain on
ubuntu-latest inside the photon-ut-base container, and two separate macOS
workflows — so the cherry-pick conflicted everywhere main had reshaped a job
definition.

Only the semantic change is taken: every job gates on the PR's draft state and
the workflows trigger on ready_for_review instead of unlabeled. Runners,
containers and job names stay as they are on 0.8, and the jobs main added later
(debug-build-from-source, RocksDB) are not imported.

Two spots the cherry-pick could not reach are converted by hand: the four
gcc9xx/gcc1xxx jobs in ci.linux.x86_64.yml, which have no counterpart in main's
matrix, and ci.macos.x86_64.yml, which main has since folded into a matrix
inside ci.macos.yml. Without them the label trigger would survive on this
branch and keep producing all-skipped runs.

---------

## Conflicts

- 786fffa5e4015ccc37ed2bbc3a128ce87ac8d96e: .github/workflows/ci.linux.x86_64.yml


Coldwings:
Resolved the conflict and marked this ready for review.

**What conflicted.** One region of `.github/workflows/ci.linux.x86_64.yml`, as a modify/delete: `release/0.7` carries a smaller x86_64 matrix than `0.8` — `gcc850`, `gcc921`, `gcc1031` only, none of which run the `epoll_ng` pass, and no `fstack` job. The cherry-pick's merge base still had all of those, so git saw "this branch deleted the region, the incoming commit modified it" and presented the whole ~95-line block for a decision.

**How it was resolved.** Took this branch's side. The three jobs `0.7` actually has are gated on draft state and the workflow triggers on `ready_for_review`; what gets built and tested is untouched. A change to *when* CI runs shouldn't change *what* CI covers, so `gcc1121`, `gcc1211`, `fstack` and the `epoll_ng` step were not imported.

Net diff against `release/0.7` is purely the gating change across five files — trigger types, the `-noop` concurrency workaround removed, and one `if:` per job. `auto-pr-backport.yml` is byte-identical to the copy already merged on `release/0.8`. All ten workflow files parse as YAML, and the job inventory of the resolved file matches `release/0.7` exactly.

**One thing left alone, deliberately.** `ci.macos.x86.yml` exists only on this branch and predates the label guard: no `unlabeled` trigger, no job-level `if:`. It therefore never produced the all-skipped phantom runs this change removes, and it's the one workflow here that will still spend a runner on a draft backport. Adding a draft gate to it is a new decision rather than part of this backport, so it's flagged instead of changed.

Note that CI here was started by removing `needs-manual-merge`, not by the undraft: this PR's triggers come from `release/0.7`, which still has the old label-based workflows. That resolves itself once this merges.

Coldwings (approved):

PR #1638, opened 2026-08-31
[Backport][0.7 to 0.6] | | | ci: gate PR builds on ready-for-review, split backport control into need-backport (#1630) (#1631) (#1632) (#1637)

photonlibos:
> [Backport][0.8 to 0.7] | | ci: gate PR builds on ready-for-review, split backport control into need-backport (#1630) (#1631) (#1632) (#1637)

* [Backport][0.9 to 0.8] | ci: gate PR builds on ready-for-review, split backport control into need-backport (#1630) (#1631) (#1632)

* ci: gate PR builds on ready-for-review, split backport control into need-backport (#1630) (#1631)

red run.

* ci: resolve backport conflicts, keep 0.8 job topology

release/0.8 still runs the pre-matrix CI layout — one job per toolchain on
ubuntu-latest inside the photon-ut-base container, and two separate macOS
workflows — so the cherry-pick conflicted everywhere main had reshaped a job
definition.

Only the semantic change is taken: every job gates on the PR's draft state and
the workflows trigger on ready_for_review instead of unlabeled. Runners,
containers and job names stay as they are on 0.8, and the jobs main added later
(debug-build-from-source, RocksDB) are not imported.

Two spots the cherry-pick could not reach are converted by hand: the four
gcc9xx/gcc1xxx jobs in ci.linux.x86_64.yml, which have no counterpart in main's
matrix, and ci.macos.x86_64.yml, which main has since folded into a matrix
inside ci.macos.yml. Without them the label trigger would survive on this
branch and keep producing all-skipped runs.

---------

* ci: resolve backport conflict, keep 0.7 job topology

release/0.7 runs a smaller x86_64 matrix than 0.8: it has gcc850, gcc921 and
gcc1031 only, none of which test epoll_ng, and it has no fstack job. The
cherry-pick therefore hit a modify/delete conflict over the whole region 0.8
had grown past — the epoll_ng step of gcc1031 plus the gcc1121, gcc1211 and
fstack jobs.

Take this branch's side: the three jobs it does have are gated on draft state
and the workflow triggers on ready_for_review, while what 0.7 builds and tests
is left exactly as it was. A change to when CI runs has no business changing
what CI covers.

ci.macos.x86.yml is deliberately untouched. It predates the label guard, so it
never produced the all-skipped phantom runs this change removes; adding a draft
gate there would be a new decision, not part of this backport.

---------

## Conflicts

- c988cb6499069d94eeb98916949399b730e09595: .github/workflows/auto-pr-backport.yml,.github/workflows/ci.linux.x86_64.yml,.github/workflows/ci.macos.arm.yml,.github/workflows/ci.macos.x86_64.yml

## Dropped new files

The following new files introduced by cherry-pick were dropped because they don't exist in `release/0.6`:

- .github/workflows/auto-pr-backport.yml
- .github/workflows/ci.linux.x86_64.yml
- .github/workflows/ci.macos.x86_64.yml


Coldwings:
Resolved and marked ready for review.

**What conflicted.** One region of `ci.macos.arm.yml`, where two unrelated edits sit on adjacent lines: the label guard becoming a draft check, and a `macos-14` → `macos-15` runner bump that `release/0.6` had never taken. Took the new `if:` and kept `macos-14` — which runner this branch builds on is a separate decision from when its CI fires.

**On the dropped files.** `release/0.6` predates the workflow split, so the three files the cherry-pick reported as dropped genuinely don't exist here: this branch has `auto-pr.yml` rather than `auto-pr-backport.yml`, and its CI files are `ci.linux.x86.yml` and `ci.macos.yml`. Dropping them was the right outcome — adding foreign copies next to the existing ones would have doubled up every build. Worth stating explicitly since the same drop step removed a legitimately new test file over in #1634.

**What this backport actually accomplishes here.** Only `ci.linux.arm.yml` and `ci.macos.arm.yml` ever carried the label guard on this branch, and both are now converted, so the all-skipped phantom runs are gone from 0.6. The label-cascade half of the change has nothing to attach to, because there is no `auto-pr-backport.yml` on this branch — which also means **the backport cascade terminates here**, despite `release/0.4` existing. `ci.linux.x86.yml` and `ci.macos.yml` are untouched by design: they predate the label guard entirely, so they never produced phantom runs, and adding a draft gate to them would be a new decision.

Net diff against `release/0.6` is 12 lines each in the two arm workflows and nothing else — trigger types, the `-noop` concurrency workaround removed, one `if:` per job. Runners untouched. Both files parse as YAML.

**Two expected reds**, both pre-existing on this branch and unrelated to the change: `macOS14-arm` fails because the Homebrew gtest now requires C++17 while this branch's macOS job configures C++14, and `macOS-12-Monterey-debug` will sit pending until it times out because its runner image is retired. #1622 shows the same pair on the same base. These are being tracked separately rather than fixed inside a backport.

Heads-up: merging this will trigger this branch's legacy `auto-pr.yml`, which opens a *forward* `release/0.6` → `release/0.7` PR. That's pre-existing behaviour of the old workflow, not something this change introduces.

Coldwings (approved):

PR #1640, opened 2026-09-07
feat(ecosystem): add OSS `ip_version` option backed by a client-owned resolver

Moonpuck:
## Motivation

In some sandboxed environments IPv6 is disabled, yet `getaddrinfo` may still return AAAA (IPv6) records. The OSS client picks a resolved address round-robin, so when it selects an IPv6 address the connection fails — causing intermittent, hard-to-reproduce request failures. This PR lets users pin the address family (e.g. force IPv4).

Per review feedback, instead of a `set_resolve_filter` hook on the shared per-vCPU resolver, the HTTP client now accepts a **client-owned `Resolver`**. Since the DNS cache lives inside the resolver, giving each client a private instance isolates its results (and any filtering) from other clients — so the filter can never be bypassed by another client's cached entry for the same hostname.

## Changes

- **HTTP client**: `Client::set_resolver(Resolver*, bool ownership = false)`. `PooledDialer::dial` now takes an optional resolver and falls back to the per-vCPU default when null, so the default behavior and the UDS path are unchanged. When `ownership` is set, `ClientImpl` deletes the resolver on destruction.
- **OSS client**: new `enum class IPVersion { kBoth, kIPv4Only, kIPv6Only }` and `ClientOptions::ip_version` (default `kBoth`, appended at the struct end for source compatibility). For a single-family option the client installs a private `FilteredResolver` — a decorator wrapping its own `DefaultResolver` that keeps only addresses matching `is_ipv4()` / `is_ipv6()`.
- **net/utils**: dropped the stale *"not thread safe"* note on `new_default_resolver`; its DNS cache is guarded by cross-vCPU-capable primitives (spinlock/mutex/condvar), verified under a multi-vCPU stress test with ASAN.

## Usage

```cpp
objstore::ClientOptions opts;
opts.endpoint   = "oss-<region>.aliyuncs.com";
opts.ip_version = objstore::IPVersion::kIPv4Only;  // force IPv4 in IPv6-less sandboxes
auto client = objstore::new_oss_client(opts, auth);
```

## Testing

- **`http_client.set_resolver`**: the injected resolver is actually used (reject-all fails the request, accept-all succeeds and the filter is invoked).
- **`test-oss-resolve-ip`** (new): end-to-end via the OSS client for both families — IPv4 (`127.0.0.1`) and IPv6 (`::1` via `localhost`) — each verifying accept-correct-family and reject-wrong-family. Each case runs in a fresh photon session (isolated DNS cache); cases skip when a family is unavailable in the environment.


net/http/client.h
```diff
@@ -167,6 +168,12 @@ class Client : public Object {
     void set_bind_ips(std::vector<IPAddr> &ips) {
         m_bind_ips = ips;
     }
+    // Filter applied to DNS-resolved addresses before connecting; an address is
+    // dropped when it returns false. Empty (default) accepts all. Must stay
+    // valid for the lifetime of this client.
+    void set_resolve_filter(Delegate<bool, IPAddr> filter) {
```
Coldwings, line 174: The default resolver is thread-safe and should able to share between per-thread `PooledDailer`, so each dailer owns unique resolver is not necessary, http client should use a client-owned resolver for each dailer.

Compare to add a `set_resolve_filter`, `set_resolver` should be a better idea.

net/http/client.h
```diff
@@ -21,6 +21,7 @@ limitations under the License.
 #include <photon/net/http/message.h>
 #include <photon/net/http/url.h>
 #include <photon/common/object.h>
+#include <photon/common/callback.h>
```
lihuiba: is callback.h still needed?

net/http/client.cpp
```diff
@@ -83,18 +83,19 @@ class PooledDialer {
...
+        return dial(x.host_no_port(), x.port(), x.secure(), resolver, timeout);
     }
 
     ISocketStream* dial(std::string_view uds_path, uint64_t timeout = -1ULL);
 };
 
-ISocketStream* PooledDialer::dial(std::string_view host, uint16_t port, bool secure, uint64_t timeout) {
+ISocketStream* PooledDialer::dial(std::string_view host, uint16_t port, bool secure, Resolver* resolver, uint64_t timeout) {
     LOG_DEBUG("Dialing to `:`", host, port);
+    if (!resolver) resolver = this->resolver.get();  // fall back to the per-vCPU default resolver
```
lihuiba: we should assume resolver is always valid here, and make the default in ```ClientImpl```

net/http/client.cpp
```diff
@@ -172,11 +173,17 @@ class ClientImpl : public Client {
...
+    ~ClientImpl() {
+        if (m_resolver_ownership) delete m_resolver;
+    }
     PooledDialer& get_dialer() {
         thread_local PooledDialer dialer;
         dialer.init(m_tls_ctx, m_bind_ips);
         return dialer;
     }
+    Resolver* resolver_of(PooledDialer& dialer) {
+        return m_resolver ? m_resolver : dialer.resolver.get();
```
lihuiba: put the default resolver instance inside ClientImpl

Moonpuck: Done. `dial()` now assumes a valid resolver, and the default **selection** lives in `ClientImpl::resolver_of` (`m_resolver ? m_resolver : dialer.resolver.get()`).

I kept the default resolver **object** in the thread-local dialer on purpose: `DefaultResolver`'s `ObjectCache` holds a `photon::Timer`, whose coroutine is bound to the vCPU that created it and must be `thread_join`ed while that vCPU is still alive. Since a client can be shared across vCPUs (`http_client.vcpu`), a client-owned default would bind the Timer to some ephemeral vCPU — once that vCPU `fini()`s, destroying the client joins a dead vCPU's coroutine and hangs. The dialer avoids this via its `fini_hook`, which resets the resolver while the owning vCPU is still alive.

A single global default resolver (one DNS cache shared across all vCPUs) would be a nice follow-up, but it isn't achievable in this change — it would first require the `ObjectCache` to support a Timer-free (lazy-expiry) mode, which is a separate `common/` change.

Moonpuck:
Currently the default resolver stays in the thread-local `PooledDialer`, and `dial()` falls back
to it (`if (!resolver) resolver = this->resolver.get();`).

lihuiba (approved):

PR #1641, opened 2026-09-08
feat(oss): read multiple byte ranges of an object in one GET

han-ian:
Scattered ranges of one object cost one GET each, which dominates the
latency of small reads. OSS can answer a whole list of them in a single
response, so get_object_ranges() asks for one and scatters the parts
into the caller's iovs:

    Range: bytes=0-1,100-149,900-999
    x-oss-multi-range-behavior: multi-range
    ->  206, Content-Type: multipart/byteranges;boundary=<uuid>
        three parts, each carrying its own Content-range

The list must hold more than one interval, ascending, non-overlapping
and inside the object. OSS does not reject a violating list, it silently
reorders or merges it, or answers something else altogether, and the
reply reads exactly like an answer to the list as sent -- so a violating
list is rejected locally with EINVAL. That leaves one part per requested
range, verified part by part while the body is scattered, and success is
reported only once every iov is full.

The signature cache needs one fix to go with it: every x-oss-* header is
signed, but CachedAuthenticator keys the cached Authorization on
bucket/object alone. A multi-range GET and a plain ranged GET of the
same object therefore answer 403 on each other's cached signature and
pay an extra round trip to re-sign, so the x-oss-* headers set by the
caller now take part in the key as well.

### DIFF
--- a/ecosystem/oss.cpp
+++ b/ecosystem/oss.cpp
@@ -17,6 +17,7 @@ limitations under the License.
 #include "oss.h"
 
 #include <netinet/in.h>
+#include <strings.h>
 #include <photon/common/alog-stdstring.h>
 #include <photon/common/alog.h>
 #include <photon/common/checksum/digest.h>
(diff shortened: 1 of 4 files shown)


Coldwings (approved):
LGTM

lihuiba (approved):

PR #1647, opened 2026-09-15
fix(tls): verify peer certificate against the requested hostname

juliusl:
## Problem

Photon checks that a server certificate chains to a trusted CA, but never checks that the certificate was **issued for the host being contacted**. A chain check alone proves a certificate is genuine, not who it belongs to, so any certificate signed by a trusted CA is accepted for any name.

Despite its name, `tls_stream_set_hostname()` only called `SSL_set_tlsext_host_name()` — SNI, which tells the server which certificate to send and places no constraint on the one that comes back. `X509_VERIFY_PARAM_set1_host()`, `SSL_set1_host()` and `X509_check_host()` appear nowhere in the tree, and all three `SSL_CTX_set_verify()` call sites pass a `nullptr` callback.

`net::http::Client` is affected in its default configuration, where it enables `VerifyMode::PEER` with the comment `// act like curl` — but unlike curl, no name check follows.

Separately, `PooledDialer` keyed its TLS connection pool by `EndPoint`, so two hostnames resolving to the same address shared one pooled connection, validated for whichever connected first. Fixing the name check without this would leave the same outcome reachable through connection reuse.

**Affected:** v0.9.0 through v0.9.6 and current `main`. `set_verify_mode()` and the HTTP client's `VerifyMode::PEER` default both arrived in eda9d90 (#621), first released in v0.9.0, so `release/0.9` is affected. `release/0.8` and earlier expose no certificate verification API at all, so this gap does not apply there.

## Changes

**Bind the name (`net/security-context/tls-stream.cpp`)**

- `tls_stream_set_hostname()` now sets verification parameters in addition to SNI: `X509_VERIFY_PARAM_set1_host()` for hostnames, `X509_VERIFY_PARAM_set1_ip_asc()` for IP literals, since `X509_check_host()` never matches `iPAddress` SANs.
- `X509_CHECK_FLAG_NO_PARTIAL_WILDCARDS` rejects wildcards covering part of a label (`www*.example.com`). Single-label, leftmost-only matching is already OpenSSL's default.
- The client handshake is deferred to the first I/O so that SNI is carried in the ClientHello; the verification parameters are set in that same window.

**Key the HTTP connection pool by hostname (`net/http/client.cpp`)**

`PooledDialer::tlssock` becomes an `ISocketPool` keyed by `host:port`, so a connection verified for one host is never reused for another. The `tls_stream_set_hostname()` result is now checked, and the request fails closed.

**Reject a hostname request on a `VerifyMode::NONE` context**

`SSL_set1_host()` is only consulted while the chain is verified, so under `SSL_VERIFY_NONE` it is silently inert — a mismatch is recorded in the verify result but the handshake still succeeds. `tls_stream_set_hostname()` therefore returns `-1`/`EINVAL` in that case rather than appearing to verify.

curl decouples `VERIFYHOST` from `VERIFYPEER` because it performs its own SAN matching after the handshake completes. Photon has no equivalent place to do that: the handshake runs implicitly inside `SSL_read`/`SSL_write`, so there is no completion hook to run a separate check from, and matching SANs and wildcards by hand would reimplement logic OpenSSL already provides. This change uses OpenSSL's matcher instead, which is consulted during chain verification — hence the two settings are linked here rather than independent.

Callers wanting the old behaviour say so explicitly — `tls_stream_set_sni()` for SNI only, or `TLSContext::set_verify_hostname(false)`, the analogue of `CURLOPT_SSL_VERIFYHOST=0`. `net::http::Client` honours the context's verify mode rather than hard-erroring, since a context left at `VerifyMode::NONE` has already opted out of certificate checking.

## Compatibility

**Behaviour change, intentional.** A client passing a hostname while trusting a CA-signed
certificate issued for a *different* name will now fail the handshake.

**Connection reuse.** Host-keyed TLS pooling lowers reuse for deployments fronting many
hostnames behind one address. Correctness requires it.

| API change | Impact |
|---|---|
| `tls_stream_set_hostname()` returns `int`, was `void` | Source-compatible; callers ignoring the result still compile. ABI-safe — return types are not mangled. |
| `tls_stream_set_sni()` added | New symbol. |
| `TLSContext::set_verify_hostname()`, `get_verify_mode()` added | Appended at the end of the class, but they do change vtable layout. Anyone linking a prebuilt shared library or implementing `TLSContext` themselves needs a rebuild. |

**OpenSSL.** Everything used is available from 1.0.2 onward, so no supported
configuration loses the check, and the gate is inert on 1.1.x and 3.x. Below 1.0.2 the function returns `-1`/`ENOSYS` rather than compiling the check out, so it cannot silently appear to verify. `PHOTON_BUILD_DEPENDENCIES=ON` statically links OpenSSL 1.1.1w and the macOS docs pin `openssl@1.1`, which is why the floor is not higher.

## Out of scope

`rpc::StubPool` connects by `EndPoint` with no hostname in its API, so there is nothing to check a certificate against; its TLS protects the channel, not the peer's identity. Left as a comment; adding an identity model there is outside the scope of this change.

## Tests

`test_cert_utils.h` gains a CA + leaf generator taking a SAN list, following the layout of curl's `tests/certs`. The shared certificate in `net/test/cert-key.cpp` has no SAN and `CN=DefaultCompanyLt`, so it cannot exercise name matching; it is left untouched.

Covered: mismatched name rejected; matching name accepted; common-name fallback with no SAN; wildcard matches one label but not two; IP SAN present succeeds, absent fails, and a `DNS:` spelling of an address does not satisfy an IP request; `VerifyMode::NONE` returns `-1` while both opt-outs still accept; pooled connections not reused across hostnames.

Two are negative controls rather than assertions of success:

- The mismatch rejection is paired with an opt-out case accepting *the same* certificate, showing the rejection comes from the name check and not a broken chain.
- The pool test fails when the pool is keyed by endpoint. It drains the first response body so that the connection returns to the pool; without that it does not exercise reuse.

Three existing tests changed state:

| Test | Reason | Resolution |
|---|---|---|
| `sni.hostname_in_client_hello` | Default context; asserts on ClientHello bytes | Uses `tls_stream_set_sni()`; it tests wire format, not verification |
| `client_tls.http_with_ca_cert` | Reaches `https://localhost`; shared cert has no matching name | Serves a generated cert with `DNS:localhost` |
| `client_tls.http_client_cross_thread_isolation` | Reaches `https://127.0.0.1` | Serves a generated cert with `IP:127.0.0.1` |

Locally `test-tls` (24 tests), `client_tls_test`, `test-rpc` and `websocket_test` pass. Two unrelated failures reproduce identically on an unmodified tree in the same environment: `utils.resolver_filter` in `test-socket`, which is DNS-dependent, and the `fs/test` gmock targets, which gtest 1.11 headers trip under `-Werror=maybe-uninitialized`.


juliusl:
I need a judgement call from maintainers,

Previously when SNI was set it allowed IP addresses to be set, meaning HostName could be set to an ip address. According to RFC 6066: 

>  Literal IPv4 and IPv6 addresses are not permitted in "HostName".

Since it would be a divergence of the current behavior so I opted to leave it as is for the scope of this PR.

Should I fix that in this current PR or leave it as is? (Or I guess file an issue since this might need some discussion)

Coldwings:
Thanks for checking. Let's preserve the existing SNI behavior in this PR and keep it focused on peer-identity verification and preventing cross-host TLS connection reuse. Please open a separate issue for omitting SNI when the host is an IP literal. IP SAN verification should remain enabled as implemented here. The SNI change does not need to block merging this fix.

Coldwings (approved):

PR #1651, opened 2026-09-15
[Backport][main to 0.9] | feat(oss): read multiple byte ranges of an object in one GET (#1641)

photonlibos:
> feat(oss): read multiple byte ranges of an object in one GET (#1641)

Fetching scattered ranges of one object costs one round trip each, which
dominates the latency of small reads. OSS can answer a Range list with a
multipart/byteranges body, so add get_object_ranges() to exploit it.

The range list must hold more than one interval, ascending and
non-overlapping. OSS would otherwise reorder or merge the intervals, or
drop the list and send the whole object, none of which the caller can
tell apart from a correct answer, so a violating request is rejected with
EINVAL instead of being sent. Under these constraints every part maps
onto exactly one requested range, which lets the parser verify each part
against the range it belongs to while scattering the payload straight
into the caller's iovs, and report success only once every iov is full.

A range list reaching past the end of the object is dropped the same
silent way, and is rejected with EINVAL once the answer turns out not to
be multipart.

Reading this way also needs the signature cache fixed. Every x-oss-*
header takes part in the signature, but CachedAuthenticator keys the
cached Authorization on bucket/object alone, so the multi-range headers
collide with those of a plain ranged GET of the same object: each kind
answers 403 on the other's cached signature and pays an extra round trip
to re-sign. Let the x-oss-* headers set by the caller take part in the
cache key as well.


Coldwings:
Close since the origin PR can not be built in MacOS, that PR will be reverted.

PR #1658, opened 2026-09-17
fix(ecosystem): refresh OSS GMT date on every wall-clock second change

Moonpuck:
## Background

`update_gmt_date()` cached the GMT `Date` / `x-oss-date` header strings and only refreshed them every 60 seconds, with the interval measured by `photon::now` (a CLOCK_BOOTTIME-based timestamp) while the cached values themselves came from the wall clock (`std::time`).

This breaks in the common sandbox suspend/resume scenario: when a sandbox wakes up, the system wall clock steps forward by the whole suspend duration (not a manual adjustment). The cached headers keep the pre-sleep timestamp, and the `photon::now`-based throttle does not notice the jump, so OSS requests can be signed with a stale Date for up to 60 seconds and get rejected (RequestTimeTooSkewed / signature mismatch).

## Change

Drop the interval-based throttling. Every sign() reads the wall clock directly (a vDSO call, ~tens of ns) and re-formats the cached strings whenever the second changes:

```cpp
void update_gmt_date() {  // avoid updating GMT Time every time
  std::time_t now = std::time(nullptr);
  if (now == m_last_tim) return;
  struct tm tm{};
  if (gmtime_r(&now, &tm)) {
    m_last_tim = now;
    strftime(m_gmt_date, GMT_DATE_LIMIT, "%a, %d %b %Y %H:%M:%S GMT", &tm);
    strftime(m_gmt_date_iso8601, GMT_DATE_LIMIT, "%Y%m%dT%H%M%SZ", &tm);
  }
}
```

The 1-second refresh grain exactly matches the resolution of the date formats, so it is never coarser than correctness requires, while `gmtime_r` + `strftime` still run at most once per second. The now-unused `GMT_UPDATE_INTERVAL` constant is removed.

After this change, the first request following a clock jump is already signed with the correct post-wake time.

## Files changed

- `ecosystem/oss.cpp` — rewrite `update_gmt_date()`
- `ecosystem/oss_constants.h` — drop unused `GMT_UPDATE_INTERVAL`

## Validation

- Compiles cleanly, no warnings.
- No interface changes; both signing paths (V1 / V4) are unaffected.
- No performance impact observed in the ossfs2 project's performance regression tests.


xqc-xiong (approved):

zhenjiaseu (approved):
LGTM

PR #1659, opened 2026-09-18
[Backport][main to 0.9] | fix(ecosystem): refresh OSS GMT date on every wall-clock second change (#1658)

photonlibos:
> fix(ecosystem): refresh OSS GMT date on every wall-clock second change (#1658)

The previous logic throttled GMT date updates via photon::now (a
monotonic-style timestamp) with a 60s interval. After a system clock
jump, the cached Date header could stay stale for up to 60 seconds,
causing OSS requests to be rejected (RequestTimeTooSkewed / signature
mismatch) until the throttle expired.

Now each sign() reads std::time() directly (a cheap vDSO call) and
re-formats the GMT strings whenever the second changes. The throttle
grain matches the 1-second resolution of the Date format, so it is
never coarser than correctness requires, and gmtime_r/strftime still
run at most once per second.


zhenjiaseu (approved):
LGTM

